PR #14, opened 2019-09-25
add notebook tutorials

simonkassel:
This PR adds two tutorial notebooks. The first gives a basic intro to and is adapted from the sat-stac tutorial. The second creates a stac for spacenet rio data and does some basic manipulations to it using the map_items method. 

The only change that this makes to pystac is adding the ability to pass keyword arguments to the item_mapper function.

to test:
- run jupyter notebooks (creating the cogs in the spacenet notebook takes some time)


pystac/catalog.py
```diff
@@ -138,7 +139,7 @@ def save(self):
 
         STAC_IO.save_json(self.get_self_href(), self.to_dict())
 
-    def map_items(self, item_mapper):
+    def map_items(self, item_mapper, **kwargs):
```
lossyrob: What's the inclusion of kwargs buy us here? Since we're passing in the method, the user can construct a method that closes over any parameters that don't involve the item. 

In your notebook example, I see:

```
spacenet_cog = spacenet.map_items(cogify_if_labeled, labels=labels, data_dir='/home/simon/Desktop/')
```

This could be written as:
```
spacenet_cog = spacenet.map_items(lambda item: cogify_if_labeled(item, labels=labels, data_dir='/home/simon/Desktop/'))
```
or
```
mod_cogs = lambda item: cogify_if_labeled(item, labels=labels, data_dir='/home/simon/Desktop/')
spacenet_cog = spacenet.map_items(mod_cogs)
```
Without the kwargs need (i.e we can take care of these needs with [partial application](https://en.wikipedia.org/wiki/Partial_application)).

simonkassel: oh, just saw this. Yeah, that makes sense. I will change

tutorials/pystac-introduction.ipynb
```diff
@@ -0,0 +1,894 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 51,
+   "metadata": {
+    "scrolled": true
+   },
+   "outputs": [],
+   "source": [
+    "# add a new catalog to a root catalog\n",
+    "kitten_json = {\n",
```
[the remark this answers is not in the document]
simonkassel: sounds good

simonkassel:
ok, I made these changes. Switched to vegas and use a subset of 10 images. It takes ~3.5 minutes to run the item mapper that creates the cogs. We could make the sample smaller if that still seems like too long.

PR #20, opened 2019-10-09
add collection property

### LINKED ISSUE #23: Add 'collection' property to item
needs to be added as a string property

simonkassel:
Add's a 'collection' property to the `Item` class

- 'collection' is the id of the collection to which the item belongs
- Optional parameter
- property is updated when `set_collection` is called

closes #23

### DIFF
--- a/pystac/item.py
+++ b/pystac/item.py
@@ -18,13 +18,15 @@ class Item(STACObject):
                  datetime,
                  properties,
                  stac_extensions=None,
-                 href=None):
+                 href=None,
+                 collection=None):
         self.id = id
         self.geometry = geometry
         self.bbox = bbox
         self.datetime = datetime
         self.properties = properties
         self.stac_extensions = stac_extensions
+        self.collection = collection
 
         self.links = []
         self.assets = {}
@@ -62,6 +64,7 @@ class Item(STACObject):
                 link_type = LinkType.ABSOLUTE
         self.remove_links('collection')
         self.add_link(Link.collection(collection, link_type=link_type))
+        self.collection = collection.id
         return self
 
     def to_dict(self, include_self_link=True):
@@ -130,6 +133,9 @@ class Item(STACObject):
         bbox = d['bbox']
         properties = d['properties']
         stac_extensions = d.get('stac_extensions')
+        collection = None
+        if 'collection' in d.keys():
+            collection = d['collection']
 
         datetime = properties.get('datetime')
         if datetime is None:
@@ -141,7 +147,8 @@ class Item(STACObject):
                     bbox=bbox,
                     datetime=datetime,
                     properties=properties,
-                    stac_extensions=stac_extensions)
+                    stac_extensions=stac_extensions,
+                    collection=collection)
 
         for l in d['links']:
             item.add_link(Link.from_dict(l))
--- a/pystac/item.py
+++ b/pystac/item.py
@@ -74,7 +74,7 @@ class Item(STACObject):
 
         assets = dict(map(lambda x: (x[0], x[1].to_dict()), self.assets.items()))
 
-        self.properties['datetime'] = '{}Z'.format(self.datetime.replace(microsecond=0, tzinfo=None))
+        self.properties['datetime'] = self.datetime.isoformat()
 
         d = {
             'type': 'Feature',
--- a/pystac/item.py
+++ b/pystac/item.py
@@ -89,6 +89,9 @@ class Item(STACObject):
 
         if self.stac_extensions is not None:
             d['stac_extensions'] = self.stac_extensions
+        
+        if self.collection:
+            d['collection'] = self.collection
 
         return d
 
--- a/pystac/item.py
+++ b/pystac/item.py
@@ -188,7 +188,7 @@ class Asset:
         self.href = href
         self.title = title
         self.media_type = media_type
-        self.properties = None
+        self.properties = properties
 
         # The Item which owns this Asset.
         self.item = None
@@ -220,8 +220,8 @@ class Asset:
             d['title'] = self.title
 
         if self.properties is not None:
-            for k in properties:
-                d[k] = properties[k]
+            for k, v in self.properties.items():
+                d[k] = v
 
         return d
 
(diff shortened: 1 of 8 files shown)


simonkassel:
@lossyrob I added unit tests but was not sure exactly how to test both of those methods so let me know if there are any tests you would like me to add

tests/test_item.py
```diff
@@ -0,0 +1,42 @@
...
+
+class ItemTest(unittest.TestCase):
+    def test_to_from_dict(self):
+        m = TestCases.get_path('data-files/itemcollections/sample-item-collection.json')
+        with open(m) as f:
+            item_dict_list = json.load(f)['features']
+        
+        item = Item.from_dict(item_dict_list[0])
+        self.assertIsInstance(item, Item)
+        self.assertIsInstance(item.id, str)
```
lossyrob: I'd say, these type tests are less valuable since the types are defined in the class ... it's probably going to be a rare case that one of these fails due to an actual bug, and more likely that a refactor would happen that this would be a nuisance test for. 

For me, this is an instance where the fact that the Item to_dict/from_dict gets run as part of the unit test is the value part, as any bugs will normally throw an exception. So it's really a coverage thing - which is helpful for dynamic languages like python.

Instead of an exhaustive list of type checks, I'd recommend picking out a few valuable cases to test values against - e.g. the `get_self_href` below is a good one. These will test the actual workings of the mechanism on this particular test case, rather than the mechanism itself broadly (the latter being more like the testing e.g. of Link types in the link property is more of that sort of test).

Potentially an easier way to do a more full test that is more broadly applicable is to just run the dict through `from_dict` and `to_dict`, and check that the dicts are equal. This might even reveal ways in which that process makes legitimate changes to the input dict.

This could also serve as a broader utility method, so that each of the STACObject subtypes could have a to/from dict test against sample cases.

simonkassel: yeah, this all makes sense. I changed the tests to check to to/from methods directly using a function like you mentioned. I put it in `test_utils.py` as a standalone function but let me know if you think it should be a method that goes elsewhere.

I added a few additional checks to test the Item object before it returns a dict but most of that stuff is taken out. Let me know if you want me to add it back in.

tests/test_item.py
```diff
@@ -0,0 +1,28 @@
...
+from tests.utils import (TestCases, RANDOM_GEOM, RANDOM_BBOX, test_to_from_dict)
+
+
+class ItemTest(unittest.TestCase):
+    def test_to_from_dict(self):
+        m = TestCases.get_path('data-files/itemcollections/sample-item-collection.json')
+        with open(m) as f:
+            item_dict = json.load(f)['features'][0]
+        
+        self.assertTrue(test_to_from_dict(Item, item_dict))
```
lossyrob: Nice. I originally said this could be a utility, but didn't think about how simple it would be - sorry! So I think just using the unittest.assertDictEquals would be the best move here:

```python
self.assertDictEqual(Item.from_dict(item_dict).to_dict(), item_dict)
```

and removing the utility method.

simonkassel: yeah, although I've encountered one edge case when testing the eo class: datetimes. since there are multiple different valid RFC339 strings for a timestamp, the to_dict method doesn't know the exact format of the string in the dict. Do we want to try to ensure that the the timestamp string that each object writes is the exact same as the one it reads?

lossyrob: Good call, that makes it tricky. I think we should allow for the datetime string to represent the same thing, but be different encodings.

Given that, it probably makes sense to write our own recursive dict checking code that handles these and any other exceptional cases that we may come across. I'd probably aim for it not too be too robust (no need to spend a ton of time engineering test code), but something that can handle the datetime case and have a place if we run into other exceptions to the `==` rule.

simonkassel: ok, so I'll just expand the `test_to_from_dict` method and leave it in the same place

lossyrob: Makes sense.

I will say, with that method just returning a boolean, there's not going to be a lot of help to the person debugging the test (including future you) if it'll fail with `False not equal to True`.

Since this is a test utility, you might want to just pass in `self` and do the asserts at the property level to check equality of the dictionary items, so that it'll fail with the actual values that didn't match.

You can also use the `msg` parameter in [assertEqual](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertEqual) to give the dict key that it failed on.

pystac/item.py
```diff
@@ -71,7 +74,7 @@ def to_dict(self, include_self_link=True):
 
         assets = dict(map(lambda x: (x[0], x[1].to_dict()), self.assets.items()))
 
-        self.properties['datetime'] = '{}Z'.format(self.datetime.replace(microsecond=0, tzinfo=None))
+        self.properties['datetime'] = self.datetime.isoformat()
```
lossyrob: This datetime format was intentional, as it matches https://tools.ietf.org/html/rfc3339#section-5.6 per the spec, and matches all the examples of datetime's in the repo.

tests/utils.py
```diff
@@ -162,3 +163,25 @@ def validate_dict(self, d, obj_type):
         except jsonschema.exceptions.ValidationError as e:
             print('Validation error in {}'.format(obj_type))
             raise e
+
+def test_to_from_dict(stac_object_class, d):
+    def _parse_times(dd):
+        a_dict = copy(dd)
```
lossyrob: Unclear what the copy/return value do here, since it seems to rely on in-place mutation of the dict (and sub-dictionaries are not reset into the parent copies).

pystac/item.py
simonkassel: got it

tests/utils.py
simonkassel: yeah, you're right on the return value. I copied the dict in `_parse_times` for the purpose of the input dict 'd'. Otherwise it will modify d and make it so that you can't reuse it in the test function, which you may want to do.

simonkassel: and then of course I guess if I include the copy then I need the return

tests/test_item.py
simonkassel: this makes sense

but i'm not sure how to use the msg param to report the key that it failed on. How would I access that within the assert method?

lossyrob: After thinking about it, if you just use `assertDictEquals`, you should be able to avoid recursing through each property and not have to use the `msg` param.

PR #21, opened 2019-10-14
eo extension

### LINKED ISSUE #11: Implement EO Extension
Includes implementing an  `EOItem` and `EOAsset`. May have to refactor how Assets are created in order to account for Asset extensions.

Also an open question: how do we handle extension properties in collections?

https://github.com/radiantearth/stac-spec/tree/master/extensions/eo

simonkassel:
Implements the eo extension

Adds the following classes in `eo.py`:
- EOItem (inherits from Item)
- EOAsset (inherits from Asset)
- Band

`EOItem`s and `EOAsset`s are essentially Item's and assets with additional attributes. They both follow a pattern in which we create Items and Assets that store the additional fields in a properties dict then convert them to EOItems and EOAssets respectively. 

Bands are essentially dicts that hold a series of optional parameters about the band. EOAssets have a 'band' field that include lists of indices of the corresponding bands within the list of Band objects that their item contains.

Closes #11

### DIFF
--- a/pystac/item.py
+++ b/pystac/item.py
@@ -189,8 +189,7 @@ class Asset:
         GEOJSON = 'application/geo+json'
         GEOPACKAGE = 'application/geopackage+sqlite3'
         HDF5 = 'application/x-hdf5'  # Hierarchical Data Format version 5
-        # Hierarchical Data Format versions 4 and earlier.
-        HDF = 'application/x-hdf'
+        HDF = 'application/x-hdf' # Hierarchical Data Format versions 4 and earlier.
 
     def __init__(self, href, title=None, media_type=None, properties=None):
         self.href = href
@@ -228,8 +227,8 @@ class Asset:
             d['title'] = self.title
 
         if self.properties is not None:
-            for k in self.properties:
-                d[k] = self.properties[k]
+            for k, v in self.properties.items():
+                d[k] = v
 
         return deepcopy(d)
 
@@ -256,9 +255,3 @@ class Asset:
                      media_type=media_type,
                      title=title,
                      properties=properties)
-
-    def is_eo(self):
-        if self.properties:
-            if self.properties.get('eo:bands', None):
-                return True
-        return False
--- a/pystac/item.py
+++ b/pystac/item.py
@@ -8,7 +8,9 @@ from pystac import STAC_VERSION
 from pystac.stac_object import STACObject
 from pystac.io import STAC_IO
 from pystac.link import (Link, LinkType)
-from pystac.utils import (make_relative_href, make_absolute_href, is_absolute_href)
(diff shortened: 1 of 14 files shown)


simonkassel:
rebased from `sk/add-item-collection-property`

pystac/item.py
```diff
@@ -247,3 +256,9 @@ def from_dict(d):
                      media_type=media_type,
                      title=title,
                      properties=properties)
+
+    def is_eo(self):
+        if self.properties:
+            if self.properties.get('eo:bands', None):
+                return True
+        return False
```
lossyrob: I don't think core classes should reference extensions. Is there a way to do whatever this is needed for on the EOItem class instead? Can the user just check the type instead of using this?

tests/test_eo.py
```diff
@@ -0,0 +1,217 @@
+import json
```
lossyrob: Nice coverage!

pystac/item.py
```diff
@@ -220,15 +228,16 @@ def to_dict(self):
             d['title'] = self.title
 
         if self.properties is not None:
-            for k, v in self.properties.items():
-                d[k] = v
+            for k in self.properties:
+                d[k] = self.properties[k]
```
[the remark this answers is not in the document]
simonkassel: I don't know. I'm not sure why I did this and can't figure it out so I'll change it back

pystac/item.py
simonkassel: Ok, that makes sense. they couldn't check the type because I use this to iterate over the assets within and item and check whether they should be converted to EO assets. I just turned it into a standalone function within eo.py. I could also put it in utils if that made more sense.

tests/utils.py
```diff
@@ -122,13 +122,17 @@ def test_case_3():
...
+    # schemas for Label and EO
+    SCHEMA_BASE_URI = 'https://raw.githubusercontent.com/simonkassel/stac-spec/sk/refactor-extension-schemas'
+    
     schemas = {
         Catalog: 'catalog-spec/json-schema/catalog.json',
         Collection: 'collection-spec/json-schema/collection.json',
         Item: 'item-spec/json-schema/item.json',
-        LabelItem: 'extensions/label/schema.json',
+        LabelItem: 'extensions/label/json-schema/stac-label-item-schema.json',
+        EOItem: 'extensions/eo/json-schema/stac-extension-eo.json'
```
lossyrob: More a comment on the stac-spec change, but I think this should try to match the other filenames, e.g.

```
json-schema/label-item.json
json-schema/eo-item.json
```

Since none of the others have  'stac' or 'schema' in them

simonkassel: yeah, i think that makes sense, will change in both places

PR #25, opened 2019-10-21
[WIP] account for extensions in collections

simonkassel:
Figure out how we we handle extension properties in collections. 

I.e. if a collection includes extended items like `EOItem`s, how should the collection store the properties specific to the EO extension.


simonkassel:
Currently, the collection includes eo properties if they are uniform across all eo items and Null values if they are not. 

this comment complicates that matter https://github.com/radiantearth/stac-spec/issues/413#issuecomment-469778789. Should EO items with differing properties even be allowed to be in the same collection?

lossyrob:
This functionality was put in place via another means - an Item merges in the Collection common properties when loading.

PR #38, opened 2019-11-04
Simplify install_requires

sgillies:
Putting python-dateutils directly in install_requires is simpler, allows code and a file to be deleted, and is more aligned with the guidance of the python packaging authority. See https://github.com/pypa/sampleproject/blob/master/setup.py#L143-L149.

I've also made it so that the testing server installs pystac before running the tests. Not only does this pull in the requirement listed in setup.py, it also checks that the package is installable, and tests the code as it would be run in production.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ python:
   - "3.7"
 
 install:
-  - pip install -r requirements.txt
   - pip install -r requirements-dev.txt
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ python:
 
 install:
   - pip install -r requirements-dev.txt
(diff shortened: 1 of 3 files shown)


lossyrob:
Thank you!

matthewhanson:
I think this was in there because they were using sat-stac as an initial template, and it's how I'd just been doing things for a while, although I've since forgotten why.

I think it was because separating out the requirements.txt file from setup allowed for quicker Docker builds during development, because I could copy and install the requirements as an earlier layer, and then edit the code and rebuild without having to reinstall the requirements. Of course, for  low-level library like this that has a single dependency that's not even much of a benefit.

I didn't realize this was the preferred "official" way - I think I've got several libraries I should be updating.

PR #48, opened 2019-11-07
fixing for new sn2 s3 structure

nrweir:
Corrected the various paths in the existing SpaceNet 2 Vegas tutorial to work for the new dataset organization. Ran through everything, it looks to me like it works.


nrweir:
Early example implementations of Label Extension used the actual label's geometries for the `geometry` parameter of the item, but it was eventually determined that the geometry should correspond to the entire area labeled (see [label extension geometry field description](https://github.com/radiantearth/stac-spec/tree/master/extensions/label#core-item-fields) and the discussion in [this issue](https://github.com/radiantearth/stac-spec/issues/524)). I therefore updated the geometry parameter to cover the same geography as the bbox (see [here](https://github.com/azavea/pystac/pull/48/files#diff-dc86303266d0303b4e887eba0f784b9cR315), which corresponds to the area labeled in this SpaceNet dataset's case.

docs/tutorials/pystac-spacenet-tutorial.ipynb
```diff
@@ -438,7 +433,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python [default]",
+   "display_name": "Python 3",
```
lossyrob: This may or may not upset `nbsphinx` for the Read The Docs rendering after merge - if so, I'll set this back to [default]

lossyrob: Nope, all good!

PR #53, opened 2019-11-19
Be consistent about owner vs item for owning item

jisantuc:
Overview
----

This PR makes the API around setting an owning item more obvious, and also refers to an actually existing property in `get_absolute_href` for `Asset`s.

Testing
-----

- tests

### DIFF
--- a/pystac/eo.py
+++ b/pystac/eo.py
@@ -262,7 +262,7 @@ class EOAsset(Asset):
         properties (dict): Optional, additional properties for this asset. This is used by
             extensions as a way to serialize and deserialize properties on asset
             object JSON.
-        item (Item or None): The Item this asset belongs to.
+        owner (Item or None): The Item this asset belongs to.
     """
     def __init__(self,
                  href,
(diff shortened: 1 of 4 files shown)


pystac/item.py
```diff
@@ -330,7 +330,7 @@ def __init__(self, href, title=None, media_type=None, properties=None):
         self.properties = properties
 
         # The Item which owns this Asset.
-        self.item = None
+        self.owner = None
```
lossyrob: Can you also update the Attributes PyDocs above?

pystac/eo.py
```diff
@@ -356,9 +356,9 @@ def get_bands(self):
             of asset.
         """
 
-        if not self.item:
+        if not self.owner:
```
lossyrob: The Attributes Pydocs for EOItem should also be updated.

jisantuc: 09f883a

pystac/item.py
jisantuc: 09f883a

PR #58, opened 2019-12-13
Fix link to STAC's site

brunobcardoso:
The link to STAC's site was wrong and pointing to a non-existing address.

https://stacgeo.org => https://stacspec.org

PS: Also changed test case URLs just to keep the consistency.


lossyrob:
Thanks for this!

Issue #56, opened 2019-12-02
Datetime format doesn't work well with libraries expecting ISO8601 formatted strings


kbgg:
While the STAC spec specifies that datetimes should be formatted according to [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6), the space separator between the date and time causes some issues with libraries expecting an ISO8601 formatted string. Changing the separator to 'T' would make the datetime compliant with both RFC 3339, section 5.6 as well as ISO8601.

For example, currently a datetime is formatted like this `2019-01-01 00:00:00Z` and the updated datetime string would be formatted like this `2019-01-01T00:00:00Z`.

lossyrob:
Thanks for the report - hope to get to this in the next week or so!

PR #70, opened 2020-01-24
Respect tzinfo in the provided datetime

tschaub:
It looks like the fix for #56 got closer to being able to represent dates as [RFC-3339](https://tools.ietf.org/html/rfc3339#section-5.6) formatted strings.  However, it looks like any `tzinfo` in the provided datetime is ignored.  So the current `utils.datetime_to_str()` function works for timezone naive datetimes, but not for timezone aware datetimes.

This branch proposes to treat timezone naive datetimes as UTC datetimes (in line with the current assumptions) but respects any `tzinfo` in the provided datetime.

### DIFF
--- a/pystac/utils.py
+++ b/pystac/utils.py
@@ -133,7 +133,7 @@ def is_absolute_href(href):
 
 
 def datetime_to_str(dt):
-    """Convert a python datetime to an ISO8601 stirng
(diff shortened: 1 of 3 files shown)


tschaub:
Thanks for the nice library, @lossyrob.  

I updated the branch to prefer the "Z" suffix for zero-offset timestamps.

PR #80, opened 2020-02-05
allow v0.7.0 sar items to fit in version range

phloem7:
v0.7.0 items with sar extension properties are incorrectly categorized as v0.6.2 due to a rogue "not" in serialization. These properties were lists prior to v0.7.0. `sar:absolute_orbit` is most problematic, as it's the only shared attribute name between the properties that turned from list in 0.6.2 to integer in 0.7.0.

0.6.2: https://github.com/radiantearth/stac-spec/tree/v0.6.2/extensions/sar
0.7.0: https://github.com/radiantearth/stac-spec/tree/v0.7.0/extensions/sar

Test data from: https://github.com/radiantearth/stac-spec/blob/v0.7.0/extensions/sar/examples/sentinel1.json

### DIFF
--- a/pystac/serialization/identify.py
+++ b/pystac/serialization/identify.py
@@ -137,7 +137,7 @@ def _identify_stac_extensions(object_type, d, version_range):
                         'sar:pixel_spacing', 'sar:looks'
                 ]:
                     if prop in d['properties']:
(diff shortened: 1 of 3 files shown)


lossyrob:
This is great, thanks for the catch and the new test case!

PR #96, opened 2020-05-19
Update to 0.9

simonkassel:
This PR updates pystac to comply with STAC version 0.9. It does not add new extensions but updates the existing code (as well as unit test example JSON files to work with the newest version of STAC).

Individual changes:
- updates unit test example data files so that they comply with 0.9
- adds `CommonMetadata` class which stores all common fields present in item properties that are not actual attributes of items. 
- removes specific properties from EOItems
- removes version attribute from Collections
- Changes docstring for license aregument in collections
- removes GSD and accuracy for Bands objects
- adds rolls to asset definition
- adds utility function to get bbox from geojson without needing shapely. Could be used if we want to allow users to construct an item without needing to provide the bbox (though geometry would still be required)
- changes stac_object from_file method to only try setting a root_link if the item has one. This was failing on the unit tests when I was trying to read an item (with no root) in from a file

Addresses many of the items in #96


tests/data-files/examples/0.9.0/item-spec/examples/CBERS_4_MUX_20181029_177_106_L4.json
```diff
@@ -44,6 +44,33 @@
...
+                "center_wavelength": 0.66,
+                "full_width_half_max": 0.06
+            },
+            {
+                "name":"B8",
+                "common_name": "NIR",
+                "center_wavelength": 0.83,
+                "full_width_half_max": 0.12
+            }
+        ],
```
simonkassel, line 73: i was able to fill in this band information just from googling the CBERS-4 but the same info was not as obvious for DigitalGlobe, Landsat-8 and Planet examples. The tests pass with just an empty list value for `eo:bands` but it may be better if we fill in that info in the example.

pystac/item.py
```diff
@@ -384,4 +411,102 @@ def from_dict(d):
                      media_type=media_type,
                      title=title,
                      description=description,
+                     roles=roles,
                      properties=properties)
+
+
+class CommonMetadata:
```
lossyrob, line 418: General comment on the implementation of CommonMetadata.

The current implementation would work well for retrieving metadata. But the API implement here for setting data could be improved. For instance I would image a user would think this would work, but it won't:

```python
item.common_metadata.license = 'something'
print(item.common_metadata.license)
> None
```

One would have to do this to property set the license:

```python
cm = item.common_metadata
cm.license = 'something'
item.set_common_metadata(cm)
print(item.common_metadata.license)
> something
```

I think there's a way to implement this property object idea while having all of the data still stored in the original item. Consider an approach where:

- Instead of the CommonMetada __init__ building up internal properties from the passed in `properties` object, just store that properties object.
- Define property getters and setters for each of the common metadata properties. The getters operate the same as the logic you use to set the properties currently in the __init__ - you return the values directly from the `properties` object. The setters then place the value directly into the properties object in the format that `to_dict` would already use. This way you set the properties directly on the property object of the item in the correct forrmat.

For example, if I replace the current implementation of CommonMetadata with this:

```python
class CommonMetadata:
    def __init__(self, properties):
        self.properties = properties

    @property
    def license(self):
        return self.properties.get('license')

    @license.setter
    def license(self, v):
        self.properties['license'] = v
```

And then run a similar set of code, I get the  expected result:

```python
item.common_metadata.license = 'something'
print(item.common_metadata.license)
> something
```

This also gets rid of the need for a `to_dict` method, as the properties are set directly into the item's properties and will be written out with the Item's `to_dict` method properly:

```python
print(item.properties)
> {'datetime': '2019-10-04T18:55:37Z', 'license': 'something'}
```

If we go this route, I would also recommend getting rid of the CommonMetadata.from_dict method - this is a sort of 'view' object that is different from other stac objects in the codebase and we wouldn't expect users to really use them directly.

simonkassel: Makes a lot of sense, just took a shot at implementing this and pushed it up

simonkassel: this commit https://github.com/azavea/pystac/tree/133955b146325dc4040a4aaaf7e0044549fe7ff6

PR #97, opened 2020-06-04
Update to 0.9

simonkassel:
New version of #96, with cleaned up commit history

----
This PR updates pystac to comply with STAC version 0.9. It does not add new extensions but updates the existing code (as well as unit test example JSON files to work with the newest version of STAC).

Individual changes:
- updates unit test example data files so that they comply with 0.9
- adds `CommonMetadata` class which stores all common fields present in item properties that are not actual attributes of items. 
- removes specific properties from EOItems
- removes version attribute from Collections
- Changes docstring for license aregument in collections
- removes GSD and accuracy for Bands objects
- adds rolls to asset definition
- adds utility function to get bbox from geojson without needing shapely. Could be used if we want to allow users to construct an item without needing to provide the bbox (though geometry would still be required)
- changes stac_object from_file method to only try setting a root_link if the item has one. This was failing on the unit tests when I was trying to read an item (with no root) in from a file

Addresses many of the items in #96

### DIFF
--- a/pystac/item.py
+++ b/pystac/item.py
@@ -264,7 +264,15 @@ class Item(STACObject):
         return CommonMetadata(self.properties)
 
     def set_common_metadata(self, common_metadata, override=False):
-        for k, v in common_metadata.to_dict().items():
+        """Update common metadata properties within an item from a CommonMetadata
+        object
+
+        Args:
+            common_metadata (CommonMetadat): The CommonMetadata object which
+                will define fields in the item's properties
+            override (bool, optional): Whether or not to override . Defaults to False.
+        """
+        for k, v in common_metadata.properties.items():
             if k in self.properties:
                 if override:
                     self.properties[k] = v
@@ -425,8 +433,9 @@ class CommonMetadata:
             common metadata fields in
 
     Attributes:
+        properties (dict): Dict of all common metadata attributes
         title (str): Human readable title describing the item
-        description (str): Detailed, description of the item
+        description (str): Detailed description of the item
         start_datetime (datetime): Start date and time for the item
         end_datetime (datetime): End date and time for the item
         license (str): Item's license(s), either SPDX identifier of 'various'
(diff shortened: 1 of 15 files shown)


pystac/item.py
```diff
@@ -258,6 +259,31 @@ def from_dict(cls, d, href=None, root=None):
...
+    @property
+    def common_metadata(self):
+        """Access the item's common metadat fields as a CommonMetadata object
+
+        Returns:
+            CommonMetada: contains all common metadata fields in the items properties
+        """
+        return CommonMetadata(self.properties)
+
+    def set_common_metadata(self, common_metadata, override=False):
```
lossyrob, line 271: Since properties edited through the common_metadata will edit the item's properties directly now, we should drop the `set_metadata` method.

pystac/stac_object.py
```diff
@@ -419,9 +419,12 @@ def from_file(cls, href):
 
         # If this is a root catalog, set the root to the catalog instance.
         root_link = o.get_root_link()
-        if not root_link.is_resolved():
-            if root_link.get_absolute_href() == href:
-                o.set_root(o, link_type=root_link.link_type)
+
+        if root_link is not None:
```
lossyrob, line 423: Is this a fix to something you hit on? Wondering if we should make this fix independently as I think we'll release a 0.8 bugfix version before the 0.9 upgrade.

tests/test_item.py
```diff
@@ -50,3 +52,86 @@ def test_read_eo_item_owns_asset(self):
...
+        self.assertEqual(self.EXAMPLE_CM_DICT['platform'], item_2_copy.properties['platform'])
+        self.assertNotEqual(item_2_copy.properties['platform'], self.ITEM_2.properties['platform'])
+        self.assertIsInstance(item_2_copy.properties['start_datetime'], str)
+        self.assertEqual(item_2_copy.properties['start_datetime'],
+                         self.EXAMPLE_CM_DICT['start_datetime'])
+        self.assertEqual(item_2_copy.properties['platform'], self.EXAMPLE_CM_DICT['platform'])
+        for i, provider in enumerate(item_2_copy.properties['providers']):
+            self.assertDictEqual(provider, self.EXAMPLE_CM_DICT['providers'][i])
+
+    def test_common_metadata_getter_setters(self):
```
lossyrob, line 119: Does this cover all of the property getter/setters that have non-trival logic?

Not necessary but consider breaking each property that has non-trivial logic into it's own test for clarity.

pystac/stac_object.py
simonkassel: Ok, I remembered what the issue was here. This was erroring out the unit tests [here](https://github.com/azavea/pystac/blob/b318a24b89e8854f7f80a4e0d71217acad2b5795/tests/test_item.py#L61) when it  tried to create an item from [this file](https://github.com/azavea/pystac/blob/sjk/interactive-rebase-0.9/tests/data-files/examples/0.9.0/item-spec/examples/datetimerange.json). It doesn't have a `root_link`.

You're right, this seems like a general bug fix so I can put it in the bug fix PR

simonkassel: on second thought, reverting this change will cause the CI to fail so maybe it makes sense to keep it in. I guess it depends on the order that the branches get merged. I will hold off on changing this for now. Let me know what course of action you think would be best.

lossyrob:
Looking great, merging into the feature branch for 0.9

PR #98, opened 2020-06-04
Fix #88

### LINKED ISSUE #88: Error when saving and then loading Catalog
Using PySTAC 0.3.3, the following code executed as a script works as expected (it saves a STAC and then opens it and prints the items). But if you uncomment the code in Block A, Line B will fail with the following stack trace. This especially strange because all that Block A does is read from the catalog before saving it. This came up in the context of trying to modify some links to GeoTIFF files in an existing stack, and then saving the modified copy. I found that the following was the minimal example that replicates the error.

```python
from urllib.parse import urlparse
from os.path import join

import boto3
from pystac import STAC_IO
from pystac import Catalog, LabelItem, CatalogType

# Copied from PyStac tutorial.
def setup_stac_s3():
    def my_read_method(uri):
        parsed = urlparse(uri)
        if parsed.scheme == 's3':
            bucket = parsed.netloc
            key = parsed.path[1:]
            s3 = boto3.resource('s3')
            obj = s3.Object(bucket, key)
            return obj.get()['Body'].read().decode('utf-8')
        else:
            return STAC_IO.default_read_text_method(uri)

    def my_write_method(uri, txt):
        parsed = urlparse(uri)
        if parsed.scheme == 's3':
            bucket = parsed.netloc
            key = parsed.path[1:]
            s3 = boto3.resource("s3")
            s3.Object(bucket, key).put(Body=txt)
        else:
            STAC_IO.default_write_text_method(uri, txt)

    STAC_IO.read_text_method = my_read_method
    STAC_IO.write_text_method = my_write_method

# Open a catalog and then save a copy locally.
setup_stac_s3()
stac_uri = ('s3://rasterfoundry-production-data-us-east-1/stac-exports/'
            'fd478c2b-3f71-41e4-a87b-e97a8a0d0afa/catalog.json')
cat = Catalog.from_file(stac_uri)
'''
# Block A. If this is uncommented, then Line B will fail.
for item in cat.get_all_items():
    print(item)
'''
new_stac_uri = '/opt/data/foo/test-stac-catalog'
cat.normalize_hrefs(new_stac_uri)
cat.save(catalog_type=CatalogType.SELF_CONTAINED)

# Open the local copy and iterate over it.
new_stac_uri = '/opt/data/foo/test-stac-catalog'
cat = Catalog.from_file(join(new_stac_uri, 'catalog.json'))
# Line B
for item in cat.get_all_items():
    print(item)
```

Stack trace:
```shell
  File "/opt/conda/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/opt/conda/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/opt/src/foo/pystac_test.py", line 51, in <module>
    for item in cat.get_all_items():
  File "/opt/conda/lib/python3.6/site-packages/pystac/catalog.py", line 281, in get_all_items
    yield from child.get_all_items()
  File "/opt/conda/lib/python3.6/site-packages/pystac/catalog.py", line 281, in get_all_items
    yield from child.get_all_items()
  File "/opt/conda/lib/python3.6/site-packages/pystac/catalog.py", line 279, in get_all_items
    yield from self.get_items()
  File "/opt/conda/lib/python3.6/site-packages/pystac/stac_object.py", line 260, in get_stac_objects
    link.resolve_stac_object(root=self.get_root())
  File "/opt/conda/lib/python3.6/site-packages/pystac/link.py", line 160, in resolve_stac_object
    obj = STAC_IO.read_stac_object(target_href, root=root)
  File "/opt/conda/lib/python3.6/site-packages/pystac/stac_io.py", line 119, in read_stac_object
    return cls.stac_object_from_dict(d, href=uri, root=root)
  File "/opt/conda/lib/python3.6/site-packages/pystac/serialization/__init__.py", line 31, in stac_object_from_dict
    merge_common_properties(d, json_href=href, collection_cache=collection_cache)
  File "/opt/conda/lib/python3.6/site-packages/pystac/serialization/common_properties.py", line 47, in merge_common_properties
    collection = STAC_IO.read_json(collection_href)
  File "/opt/conda/lib/python3.6/site-packages/pystac/stac_io.py", line 96, in read_json
    return json.loads(STAC_IO.read_text(uri))
  File "/opt/conda/lib/python3.6/site-packages/pystac/stac_io.py", line 61, in read_text
    return cls.read_text_method(uri)
  File "/opt/src/foo/pystac_test.py", line 19, in my_read_method
    return STAC_IO.default_read_text_method(uri)
  File "/opt/conda/lib/python3.6/site-packages/pystac/stac_io.py", line 12, in default_read_text_method
    with open(uri) as f:
FileNotFoundError: [Errno 2] No such file or directory: '/opt/data/foo/test-stac-catalog/12cb17a8-ae08-469c-a2be-4d0619240014/400c22e3-5b54-438b-b600-5e9bd6d0a498/3c67b59c-2e6f-47fb-ba3c-0dd106941096/collection.json'
```

lossyrob:
There was an error in the logic of `fully_resolve` that was preventing the recursive resolution from happening when a link was previously resolved to a STAC object.  (See changes in a696173 for fix).

This PR also includes a fix for an issue I saw where a STAC Object's `self_href` was not being set properly when reading self-contained catalogs.

Additionally, the collection cache is set correctly when a collection's self_href is updated.

Fixes #88


pystac/collection.py
```diff
@@ -76,6 +76,30 @@ def __init__(self,
...
+            update the HREF cache.
+        """
+        root = self.get_root()
+        if root is not None:
+            root._resolved_objects.remove(self)
+
+        super().set_self_href(href)
+
+        if root is not None:
+            root._resolved_objects.cache(self)
```
simonkassel, line 99: I just want to make sure this is making sense here because at first pass it seemed redundant to me: un-resolve the collection to just re-resolve it. Is the idea that in line 96, the href is set correctly according to the Catalog_Type of root?

lossyrob: Yeah, removing it from the cache de-registers it with the previous HREF - collections are cached by href in order to facilitate the collection metadata merging (the thing that moved to the commons extension in 1.0). The confusing logic is a good point and I'll add comments to clarify.

simonkassel: Ok, that makes sense

PR #107, opened 2020-06-28
Refactor extensions.

lossyrob:
This modifies extensions to wrap STACObjects instead of inhereting
from them. These changes apply to the 'eo' and 'label' extensions -
not the single-file-stac extensions, which is different and will be
dropped for STAC 1.0.

TODO:
- [x] Finish EO extension
- [x] Finish Label extension
- [x] Create View extension
- [x] Test custom extension
- [x] Documentation


pystac/extensions/eo.py
```diff
@@ -0,0 +1,273 @@
+from copy import deepcopy
+
+from pystac import (STACError, Extensions)
+from pystac.item import (Item, Asset)
+from pystac.extensions.base import (ItemExtension, ExtensionDefinition, ExtendedObject)
+
+
+class EOItemExt(ItemExtension):
+    """EOItem represents a snapshot of the earth for a single date and time.
```
lossyrob, line 9: Reminder to fix docstring

lossyrob:
Changes to implement based on conversation with @simonkassel:

- Rename `implements` to `extend`, e.g. `item.ext.extends('label')`
- Make implementing an extension explicit - e.g. if you call `item.ext.label` with an item that doesn't extend label, throw an error. Allow implementing an extension through a new method e.g. `item.ext.extend('label')`

simonkassel:
@lossyrob do you want me to make these changes https://github.com/azavea/pystac/pull/107#issuecomment-654956097

lossyrob:
@simonkassel I forgot to note here, but I did implement them and then had a change of heart after writing things out. I went with

```python
item.ext.enable("eo") # Enabled the extension
item.ext.implements("eo") # Checks if the item implements the extension
```

I also have the changes to force explicit enabling in b317ecf. Thank you though!

simonkassel:
this looks really good.

Jumping in here last minute after you've put much more thought into it (so feels free to disregard), but here are a couple thoughts I had.

So if I understand it correctly: 

consider: `item.stac_extensions = [ ]`

1. `eo_ext = `item.ext.eo` would throw an error
2. `item.ext.enable("eo")` basically just adds 'eo' to the item's stac_extensions property 
3. `item.ext.implements("eo")` checks whether 'eo' is in 'stac_extensions'
4. `eo_ext = item.ext.eo` creates the `EOItemExt` object whether the required props `bands`, `gsd`, etc are in item.properties or not, but at this point `eo_ext.gsd` (for example) is not defined
5. and then `eo_ext.apply(bands=x, gsd=y, ...)` actually changes the properties attribute of the underlying item

Is there one too many steps in there? I guess why not just add the wrap 2 into 1, if you don't need to have the underlying properties to create the extension item anyway? or maybe 4 and 5 are combined so item.ext.eo is a method that takes `gsd`, `bands`, etc. 

Also, in this case (as I noticed in [this case](https://github.com/azavea/pystac/pull/110)), to access the item's non-extension properties you've got to go `eo_ext.item.[item_property]`. That works for sure, and maybe helps  to differentiate item properties from extension properties (as I type this maybe I'm talking myself into it) but maybe it would be easier to access `id` directly from `eo_ext`.

lossyrob:
Thanks for the comments @simonkassel, this is very helpful to process even if they don't result in direct changes.

> why not just add the wrap 2 into 1

That's how I initially had it. It felt a little bit magical to add the extension ID to the `stac_extension` property just by accessing it - and it feels like you might get into trouble with it. Say you were reading a STAC Item you thought it implemented an extension, but it didn't - now you say `item.ext.eo.cloud_cover` and see that it's not set, assuming that this was an intentional non-value for an optional parameter. But the item didn't implement `eo` to begin with. Also now the item _does_ implement `eo`, but just because of a mistaken access. I think it's better to throw in this case and allow the user to see that they were mistaken something was an `eo` item. I also don't think it's too heavy of a lift to force someone who is creating the object to explicitly enable the extension, either by passing it in as as an element to the `stac_extension` or by calling e.g. `item.ext.enable(pystac.Extensions.EO)`. While I agree this is a bit annoying typing-wise, my sense is that making it explicit will save some hours of debugging for folks somewhere down the road.

> or maybe 4 and 5 are combined so item.ext.eo is a method that takes gsd, bands, etc.

I like where that idea is coming from, but am unsure how to implement it. The reason for having the `apply` and `create` (which are not elegant to me) is because the constructor of the extension (e.g. LabelItemExt) needs to take in the item, and Python doesn't have available overloaded constructors. That said, that wouldn't necessarily block some sort of syntax like

```
item.ext.eo(gsd, bands)
```

However the dynamic property magic of `eo` is through the `__getattr__`. I just tested and you can't do something clever like return a function from `__getattr__` to enable that sort of syntax. Besides that, you wouldn't want to force passing in arguments because that works in the construction case (when you are creating an item for example) and not on the read side (where you just want to access the extension properties).

This is all too say that, I agree that it's a bit more wonky and step-heavy than desired but I'm not sure how to shift it for the better and to line up with your on-point comments.

> Also, in this case (as I noticed in this case), to access the item's non-extension properties you've got to go eo_ext.item.[item_property]

I wasn't imagining that users would want to access the item properties through the extension wrapper, as they should have a reference themselves. In my own usage of this I've been accessing the wrapper directly off of the object, like

```
item.ext.eo.gsd = 1.2
item.ext.eo.set_asset_bands(asset, band_names
etc...
```

And if you want to capture the extension wrapper (which some parts of the unit test do), you should still have the original reference to the item property:

```
eo_ext = item.ext.eo
eo_ext.gsd
item.description
```

PR #110, opened 2020-07-09
fix __repr__ bug

simonkassel:
This surfaced when trying to print out an eoextension obj:

tries to reference the ext id rather than the underlying item


lossyrob:
Good catch! CI is failing because of linter failures that are fixed in the main branch. :+1:

Issue #117, opened 2020-07-15
Update PySTAC to use STAC 1.0.0.beta


simonkassel:
- [x] Update schemas to 1.0.0-beta.2 versions and fix so that tests pass.

Changes to account for (from [CHANGELOG](https://github.com/radiantearth/stac-spec/blob/master/CHANGELOG.md#v100-beta2---2020-07-08)). Includes updates from both `v1.0.0-beta.1` and `v1.0.0-beta.2`.

Crossed out items did not require changes in PySTAC.

### Added
<ins>v1.0.0-beta.1 - 2020-07-08</ins>
- [x] ~JSON-schema file in the Point Cloud extension.~

<ins>v1.0.0-beta.2 - 2020-07-08</ins>
- [x] ~'alternate' as a listed 'rel' type with recommended 'text/html' to communicate there is an html version.~
- [x] ~Added a code of conduct based on github's template.~
- [x] ~Overview document that gives a more explanatory discussion of the various parts of the spec and how they relate~
- [x] ~Several new sections to 'best practices' document.~
- [x] Added the ability to define Item properties under Assets (item-spec/item-spec.md) #122 
  - See [additional fields for assets](https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#additional-fields-for-assets) and the relevant [best practices](https://github.com/radiantearth/stac-spec/blob/master/best-practices.md#common-use-cases-of-additional-fields-for-assets) documentation.
- [x] Add proj:shape and proj:transform to the projections extension. (@lossyrob)
- [ ] Collection-level assets extension (see [extension docs](https://github.com/radiantearth/stac-spec/tree/master/extensions/collection-assets)) (@lossyrob)
- [x] ~Instructions on how to run check-markdown locally~
- [ ] Timestamps extensions (adds fields published, expires and unpublished)
- [x] `created` and `updated` can be used in the assets to specify the creation / update times of the assets. (@lossyrob)
- [ ] Tiled Assets extension, for representing data that has been split into tiles (@lossyrob)

### Changed
<ins>v1.0.0-beta.1 - 2020-07-08</ins>
- [x] Clarification on null geometries, making bbox not required if a null geometry is used. (@simonkassel )
- [x] Multiple extents (bounding boxes / intervals) are allowed per Collection (@simonkassel )

<ins>v1.0.0-beta.2 - 2020-07-08</ins>
- [x] ~Label extension types were clarified and types in README and JSON schema were brought into alignment~
[...]

PR #123, opened 2020-07-29
Some changes for upgrade to STAC-Spec v1.0.0.beta

simonkassel:
Addresses changes in #117 

**Completed so far:**

>  Clarification on null geometries, making bbox not required if a null geometry is used.
- Made bboxes an optional parameter but kept geometry required

> Multiple extents (bounding boxes / intervals) are allowed per Collection

> Moved eo:gsd from eo extension to core gsd field in Item common metadata
asset extension renamed to item-assets and renamed assets field in Collections to item_assets 

> ItemCollection removed from stac-spec core repo, will migrate to stac-api-spec as that is the only place it is used.

- I removed the ItemCollection inheritance from SingleFileStac and began transitioning it to inherit from catalog but wasn't quite able to iron out all the kinks.


pystac/item.py
```diff
@@ -22,11 +22,11 @@ class Item(STACObject):
         geometry (dict): Defines the full footprint of the asset represented by this item,
             formatted according to `RFC 7946, section 3.1 (GeoJSON)
             <https://tools.ietf.org/html/rfc7946>`_.
-        bbox (List[float]):  Bounding Box of the asset represented by this item using
-            either 2D or 3D geometries. The length of the array must be 2*n where n is the
-            number of dimensions.
         datetime (datetime): Datetime associated with this item.
         properties (dict): A dictionary of additional metadata for the item.
+        bbox (List[float] or None):  Bounding Box of the asset represented by this item using
```
lossyrob, line 27: Merging, but making a note of how I will change this - I'm going to revert back to the bbox's original position and have it not default to None. This is for a couple of reasons:
- This change will break the API and make upgrades more difficult, which isn't too heavy of a price but always good to avoid where possible
- The null geometry case is special, and so having a default of None goes against the standard default case (where bbox is required).

I think having the user be explicit about passing in None for the special null-geometry case is fine, and better than potential confusion about whether or not the bbox is required in the standard case where there is a geometry.

PR #139, opened 2020-08-13
Implement validation

### LINKED ISSUE #118: Add validation capabilities to PySTAC
When using PySTAC it would be convenient to be able to validate STAC objects. This is true for incoming STAC objects - for example, you might want to run validation against the JSON of a STAC object against an older version before reading in with PySTAC (and thereby migrating it to the latest STAC version). This is also true of STAC you construct with PySTAC - though the structure of PySTAC tends towards making valid STAC objects, a user can set a property that breaks the specification (e.g. setting a string instead of an array of strings) and it would be convenient to be able to check validity on-demand to avoid this.

I believe that validation should not happen by default in the core package. For one, this would add a dependency to the library, which aims to minimize dependencies. Also, validation takes time - and when dealing with many STAC objects, it should be up to the user if they want to pay the cost of that validation time or avoid it altogether.

An approach to solving this is as follows:

### Validate STACObjects

- Create a new package, `pystac_validator`. This will expand on the [SchemaValidator](https://github.com/azavea/pystac/blob/develop/tests/utils/validator.py) already used in the unit tests, and be based off of `jsonschema` validation. It will also implement an abstract class so that other validators can easily be utilized by extending this class.
- Allow the registering of a validator in the core library. This allows flexibility for users to implement their own validation logic.
- In the core library, create a `.validate` method on `STACObject`. This method will:
  - Check if there is a validator registered with PySTAC. If so, use that to validate the STACObject.
  - If no validator is registered, check to see if the `pystac_validator` library is importable. If so, use that validator. 
  - If no validator is registered and `pystac_validator` cannot be imported, throw an exception.

This way validation can happen on-demand for any `STACObject` by calling its `validate` method. 

This can be done in a way that lets catalogs be validated completely in a single call.

### Validate JSON

In addition to validating `STACObject`s, users may want to validate incoming JSON before they are read in as PySTAC objects. This could be done by a `pystac.validate(d)` method that will validate a dict based on the type and extensions identified in `pystac.serialization.identify`.

lossyrob:
This PR implements validation in PySTAC.

`Catalog`, `Collection` and `Item` have a `validate()` method that will validate their serialized versions against STAC schema json hosted at https://schemas.stacspec.org/. There's also a `pystac.validation.validate_json` method that can validate STAC from earlier versions, back to 0.6.0.

The design of this feature allows for users to write their own implementations of `STACValidator` and register it with PySTAC via the `pystac.validation.set_validator` method. Users can also override the `DefaultSchemaUriMap` to provide custom mappings between STAC object types and extensions to schema URIs.

This feature requires `jsonschema` to be installed if using the default validator. This can be accomplished by specifying the 'validation' extras on install:

```
> pip install pystac[validation]
```

Also included in this PR is moving the `urllib` capability for reading `http` and `https` URIs from the unit tests to the default implementations in `STAC_IO`, and so allowing `http` and `https` STAC to be read by default without a custom read implementation.

Fixes #118


pystac/validation/__init__.py
```diff
@@ -0,0 +1,109 @@
...
+    This method delegates to the call to :meth:`pystac.validation.STACValidator.validate`
+    for the STACValidator registered via :meth:`~pystac.validation.set_validator` or
+    :class:`~pystac.validation.JsonSchemaSTACValidator` by default.
+
+    Args:
+        stac_dict (dict): Dictionary that is the STAC json of the object.
+        stac_object_type (str): The stac object type of the object encoded in stac_dict.
+            One of :class:`~pystac.STACObjectType`.
+        stac_version (str): The version of STAC to validate the object against.
+        extensions (List[str]): Extension IDs for this stac object.
```
simonkassel, line 57: Consider combining theses two functions: allowing `stac_object_type` to be none and infer from dict. Change name to `validate_dict()`

Issue #148, opened 2020-08-17
Fix issue with retrieving docstring from ExtensionIndex


lossyrob:
Currently, if you execute 

```python
import pystac
catalog = pystac.Catalog(id='test', description='test')
help(catalog.ext)
```

an exception is thrown. This is because the `ExtensionIndex` class implements a custom `__getattr__` method to allow for property-like access of extension functionality.

The goal of this issue is to modify the ExtensionIndex so that the help function above succeeds, and to ensure that other "magic" Python functions avoid getting interpreted as extension IDs.

PR #159, opened 2020-08-18
Item Extension getattr Overload 

gwnoseworthy:
Modified the custom getattr to check for dunder attributes that exist in the object.  Falls back on __getattribute__. which allows help to run. Originally implimented using super() call but this feels slightly cleaner.


gwnoseworthy:
Fixes #148

PR #161, opened 2020-08-18
add timestamps extension

### LINKED ISSUE #135: Implement extension: Timestamps
[Timestamps Extension Specification](https://github.com/radiantearth/stac-spec/blob/v1.0.0-beta.1/extensions/timestamps/README.md)

simonkassel:
adds `timestamps` extension

closes #135 

One question I have is about the `apply` method in `TimestampsItemExt` (and this question could be relevant to other extensions as well): if you pass null values for optional arguments defining properties that are already defined in that item, should the null value override the existing property?

So for example:

`item` already has `published` and `expired` defined. 

if i do `item.ext.timestamps.apply(unpublished=d)`, this will convert those other two properties to None. Should this be the case?

### DIFF
--- a/pystac/extensions/timestamps.py
+++ b/pystac/extensions/timestamps.py
@@ -1,0 +1,161 @@
+import pystac
+from pystac import Extensions
+from pystac.extensions.base import (ExtendedObject, ExtensionDefinition, ItemExtension)
+from pystac.item import Item
+from pystac.utils import datetime_to_str, str_to_datetime
+
+
+class TimestampsItemExt(ItemExtension):
+    """TimestampsItemExt is the extension of an Item in that
+    allows to specify additional timestamps for assets and metadata.
+
+    Args:
+        item (Item): The item to be extended.
+
+    Attributes:
+        item (Item): The Item that is being extended.
+
+    Note:
+        Using TimestampsItemExt to directly wrap an item will add the 'timestamps'
+        extension ID to the item's stac_extensions.
+    """
+    def __init__(self, item):
+        if item.stac_extensions is None:
+            item.stac_extensions = [Extensions.TIMESTAMPS]
+        elif Extensions.TIMESTAMPS not in item.stac_extensions:
+            item.stac_extensions.append(Extensions.TIMESTAMPS)
+
+        self.item = item
+
+    @classmethod
+    def from_item(cls, item):
+        return cls(item)
+
+    @classmethod
+    def _object_links(cls):
+        return []
+
+    def apply(self, published=None, expires=None, unpublished=None):
+        """Applies timestamps extension properties to the extended Item.
+
+        Args:
+            published (datetime or None): Date and time the corresponding data
+                (see below) was published the first time.
+            expires (datetime or None): Date and time the corresponding data
+                (see below) expires (is not valid any longer).
+            unpublished (datetime or None): Date and time the corresponding data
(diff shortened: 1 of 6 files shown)


lossyrob:
> One question I have is about the apply method in TimestampsItemExt (and this question could be relevant to other extensions as well): if you pass null values for optional arguments defining properties that are already defined in that item, should the null value override the existing property?

That seems correct to me. I would expect that if `None` is passed in or if the param defaults to `None`, that would set the value accordingly

pystac/extensions/timestamps.py
```diff
@@ -0,0 +1,161 @@
...
+            self.item.properties[key] = timestamp
+        else:
+            timestamp_str = datetime_to_str(timestamp)
+            if asset is not None:
+                asset.properties[key] = timestamp_str
+            else:
+                self.item.properties[key] = timestamp_str
+
+    @property
+    def published(self):
```
lossyrob, line 79: I'd suggest putting a docstring that reads "get or sets ..." that includes the property description from the STAC spec, e.g. "Gets or sets the date and time the corresponding data (see below) was published the first time."

The convention is to doc both the property getter and setter in the @property getter: https://stackoverflow.com/questions/16025462/what-is-the-right-way-to-put-a-docstring-on-python-property

pystac/extensions/timestamps.py
```diff
@@ -0,0 +1,161 @@
...
+    @property
+    def expires(self):
+        return self.get_expires()
+
+    @expires.setter
+    def expires(self, v):
+        self.set_expires(v)
+
+    def get_expires(self, asset=None):
+        """Get an Item or Asset expires datetime
```
lossyrob, line 114: Since the [spec for timestamps](https://github.com/radiantearth/stac-spec/tree/v1.0.0-beta.2/extensions/timestamps#item-fields) states the fields have different meanings if the property is on items or on assets, consider adding some additional documentation about that in these methods that get and set the asset properties.

tests/extensions/test_timestamps.py
```diff
@@ -0,0 +1,42 @@
...
+from tests.utils import TestCases, test_to_from_dict
+
+
+class TimestampsTest(unittest.TestCase):
+    LANDSAT_EXAMPLE_URI = TestCases.get_path('data-files/timestamps/example-landsat8.json')
+
+    def setUp(self):
+        self.maxDiff = None
+
+    def test_to_from_dict(self):
```
lossyrob, line 16: Consider using the test layout that's used in other extension test, e.g. [test_view.py](https://github.com/azavea/pystac/blob/develop/tests/extensions/test_view.py), which makes a single test for each property and clearly tests the get and set. Ideally the layout for tests is consistent for each extension.

I'm also not sure how much value the test_to_from_dict is doing here - that is testing Item functionality which isn't the subject of these tests.

lossyrob:
Also note that you should add a CHANGELOG entry for this. Thanks!

tests/extensions/test_timestamps.py
simonkassel: Good call. I rewrote all the tests sticking to the `test_view.py` format. I left the `test_to_from_dict` in for consistency but can take it out if you want

lossyrob:
Should have mentioned this before, but can you add an `api.rst` doc entry for the extension?

simonkassel:
sure, will do

PR #176, opened 2020-08-21
Support for Pointcloud extensions 

chelm:
This PR adds support pointclouds as an item extension (https://github.com/radiantearth/stac-spec/tree/master/extensions/pointcloud). 

Example usage: 

```python
item.ext.enable(pystac.Extensions.POINTCLOUD)
item.ext.pointcloud.apply(1000000, 'lidar', 'laszip', ept['schema'])
```

This work was done as part of STAC Sprint #6 and all tests are passing. Open to any suggestions or changes on this as well.


pystac/extensions/pointcloud.py
```diff
@@ -0,0 +1,347 @@
...
+        """Get or sets a
+
+        The schemas represent the structure of the data attributes in the pointcloud,
+        and is represented as a sequential array of items that define the dimensions
+        and their types,
+
+        Returns:
+            List[dict]
+        """
+        return self.get_schemas()
```
lossyrob, line 177: Would it make sense to define a `SchemaObject` explicitly as a Python object, or do you think it would be more convenient to keep it as a dict? The [label extension](https://github.com/stac-utils/pystac/blob/v0.5.2/pystac/extensions/label.py#L300) has an example of returning objects from extensions. I think for consistency it might be better to define an object, however I'll defer to your thoughts on usability as the first primary user of the extension.

If it does stay as `dict`, perhaps the docstring should include an example of a valid schema dict to make it easier for users to see how they should be constructed (similar to the 'statistics' example below)

pystac/extensions/pointcloud.py
```diff
@@ -0,0 +1,347 @@
...
+        else:
+            asset.properties['pc:density'] = density
+
+    @property
+    def statistics(self):
+        """Get or sets the statistics for each property of the dataset.
+
+        A sequential array of items mapping to pc:schemas defines per-channel statistics.
+
+        Exmample::
```
lossyrob, line 254: Typo

pystac/extensions/pointcloud.py
```diff
@@ -0,0 +1,347 @@
...
+
+            item.ext.pointcloud.statistics = [{ 'name': 'red', 'min': 0, 'max': 255 }]
+
+        Returns:
+            List[dict]
```
lossyrob, line 259: Same comment as with the `schemas` property - consider using a specific object for so that users have clear guidance on how to construct or use what's returned here.

pystac/extensions/pointcloud.py
```diff
@@ -0,0 +1,347 @@
...
+        If an Asset is supplied, sets the property on the Asset.
+        Otherwise sets the Item's value.
+        """
+        if asset is None:
+            self.item.properties['pc:statistics'] = statistics
+        else:
+            asset.properties['pc:statistics'] = statistics
+
+    @property
+    def epsg(self):
```
lossyrob, line 293: EPSG is another extension - this may be leftover from using the `proj` extension as a template?

tests/extensions/test_pointcloud.py
```diff
@@ -0,0 +1,102 @@
...
+import unittest
+# from copy import deepcopy
+
+import pystac
+from pystac import (Item, Extensions)
+from pystac.extensions import ExtensionError
+from tests.utils import (TestCases, test_to_from_dict)
+
+
+class PointcloudTest(unittest.TestCase):
```
lossyrob, line 11: Missing property tests for density and statistics.

chelm:
Thanks @lossyrob for the comments. I'll get on them all this week as a i have time

pystac/extensions/pointcloud.py
chelm: yup this was my bad, proj extension handles this

tests/extensions/test_pointcloud.py
chelm: added

pystac/extensions/pointcloud.py
chelm: all fixed

pystac/extensions/pointcloud.py
chelm: I like this. Makes sense i added PointcloudSchema and PointcloudStatistic as classes and tested get/set on each of their props too.

pystac/extensions/pointcloud.py
chelm: add PointcloudStatistic to handle this.

chelm:
Ok i've resolved the comments and this extension is feeling pretty good. Sorry it took me so long to get to this. 

Now i just need to look at why the PR checks failed.

lossyrob (approved):

lossyrob:
Looks great, thanks!

PR #193, opened 2020-10-09
Add support for the version extension.

schwehr:
The initial version only implements version for Items.  I still need to do version for Collections.  Also, I'm sure more tests are needed.

https://github.com/radiantearth/stac-spec/tree/dev/extensions/version

### DIFF
--- a/pystac/extensions/version.py
+++ b/pystac/extensions/version.py
@@ -88,7 +88,7 @@ class VersionItemExt(base.ItemExtension):
 
     @classmethod
     def _object_links(cls):
-        return []  # TODO(schwehr): What should this return?
+        return [LATEST_VERSION, PREDECESSOR_VERSION, SUCCESSOR_VERSION]
 
 
 class VersionCollectionExt(base.CollectionExtension):
@@ -150,7 +150,7 @@ class VersionCollectionExt(base.CollectionExtension):
 
     @classmethod
     def _object_links(cls):
-        return []  # TODO(schwehr): What should this return?
+        return [LATEST_VERSION, PREDECESSOR_VERSION, SUCCESSOR_VERSION]
 
     def apply(self, version, deprecated=None, latest=None, predecessor=None, successor=None):
         self.version = version
--- a/pystac/extensions/version.py
+++ b/pystac/extensions/version.py
@@ -4,6 +4,7 @@ https://github.com/radiantearth/stac-spec/tree/dev/extensions/version
 """
 
 import pystac
+from pystac import collection
 from pystac import Extensions
 from pystac import item
 from pystac import link
@@ -90,7 +91,80 @@ class VersionItemExt(base.ItemExtension):
         return []  # TODO(schwehr): What should this return?
 
 
-# TODO(schwehr): class VersionCollectionExt(base.CollectionExtension):
+class VersionCollectionExt(base.CollectionExtension):
+    """Add an asset version string to a STAC Collection."""
+    def __init__(self, a_collection):
+        self.collection = a_collection
 
-VERSION_EXTENSION_DEFINITION = base.ExtensionDefinition(
-    Extensions.VERSION, [base.ExtendedObject(item.Item, VersionItemExt)])
+    @property
+    def version(self):
+        return self.collection.extra_fields.get('version')
+
+    @version.setter
+    def version(self, v):
+        self.collection.extra_fields['version'] = v
+
+    @property
+    def deprecated(self):
+        return bool(self.collection.extra_fields.get('deprecated'))
+
+    @deprecated.setter
+    def deprecated(self, v):
+        if not isinstance(v, bool):
+            raise pystac.STACError('deprecated must be a bool')
+        self.collection.extra_fields['deprecated'] = v
+
+    @property
+    def latest_link(self):
+        links = self.collection.get_links(LATEST_VERSION)
+        if links:
+            return links[0]
+
+    @latest_link.setter
+    def latest_link(self, source_collection):
+        self.collection.add_link(link.Link(LATEST_VERSION, source_collection, MEDIA_TYPE))
+
+    @property
+    def predecessor_link(self):
+        links = self.collection.get_links(PREDECESSOR_VERSION)
+        if links:
+            return links[0]
+
+    @predecessor_link.setter
+    def predecessor_link(self, source_collection):
+        self.collection.add_link(link.Link(PREDECESSOR_VERSION, source_collection, MEDIA_TYPE))
+
+    @property
+    def successor_link(self):
+        links = self.collection.get_links(SUCCESSOR_VERSION)
+        if links:
+            return links[0]
+
+    @successor_link.setter
+    def successor_link(self, source_collection):
+        self.collection.add_link(link.Link(SUCCESSOR_VERSION, source_collection, MEDIA_TYPE))
+
+    @classmethod
+    def from_collection(cls, a_collection):
+        return cls(a_collection)
+
+    @classmethod
+    def _object_links(cls):
+        return []  # TODO(schwehr): What should this return?
+
+    def apply(self, version, deprecated=None, latest=None, predecessor=None, successor=None):
+        self.version = version
+        if deprecated is not None:
+            self.deprecated = deprecated
+        if latest:
+            self.latest_link = latest
+        if predecessor:
+            self.predecessor_link = predecessor
+        if successor:
+            self.successor_link = successor
+
+
+VERSION_EXTENSION_DEFINITION = base.ExtensionDefinition(Extensions.VERSION, [
+    base.ExtendedObject(item.Item, VersionItemExt),
+    base.ExtendedObject(collection.Collection, VersionCollectionExt)
+])
--- a/pystac/extensions/version.py
+++ b/pystac/extensions/version.py
@@ -10,9 +10,9 @@ from pystac import item
 from pystac import link
 from pystac.extensions import base
 
-LATEST_VERSION = 'latest-version'
-PREDECESSOR_VERSION = 'predecessor-version'
-SUCCESSOR_VERSION = 'successor-version'
+LATEST = 'latest-version'
+PREDECESSOR = 'predecessor-version'
+SUCCESSOR = 'successor-version'
 
 # Media type for links.
 MEDIA_TYPE = 'application/json'
@@ -54,33 +54,33 @@ class VersionItemExt(base.ItemExtension):
 
     @property
     def latest(self):
-        links = self.item.get_links(LATEST_VERSION)
+        links = self.item.get_links(LATEST)
         if links:
             return links[0]
 
     @latest.setter
     def latest(self, source_item):
-        self.item.add_link(link.Link(LATEST_VERSION, source_item, MEDIA_TYPE))
+        self.item.add_link(link.Link(LATEST, source_item, MEDIA_TYPE))
 
     @property
     def predecessor(self):
-        links = self.item.get_links(PREDECESSOR_VERSION)
+        links = self.item.get_links(PREDECESSOR)
         if links:
             return links[0]
 
     @predecessor.setter
     def predecessor(self, source_item):
-        self.item.add_link(link.Link(PREDECESSOR_VERSION, source_item, MEDIA_TYPE))
+        self.item.add_link(link.Link(PREDECESSOR, source_item, MEDIA_TYPE))
 
     @property
     def successor(self):
-        links = self.item.get_links(SUCCESSOR_VERSION)
+        links = self.item.get_links(SUCCESSOR)
         if links:
             return links[0]
 
     @successor.setter
     def successor(self, source_item):
-        self.item.add_link(link.Link(SUCCESSOR_VERSION, source_item, MEDIA_TYPE))
+        self.item.add_link(link.Link(SUCCESSOR, source_item, MEDIA_TYPE))
 
     @classmethod
     def from_item(cls, an_item):
@@ -88,7 +88,7 @@ class VersionItemExt(base.ItemExtension):
 
     @classmethod
     def _object_links(cls):
-        return [LATEST_VERSION, PREDECESSOR_VERSION, SUCCESSOR_VERSION]
+        return [LATEST, PREDECESSOR, SUCCESSOR]
 
 
 class VersionCollectionExt(base.CollectionExtension):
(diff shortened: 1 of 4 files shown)


schwehr:
I will fix this:

```
flake8 pystac tests
pystac/extensions/version.py:24:7: E111 indentation is not a multiple of four
```

schwehr:
I will try to fix the next linter's complaints later today.

schwehr:
I need to add a test that validates an item created with a version

schwehr:
I will add validation code like this when I get a chance.

```python
  def testValidateMinimum(self):
    with self.assertRaises(pystac.validation.STACValidationError):
      self.item.validate()

    self.item.ext.version.apply(self.version)
    self.item.validate()

  def testValidateAll(self):
    deprecated = True
    latest = MakeItem(2013)
    predecessor = MakeItem(2010)
    successor = MakeItem(2012)
    self.item.ext.version.apply(self.version, deprecated, latest, predecessor, successor)

    self.item.validate()
```

It appears that I need to add geometry to the item for it to be valid. (or remove the bbox and geometry)

```python
def MakeItem(year):
  asset_id = f'USGS/GAP/CONUS/{year}'
  bbox = [-160.26, 18.85, -154.66, 22.29]
  geometry = {
      'type': 'Polygon',
     # TODO(schwehr): Match coords
      'coordinates': [[
          [100.0, 0.0],
          [101.0, 0.0],
          [101.0, 1.0],
          [100.0, 1.0],
          [100.0, 0.0]]]}
  start = datetime.datetime(year, 1, 2)

  item = pystac.Item(
      id=asset_id, geometry=geometry, bbox=bbox, datetime=start, properties={})
  item.set_self_href(URL_TEMPLATE % year)

  item.ext.enable(pystac.Extensions.VERSION)

  return item
```

schwehr:
Probably best to add a test where the deprecated arg to apply was not a bool or None, but I can do that after the rest of version gets written.

tests/extensions/test_version.py
```diff
@@ -0,0 +1,210 @@
...
+
+    def setUp(self):
+        super().setUp()
+        self.item = MakeItem(2011)
+
+        self.item.ext.enable(pystac.Extensions.VERSION)
+
+    def tearDown(self):
+        super().tearDown()
+        self.item.validate()
```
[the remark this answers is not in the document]
schwehr: No, this is bad.  I was trying to avoid so much duplication in the tests.  Looking at the docs for tearDown, I shouldn't do this.  https://docs.python.org/3/library/unittest.html#unittest.TestCase.tearDown

> Method called immediately after the test method has been called and the result recorded.

pystac/extensions/version.py
```diff
@@ -0,0 +1,170 @@
...
+    def successor_link(self, source_item):
+        self.item.add_link(link.Link(SUCCESSOR_VERSION, source_item, MEDIA_TYPE))
+
+    @classmethod
+    def from_item(cls, an_item):
+        return cls(an_item)
+
+    @classmethod
+    def _object_links(cls):
+        return []  # TODO(schwehr): What should this return?
```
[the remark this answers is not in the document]
lossyrob: My bad, forgot this was a classmethod. It should return the `rel` types that point to stac objects, not the actual links.

So `return [SUCCESSOR_VERSION, PREDECESSOR_VERSION, LATEST_VERSION]` should work for this case.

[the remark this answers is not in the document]
schwehr: Is there a reasonably direct way to test this?  I switched the order the be alphabetical.  If there was a specific reason for the order you listed, I will change it.

I'm going to look at the catalog copy test now.

schwehr:
Doh.  Committed the wrong thing, so I backed up and forced the commit.

pystac/extensions/version.py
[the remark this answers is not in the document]
lossyrob: No reason for the ordering, so alphabetical is great!

Not a very direct test; the `full_copy` logic is really what this is used for so the more complicated test is what will really ensure its working.

tests/extensions/test_version.py
```diff
@@ -0,0 +1,210 @@
...
+
+        self.item.ext.version.apply(self.version)
+
+    def testAllLinks(self):
+        deprecated = True
+        latest = MakeItem(2013)
+        predecessor = MakeItem(2010)
+        successor = MakeItem(2012)
+        self.item.ext.version.apply(self.version, deprecated, latest, predecessor, successor)
+        # Just validate in tearDown.
```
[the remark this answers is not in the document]
schwehr: What I have so far:

```
    def test_copy_1_with_version(self):
        c2011 = make_collection(2011)
        c2012 = make_collection(2012)
        c2013 = make_collection(2013)
        c2014 = make_collection(2014)

        c2012.ext.enable(pystac.Extensions.VERSION)
        c2012.ext.version.apply('v2012', latest=c2012, predecessor=c2011, successor=c2013)

        cat = pystac.Catalog('my_catalog', 'my description')
        cat.add_child(c2011)
        cat.add_child(c2012)
        cat.add_child(c2013)
        cat.add_child(c2014)

        cat_copy = cat.full_copy()
```

Was trying to figure out how best to compare the two.  Hopefully I can think through it easier where there aren't mindcraft videos playing next to me... :|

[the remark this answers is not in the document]
schwehr: Having a hard time figuring out the item link thing.  This is what I had as of the end of yesterday

```#!python
def compare_collection(col1, col2):
  if col1.id != col2.id:
    return False
  if col1.description != col2.description:
    return False
  has_ver_1 = col1.ext.implements(pystac.Extensions.VERSION)
  has_ver_2 = col2.ext.implements(pystac.Extensions.VERSION)
  if has_ver_1 != has_ver_2:
    return False
  if has_ver_1:
    print('has_ver')
    ver1 = col1.ext.version
    ver2 = col2.ext.version
    if ver1.version != ver2.version:
      return False
    if ver1.deprecated != ver2.deprecated:
      return False
    if str(ver1.latest_link) != str(ver2.latest_link):
      return False
    if str(ver1.predecessor_link) != str(ver2.predecessor_link):
      return False
    if str(ver1.successor_link) != str(ver2.successor_link):
      return False
  return True


class FullCopyTest(googletest.TestCase):

  def compare_catalogs(self, cat1, cat2):
    self.assertEqual(cat1.id, cat2.id)
    children1 = {child.id: child for child in cat1.get_children()}
    # print('children1 dict', children1)
    children2 = {child.id: child for child in cat2.get_children()}
    # print('children2 dict', children2)
    self.assertCountEqual(sorted(children1.keys()), sorted(children2.keys()))
    for child_id in children1:
      print(child_id)
      child1 = children1[child_id]
      child2 = children2[child_id]
      has_ver_1 = child1.implements(pystac.Extensions.VERSION)
      has_ver_2 = child1.implements(pystac.Extensions.VERSION)
      self.assertEqual(has_ver_1, has_ver_2)
      if has_ver_1:
        # Umm... ?
        self.assertTrue(compare_collection(child1, child2))
        link = child1.get_links(version.LATEST_VERSION)[0].resolve_stac_object()
        self.assertIs(link.target, child2)

    return True

  def testCopy1Linked(self):
    c2011 = make_collection(2011)
    c2012 = make_collection(2012)
    c2013 = make_collection(2013)
    c2014 = make_collection(2014)

    c2012.ext.enable(pystac.Extensions.VERSION)
    c2012.ext.version.apply(
        'v2012', latest=c2014, predecessor=c2011, successor=c2013)

    cat = pystac.Catalog('my_catalog', 'my description')
    cat.add_child(c2011)
    cat.add_child(c2012)
    cat.add_child(c2013)
    cat.add_child(c2014)

    cat_copy = cat.full_copy()
    pprint(cat_copy.to_dict())

    self.compare_catalogs(cat, cat_copy)
```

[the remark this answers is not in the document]
lossyrob: The easier thing would be to do a straight object identity test to ensure the object cache ensures there's a single object that's linked to in the catalog and by the item that has the version link.

It might be easier/more clear to make the suggestion in code. I can take a crack at this tomorrow if you're alright with me pushing a commit to this branch?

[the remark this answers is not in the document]
schwehr: Go for it!  The only things I really care about is learning how it's done and getting the extensions implemented.  I don't like what I did above except for the setup (assuming that it makes input of the form required.)

tests/extensions/test_version.py
```diff
@@ -104,6 +105,39 @@ def test_all_links(self):
         self.item.ext.version.apply(self.version, deprecated, latest, predecessor, successor)
         self.item.validate()
 
+    def test_full_copy(self):
```
lossyrob, line 108: @schwehr here are the two tests that I used to make sure the object links were operating correctly in a full_copy scenario.

Notice that if you don't return the correct reltypes from the `_object_link` method, these tests fail.

This is because, when making a full copy of the catalog in memory, PySTAC doesn't know that those links point to STAC objects that should be considered in the copy. In [this bit of code](https://github.com/stac-utils/pystac/blob/develop/pystac/stac_object.py#L413-L427), the logic to determine whether or not a link has to be copied (and it's target either copied or pulled from the cache of already-copied objects) is determined by the _object_links method, which [as seen here](https://github.com/stac-utils/pystac/blob/develop/pystac/item.py#L265-L266) depends on object link the Catalog or Item declares (core reltypes like child, item or collection), as well as any reltypes that are declared to be STAC objects by implemented extensions.

Let me know if this helps explain the _object_links method and why this is a good way to test that, or feel free to make changes or give ideas on way to improve!

schwehr: Arg... I'm a noob and forced commit over your change.  I'll recommit it from this :(

lossyrob: No worries! I find `git cherry-pick` to be the most useful in those types of situations.

pystac/extensions/version.py
```diff
@@ -0,0 +1,170 @@
...
+        return bool(self.item.properties.get('deprecated'))
+
+    @deprecated.setter
+    def deprecated(self, v):
+        if not isinstance(v, bool):
+            raise pystac.STACError('deprecated must be a bool')
+        self.item.properties['deprecated'] = v
+
+    @property
+    def latest_link(self):
```
lossyrob, line 56: I find the `_link` part of the method names and the differing setter values and return values a bit confusing. We aren't passing links in the setter, but the getter returns links... IMO it'd be better to have these property methods as `latest`, `predecessor`, and `successor`, similarly to how they appear in the `apply` parameter list, and then have them both get and set the target object instead of the link.

lossyrob (changes requested):

pystac/extensions/version.py
schwehr: Agreed.  Done in b35cd64.

I will also remove `_VERSION` from the constants that hold the field names  So `version.LATEST_VERSION` → `version.LATEST`.  `version.LATEST_VERSION` is painfully redundant.

lossyrob: Sounds good.

Looks like that commit still returns `Link` objects from the getters. To make things consistent, I think we should return the Item or Collection from those getters (similarly to how that is the value passed in), e.g.:

```python
@property
    def latest(self):
        links = self.item.get_links(LATEST_VERSION)
        if links:
            return links[0].resolve_stac_object(root=self.item.get_root()).target
```

the `resolve_stac_object` call on the link will make sure that the target is the relevant PySTAC object and not a link HREF. The `root` argument is a bit clunky, and I think the API needs to change around the link resolution code, but for now that should do the right thing.

lossyrob: Actually, potentially cleaner to use this method: https://github.com/stac-utils/pystac/blob/9944c5178b97333ceb5ff74b0d8a516acce5f22d/pystac/stac_object.py#L346

a la 

https://github.com/stac-utils/pystac/blob/develop/pystac/extensions/label.py#L261

schwehr: Done in https://github.com/stac-utils/pystac/pull/193/commits/d408d2e24b91b9d444b6f1285363c017e13fd7a5.  Also make VERSION and DEPRECATED constants

tests/extensions/test_version.py
schwehr: Thanks!  I spend most of my time in [piper](https://cacm.acm.org/magazines/2016/7/204032-why-google-stores-billions-of-lines-of-code-in-a-single-repository/fulltext) where cherry-picks are usually just for release branches.  Appreciate any help leveling up my git skills.

pystac/extensions/version.py
```diff
@@ -0,0 +1,172 @@
...
+
+    @predecessor.setter
+    def predecessor(self, source_item):
+        self.item.add_link(link.Link(PREDECESSOR, source_item, MEDIA_TYPE))
+
+    @property
+    def successor(self):
+        items = self.item.get_stac_objects(SUCCESSOR)
+        if items:
+            return next(items)
```
lossyrob, line 81: One last little nitpick -

Making the return in the None case would be more explicit here and elsewhere, rather than not returning and relying on Python's default mechanism to serve None back to the caller.

Here and in the other iterator cases you could combine these lines into:

```suggestion
       return next(items, None)
```

I think it'd still be readable if you even one-lined it:

```
return next(self.item.get_stac_objects(SUCCESSOR), None)
```

lossyrob (changes requested):
One small, more stylistic change - feel free to push back on it if you're used to that method :-)

Other then that this is looking :+1: and I'll merge after we sort out this last bit. Thank you!

lossyrob:
One more thing I forgot to mention - the next extension needs an entry in the API docs. We use autodoc but it requires we specify the class, and then give it a heading so it appears nicely in the TOC. See [this for an example](https://github.com/stac-utils/pystac/blob/develop/docs/api.rst#projection-extension).

pystac/extensions/version.py
schwehr: Awesome!  I haven't looked at next() in ages.  The default `None` is so much simpler.

schwehr:
> One more thing I forgot to mention - the next extension needs an entry in the API docs. We use autodoc but it requires we specify the class, and then give it a heading so it appears nicely in the TOC. See [this for an example](https://github.com/stac-utils/pystac/blob/develop/docs/api.rst#projection-extension).

Cool.  Done it fbd7abd.  The order in the API doc is a bit confusing.  I did fix up some complains from the tooling about the Point Cloud Extension.

lossyrob (approved):

lossyrob:
Thanks for fixing the pointcloud docs, I hadn't noticed that that had been breaking the doc build.

Looks great, merging!

PR #195, opened 2020-10-13
Codespell 20201013

schwehr:


lossyrob (approved):
Thanks for this, much needed.

One way to prevent such horrible misspellings (all of them mine) in the future is to run codespell as part of CI testing. Any ignore words could be put into an ignore file that we also commit. Any thoughts on that?

Issue #194, opened 2020-10-11
Support the scientific extension


schwehr:
In follow up to #193, I'm doing to work on supporting the Scientific Extension.

https://github.com/radiantearth/stac-spec/tree/dev/extensions/scientific

duckontheweb:
Hi @schwehr , just curious if there is a timeline for getting this into a new release? We're working with collections that implement this extension and it would be helpful to be able to access the properties using `collection.ext['scientific']` rather than getting them via the dictionary.

lossyrob:
@duckontheweb this issue was closed by #199, and released as part of pystac 0.5.4. Let us know how the extension works for you and submit issues for any problems you hit!

schwehr:
Thanks @lossyrob for closing this.  Sorry I didn't.

duckontheweb:
> @duckontheweb this issue was closed by #199, and released as part of pystac 0.5.4. Let us know how the extension works for you and submit issues for any problems you hit!

Awesome, thanks! I'll upgrade (I was on `0.5.3`).

PR #199, opened 2020-10-14
Add support for the scientific extension.

schwehr:
https://github.com/radiantearth/stac-spec/tree/dev/extensions/scientific

This is for issue #194.  I will be tracking the changes to PR #193 to improve this extension code.

### DIFF
--- a/pystac/extensions/scientific.py
+++ b/pystac/extensions/scientific.py
@@ -57,8 +57,8 @@ class Publication:
         return pystac.Link(CITE_AS, doi_to_url(self.doi))
 
 
-def remove_link(links: List[pystac.Link], pub: Publication):
-    url = doi_to_url(pub.doi)
+def remove_link(links: List[pystac.Link], doi: str):
+    url = doi_to_url(doi)
     for i, a_link in enumerate(links):
         if a_link.rel != CITE_AS:
             continue
@@ -123,6 +123,11 @@ class ScientificItemExt(base.ItemExtension):
 
     @doi.setter
     def doi(self, v: str) -> None:
+        if DOI in self.item.properties:
+            if v == self.item.properties[DOI]:
+                return
+            remove_link(self.item.links, self.item.properties[DOI])
+
         self.item.properties[DOI] = v
         url = doi_to_url(self.doi)
         self.item.add_link(pystac.Link(CITE_AS, url))
@@ -167,13 +172,13 @@ class ScientificItemExt(base.ItemExtension):
 
         if not publication:
             for one_pub in self.item.ext.scientific.publications:
-                remove_link(self.item.links, one_pub)
+                remove_link(self.item.links, one_pub.doi)
 
             del self.item.properties[PUBLICATIONS]
             return
 
         # One publication and link to remove
-        remove_link(self.item.links, publication)
+        remove_link(self.item.links, publication.doi)
         to_remove = publication.to_dict()
         self.item.properties[PUBLICATIONS].remove(to_remove)
 
@@ -237,6 +242,10 @@ class ScientificCollectionExt(base.CollectionExtension):
 
     @doi.setter
     def doi(self, v: str) -> None:
+        if DOI in self.collection.extra_fields:
+            if v == self.collection.extra_fields[DOI]:
+                return
+            remove_link(self.collection.links, self.collection.extra_fields[DOI])
         self.collection.extra_fields[DOI] = v
         url = doi_to_url(self.doi)
         self.collection.add_link(pystac.Link(CITE_AS, url))
@@ -283,13 +292,13 @@ class ScientificCollectionExt(base.CollectionExtension):
 
         if not publication:
             for one_pub in self.collection.ext.scientific.publications:
-                remove_link(self.collection.links, one_pub)
+                remove_link(self.collection.links, one_pub.doi)
 
             del self.collection.extra_fields[PUBLICATIONS]
             return
 
         # One publication and link to remove
-        remove_link(self.collection.links, publication)
+        remove_link(self.collection.links, publication.doi)
         to_remove = publication.to_dict()
         self.collection.extra_fields[PUBLICATIONS].remove(to_remove)
 
--- a/pystac/extensions/scientific.py
+++ b/pystac/extensions/scientific.py
@@ -41,6 +41,12 @@ class Publication:
         self.doi = doi
         self.citation = citation
 
+    def __eq__(self, other):
+        if not isinstance(other, Publication):
+            return NotImplemented
+
+        return self.doi == other.doi and self.citation == other.citation
+
     def __repr__(self):
         return f'<Publication doi={self.doi} target={self.citation}>'
 
@@ -97,7 +103,7 @@ class ScientificItemExt(base.ItemExtension):
 
     @property
     def publications(self):
-        return self.item.properties.get(PUBLICATIONS)
+        return [Publication.from_dict(pub) for pub in self.item.properties.get(PUBLICATIONS)]
 
     @publications.setter
     def publications(self, v):
@@ -147,7 +153,7 @@ class ScientificCollectionExt(base.CollectionExtension):
 
     @property
     def publications(self):
-        return self.collection.extra_fields.get(PUBLICATIONS)
+        return [Publication.from_dict(pub) for pub in self.collection.extra_fields.get(PUBLICATIONS)]
 
     @publications.setter
     def publications(self, v):
--- a/pystac/extensions/scientific.py
+++ b/pystac/extensions/scientific.py
@@ -1,22 +1,21 @@
 """Implement the scientific extension.
 
-DOI Handbook: https://doi.org/10.1000/182
-
 https://github.com/radiantearth/stac-spec/tree/dev/extensions/scientific
+
+For a description of Digital Object Identifiers (DOIs), see the DOI Handbook:
+
+https://doi.org/10.1000/182
 """
 
 import copy
-import re
 from typing import Dict, List, Optional, TypeVar
 from urllib import parse
 
 import pystac
-from pystac import collection
 from pystac import Extensions
-from pystac import item
-from pystac import link
 from pystac.extensions import base
 
+# STAC fields strings.
 PREFIX: str = 'sci:'
 DOI: str = PREFIX + 'doi'
 CITATION: str = PREFIX + 'citation'
@@ -25,18 +24,11 @@ PUBLICATIONS: str = PREFIX + 'publications'
 # Link type.
 CITE_AS: str = 'cite-as'
 
-# TODO(schwehr): What is the correct regex for doi?
-# https://github.com/radiantearth/stac-spec/issues/910
-DOI_REGEX = r'10[.][0-9]{4}([.][0-9]+)*/.+'
 DOI_URL_BASE = 'https://doi.org/'
 
 _PublicationType = TypeVar('Publication')
 _ScientificItemExtType = TypeVar('ScientificItemExt')
 _ScientificCollectionExtType = TypeVar('ScientificCollectionExtType')
-
-
-def is_doi_valid(doi: str) -> bool:
-    return bool(re.match(DOI_REGEX, doi))
 
 
 def doi_to_url(doi: str) -> str:
@@ -65,11 +57,11 @@ class Publication:
     def from_dict(d: Dict[str, str]) -> _PublicationType:
         return Publication(d['doi'], d['citation'])
 
-    def get_link(self) -> link.Link:
-        return link.Link(CITE_AS, doi_to_url(self.doi))
+    def get_link(self) -> pystac.Link:
+        return pystac.Link(CITE_AS, doi_to_url(self.doi))
 
 
-def remove_link(links: List[link.Link], pub: Publication):
+def remove_link(links: List[pystac.Link], pub: Publication):
     url = doi_to_url(pub.doi)
     for i, a_link in enumerate(links):
         if a_link.rel != CITE_AS:
@@ -80,14 +72,35 @@ def remove_link(links: List[link.Link], pub: Publication):
 
 
 class ScientificItemExt(base.ItemExtension):
-    """Add an citation and dois to a STAC Item."""
-    def __init__(self, an_item: item.Item) -> None:
+    """ScientificItemExt extends Item to add citations and DOIs to a STAC Item.
+
+    Args:
+        item (Item): The item to be extended.
+
+    Attributes:
+        item (Item): The item that is being extended.
+
+    Note:
+        Using ScientificItemExt to directly wrap an item will add the 'scientific'
+        extension ID to the item's stac_extensions.
+    """
+    item: pystac.Item
+
+    def __init__(self, an_item: pystac.Item) -> None:
         self.item = an_item
 
     def apply(self,
               doi: Optional[str] = None,
               citation: Optional[str] = None,
               publications: Optional[List[Publication]] = None) -> None:
+        """Applies scientific extension properties to the extended Item.
+
+        Args:
+            doi (str): Optional DOI string for the item.  Must not be a DOI link.
+            citation (str): Optional human-readable reference.
+            publications (List[Publication]): Optional list of relevant publications
+                referencing and describing the data.
+        """
         if doi:
             self.doi = doi
         if citation:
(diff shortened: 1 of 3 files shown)


pystac/extensions/scientific.py
```diff
@@ -0,0 +1,162 @@
...
+    def citation(self):
+        return self.item.properties.get(CITATION)
+
+    @citation.setter
+    def citation(self, v):
+        self.item.properties[CITATION] = v
+
+    @property
+    def publications(self):
+        return self.item.properties.get(PUBLICATIONS)
```
lossyrob, line 100: The `publications` property takes a list of  `Publication`s, but the getter returns a dict-encoding of them. I'd suggest keeping this consistent in dealing with the Public class, so to deserialize these in the getter using `Publication.from_dict`

pystac/extensions/scientific.py
```diff
@@ -0,0 +1,162 @@
...
+
+    @publications.setter
+    def publications(self, v):
+        self.item.properties[PUBLICATIONS] = [pub.to_dict() for pub in v]
+        for pub in v:
+            self.item.add_link(pub.get_link())
```
lossyrob, line 106: As this has some logic to add a link for a publication, publications would be harder to remove. e.g. if for some reason I wanted to clear the publications, I couldn't just use `item.ext.scientific.publications = []`.

Perhaps we should have a method for that, something like:

```python
def remove_publication(self, publication=None):  # Docs specify using None will clear all
  pass # Logic would also clear any publication links
```

lossyrob (changes requested):
Looking good. Couple comments, and also can you add this extension to the API documentation?

pystac/extensions/scientific.py
schwehr: Done in 814c5f5

pystac/extensions/scientific.py
schwehr: Done in 814c5f5

pystac/extensions/scientific.py
```diff
@@ -0,0 +1,272 @@
...
+    def citation(self) -> str:
+        return self.item.properties.get(CITATION)
+
+    @citation.setter
+    def citation(self, v: str) -> None:
+        self.item.properties[CITATION] = v
+
+    @property
+    def publications(self) -> List[Publication]:
+        return [Publication.from_dict(pub) for pub in self.item.properties.get(PUBLICATIONS, [])]
```
lossyrob, line 164: Didn't catch this in my last review, but:

The way that Publications are handled here breaks a mutability pattern in PySTAC.

For instance, when I do

```python
item.ext.scientific.doi = "something"
print(item.ext.scientific.doi)
```

I would get the expected result, "something". This is true even though the extension wrapper object is created twice - once for each call of `item.ext.scientific`.

As it is currently written, the same wouldn't hold true for publications. For example, say I had an item with a single publication that had a citation of "everyone et al". The following:

```python
item.ext.scientific.publications[0].citation = "none et al"
print(item.ext.scientific.publications[0].citation
```

would print "everyone et al" - since the Publication instance that was returned was a temporary object, setting it's citation didn't have consequence.

This could be worked around by requiring the user always get and set publications, e.g.

```python
pubs = item.ext.scientific.publications
pubs[0].citation = "none et al"
item.ext.scientific.publications = pubs
```

but this is burdensome, and surely users will forget to do this. Also, it doesn't match the expectation of mutability that's present elsewhere in the library, including the Ext classes that wrap an item rather than contain data of their own.

The way around this is to keep the wrapper train rolling - take [LabelOverview](https://github.com/stac-utils/pystac/blob/e105c4cb0bdd5f1273b85fda9828edce80b714c4/pystac/extensions/label.py#L380) for instance. It takes as a constructor the dict from the parent Item's properties and uses getters and setters to manipulate the data directly on the Item rather than hold its own state.

I think we should follow that pattern here and provide similar wrappers around Item state with the Publications  and DOI objects.

lossyrob (changes requested):

pystac/extensions/scientific.py
schwehr: I'll take a look at this today.  Now that you say this, I realize that version is likely broken.  Created #210.

schwehr: I'm having trouble thinking through how best to do this.  I'd like Publication and Doi instances to be able to live outside of an Item or Collection.  What should a Publication track?  I see a couple possibilities with it being optional if these are attached:

* The Item or Collection
* The properties/extra_fields and links
* The specific entries in properties/extra_fields and the link

Having a reference to the specific dict in `sci:publications` and in links could make changing the doi or citation fairly easy to do, but removing publications from an Item/Collection becomes trouble, yes?

I do worry about things like people taking a publication for one item and then adding that publication to all the other items in a collection and trying to permute them to match that item.  If there are references inside the publication, then all the copies will end up changing together.  And now I'm thinking which things should have deepcopies and which shouldn't getting all turned around.

I don't have an understand of what the overall strategy is for mutations.

lossyrob: I'd suggest wrapping the dictionary of properties that make up the Publication, i.e. the __init__ takes a JSON serialized version of the Publication. In other cases, I have a `create` classmethod which acts as an overloaded constructor that takes the same params as `apply` and allows the objects to live outside of an Item (e.g. [how it's done with Bands](https://github.com/stac-utils/pystac/blob/1fcfb44220965f230e73f89ee9a974b346e67351/tests/extensions/test_eo.py#L37))

I'm not sure that removing publications is too onerous, the logic in the removal would have to just account for the fact it's removing a dict from a list rather than an object from the list.

The scenario you mention is problematic, and a good flag to raise. If you were to add the publication from one item to another and change the publication instance, that change would appear in both places. This could lead to unexpected behavior. The way I can see coding around this case is to have these objects track their owners, much like a link or asset tracks their owners. That way, when a publication owned by one item is attached to another, either the ownership can switch or a deepcopy made.

It does seem to me that this scenario is less likely to cause confusion than if the publications become detached and are un-editable unlike the rest of the library. I think if we're clear that things are highly mutable, users should be careful about reusing instances set on many different objects and then editing those instances, and use a copy themselves if they want a fresh copy that won't be mutated by the original instance. It seems similar to the scenario of  a user creating a single `dict` and assigning it to many Item instance's `properties` via the constructor, and then mutating the properties. So, the scenario you outline is problematic, but I think is something we can live with. 

I'm not sure what the overall strategy for mutations is, but I'll try to comment on my perspective around mutability in PySTAC. Mutability always comes with risk of state modification when you don't want it, and ideally all things would be strictly immutable. Any change would give you back a new copy of the thing you changed, with the change you requested. However, this gets complex pretty quickly in STAC. Imaging you've read in a large catalog. Any change in that catalog would produce a new catalog with the relevant edits. If I modify a single property of an Item, then I need a new Item. I then need all of the links to that Item changed, and so on - what I really would produce is an entirely new STAC. Things get more complicated when you consider that PySTAC uses caching to enable links to the same STAC object on the filesystem to refer to the same STAC object in memory. 

Immutability is possible, and there are patterns like Lenses to make complex hierarchical immutable data structures easier to transform. When I was setting things up in PySTAC, I certainly considered this. However, Python being the dynamic, mutation-friendly language that it is, that felt like it would be swimming against the current of what users would want. I think of STACs as graphs, so I looked at the most popular graph library - networkx - to take direction from, which has mutable data structures (as an aside, I found this [interesting conversation about mutable graphs](https://github.com/dmlc/dgl/issues/33) that is related).

So to me, if we're not trying to be immutable, then we can lean into mutability and use it to our advantage, like how we allow multiple extension objects act as views into the STAC object properties and mutate it, so that the extension objects are essentially stateless. We can also ask the user to assume mutability and understand that there's risk in assigning the same instance across multiple STAC objects.

lossyrob: Looking at the code, I realized that `Doi` is not an object, but a wrapper around a string. This would be fine to keep as an object, though I think it may get awkward to do things like

```
item.ext.scientific.doi.doi
```
to get at the string. Perhaps consider moving the logic into the scientific extensions? E.g. I could imagine validation logic and the URL being directly on ScientificItemExt, like a `doi_url` property that is only a getter, and a `validate_doi` method that is called on the doi setter, which would take a string.

For `Publication`, the above comments hold - this is a object that would be represented as a dict inside the item properties.

pystac/extensions/scientific.py
```diff
@@ -0,0 +1,272 @@
...
+
+    @publications.setter
+    def publications(self, v: List[Publication]) -> None:
+        self.item.properties[PUBLICATIONS] = [pub.to_dict() for pub in v]
+        for pub in v:
+            self.item.add_link(pub.get_link())
```
lossyrob, line 170: I think I may have been mistaken if I commented on publication links - the spec does not say that publications should be in the `links` of an item, only the item DOI link. This is added above in the doi setter. With this logic we have links that are marked as 'cite_as' for the doi of the data, and also any publications that reference it. I think the correct approach is to not have publications modify links, but have `doi` modify the single cite_as link.

pystac/extensions/scientific.py
```diff
@@ -0,0 +1,272 @@
...
+    def doi(self) -> Doi:
+        return self.item.properties.get(DOI)
+
+    @doi.setter
+    def doi(self, v: Union[str, Doi]) -> None:
+        if isinstance(v, Doi):
+            v = v.doi
+        self.item.properties[DOI] = v
+        url = DOI_URL_BASE + self.doi
+        self.item.add_link(pystac.link.Link(CITE_AS, url))
```
lossyrob, line 152: This should also clear previously added links.

pystac/extensions/scientific.py
```diff
@@ -0,0 +1,272 @@
...
+    def from_item(cls: _ScientificItemExtType, an_item: item.Item) -> _ScientificItemExtType:
+        return cls(an_item)
+
+    @classmethod
+    def _object_links(cls: _ScientificItemExtType) -> List:
+        return []
+
+    @property
+    def doi(self) -> Doi:
+        return self.item.properties.get(DOI)
```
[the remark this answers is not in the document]
lossyrob: If we keep the Doi wrapper, this should wrap in a Doi as the properties will contain the string version.

pystac/extensions/scientific.py
schwehr: Rereading your comments above and this one, I will start cleaning up what I have.  I will remove the Doi class.  It's just a convenience wrapper that can easily be a function added later.

schwehr: Rereading the spec, I see this near the bottom defining a Publication:

> doi | string | DOI of a publication referencing the data.

And then after that is:

> For all DOI names respective DOI links SHOULD be added to the links section (see chapter "Relation types").

To me this implies that the publication DOI should be a link.  If it had said `sci:doi` should be added to the links, then I would have agreed that the pubs should not be in the links.

Does that make sense?

lossyrob: Ah, good call - it seems like the publication DOIs should end up as links. And based on the language for the relation type, the `cite-as` rel type seems like the right way to go. This is confusing to me - to have several links with the rel type `cite-as` that point to either DOIs for the data DOIs for publications that point to the data - but that is how the spec reads now that you've clarified it for me. Thanks!

pystac/extensions/scientific.py
```diff
@@ -0,0 +1,272 @@
...
+
+PREFIX: str = 'sci:'
+DOI: str = PREFIX + 'doi'
+CITATION: str = PREFIX + 'citation'
+PUBLICATIONS: str = PREFIX + 'publications'
+
+# Link type.
+CITE_AS: str = 'cite-as'
+
+# TODO(schwehr): What is the correct regex for doi?
```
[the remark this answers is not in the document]
schwehr: I removed the doi regex completely.  https://github.com/stac-utils/pystac/pull/199/commits/27b4e064e936cc2bf4e534029c9a488c0b4c9e76

pystac/extensions/scientific.py
```diff
@@ -0,0 +1,272 @@
+"""Implement the scientific extension.
+
+DOI Handbook: https://doi.org/10.1000/182
+
+https://github.com/radiantearth/stac-spec/tree/dev/extensions/scientific
+"""
+
+# TODO(schwehr): Document.
```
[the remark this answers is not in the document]
schwehr: Added in https://github.com/stac-utils/pystac/pull/199/commits/27b4e064e936cc2bf4e534029c9a488c0b4c9e76

pystac/extensions/scientific.py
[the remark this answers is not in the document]
schwehr: I nuked the Doi wrapper for this PR.  It's not needed.  So all doi's are now just python strings.

pystac/extensions/scientific.py
```diff
@@ -0,0 +1,272 @@
...
+
+# TODO(schwehr): What is the correct regex for doi?
+# https://github.com/radiantearth/stac-spec/issues/910
+DOI_REGEX = r'10[.][0-9]{4}([.][0-9]+)*/.+'
+DOI_URL_BASE = 'https://doi.org/'
+
+_DoiType = TypeVar('Doi')
+_PublicationType = TypeVar('Publication')
+_ScientificItemExtType = TypeVar('ScientificItemExt')
+_ScientificCollectionExtType = TypeVar('ScientificCollectionExtType')
```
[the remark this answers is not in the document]
schwehr: They are me flailing about with type hints.  These don't actually seem to do any checking in that TypeVar(name_str) is a wildcard for any type.  It seems you can't use the class as a type hint inside a class.  Removed in https://github.com/stac-utils/pystac/pull/199/commits/e9afcabd4c7a250865e8133839bfd4f3f7a06fb3

pystac/extensions/scientific.py
schwehr: I don't know if we can safely remove all the old publication links.  Case that have me worried:

* What if the doi link and one of the publications are the same?
* What if the user adds a `cite-as` link on their own?
* What if a link gets mutated by some other call?

pystac/extensions/scientific.py
schwehr: `cite-as` does seem to have too many uses.  I guess we need to assume that the DOIs will be different and the code should search for cite-as links with the quoted doi string in the link.  Created https://github.com/radiantearth/stac-spec/issues/915 to discuss this.

pystac/extensions/scientific.py
schwehr: Maybe I don't worry and I just search and destroy based on cite-as links with the doi substring

schwehr: Fixed in https://github.com/stac-utils/pystac/pull/199/commits/af38dc1373818f5a7885ec6d36fd7aee19a80431

pystac/extensions/scientific.py
schwehr: Should the publication class go away and just use a dict?  I thought python3 had a [dotted dict](https://stackoverflow.com/questions/2352181/how-to-use-a-dot-to-access-members-of-dictionary) type thing, but ?  If we wanted to ditch mutability of the fields inside the pub, we could use [collections.namedtuple](https://docs.python.org/3/library/collections.html#collections.namedtuple)

pystac/extensions/scientific.py
schwehr: https://github.com/radiantearth/stac-spec/issues/915#issuecomment-722322144 said:

> The links were just intended as simplification for clients and users, mostly for the ones not supporting the scientific extension so that user could still follow links. It was just meant as a list of related dois to follow, not to transport any additional information or so.
> If two links would be the same, there should be just one link to it. The thing is that the DOI links can easily be generated if the client understands the sci extentsion/dois anyway so those clients don't really need the links.

pystac/extensions/scientific.py
lossyrob: After taking another look at this and pushing some code around, I think this is a tricker case than other cases, for instance `LabelClasses`, because this is a list of objects and not an dict. The 'proper' way to do what I'm talking about is to generate a collection class that would define it's own sequence methods, that would translate between the Item properties and Publication objects as items were retrieved, appended etc.

However, this adds a level of complexity such that the "worth it" factor is starting to tip in the direction of "no" for me. The way you have implemented works great, and the situation I mention may come up and confuse users but perhaps won't be too much of head-scratcher to work around.

That's to say, let's go forward with this solution and maybe re-evaluate if someone complains. Thanks for bearing with me on this!

lossyrob (approved):

PR #201, opened 2020-10-16
Fix spacenet tutorial, allow for single bbox argument to SpatialExtent, and other fixes.

### LINKED ISSUE #197: Tutorial code constructing SpatialExtents is incorrect
The [tutorial code](https://pystac.readthedocs.io/en/latest/tutorials/pystac-spacenet-tutorial.html) that shows how to construct a SpatialExtent from a list of shapely geometries only passes a `list[float]` to SpatialExtent rather than a `list[list[float]]`. Scroll down to code block `[11]`:

Screen Shot 2020-10-14 at 10 36 48 AM

It looks like the two blocks in https://pystac.readthedocs.io/en/latest/tutorials/how-to-create-stac-catalogs.html and https://pystac.readthedocs.io/en/latest/tutorials/creating-a-landsat-stac.html are correct and construct a `list[list[float]]`.

### LINKED ISSUE #198: Type check SpatialExtent bboxes parameter
It's easy to accidentally forget to pass `SpatialExtent(bboxes)` as a `list[list[float]]` and just pass a single dimensional bbox.  Here's two examples of this happening, #197 and azavea/noaa-flood-mapping#105. This leads to writing a STAC Catalog incompatible with the spec.

Two potential improvements?
1. Throw a TypeError or some equivalent exception when the incorrect type is passed
2. pystac could implicitly wrap a passed `list[float]` to `list[list[float]]`

lossyrob:
This PR adds a few fixes and enhancements based on solving the issues raised in #197 and #198.

- Fixes the spacenet tutorial to pass in a list of bboxes to the spatial extent, rather than a single bbox.
- Changes that tutorial to validate each of the catalogs it produces.
- Fix an issue raised by the validation of the label items, where the `label_classes` property is required. This requirement is dropped in a yet-unreleased version of the schemas, so a note was added to remove after 1.0.0-beta.2's successor is released.
- Fixed an issue with `Link.get_href()`, which was throwing a bad exception if a relative link had None for it's target href.
- Allowed users to pass in a single bbox to `SpatialExtent` and interval to `TemporalExtent` to guard against the common mistake mentioned in #198 
- Fixed an issue that caused geometry GeoJSON produced by shapely to fail PySTAC's validation. The reason for this was because the validation happened on the unmodified dict produced by a `to_dict` call. This can produce geometries that have their coordinates encoded with tuples instead of lists. The GeoJSON schema requires these to be lists, so the validation failed with a very cryptic message (a byproduct of using `oneOf` in JSON schemas - if the object doesn't match `oneOf` the schemas listed for a real validation failure, it instead complains about how it doesn't validate correctly against the first entry). The workaround here is to transform the dicts being passed in for validation to and from JSON serialization - if you JSON serialize a property whose value is a tuple, that will be transformed into a list for the JSON-ification. This solves the issue with shapely's `mapping` output, but also will solve for any case where a JSON serialization of an object would actually pass JSON schema validation, but the direct dict encoding may not.
- Fixes the failing readthedocs by upgrading nbsphinx. This upgrade produces a lot of warnings about rendering notebook output, which was fixed by including `ipython` in the dev dependencies.

Fixes #197 
Fixes #198

### DIFF
--- a/pystac/collection.py
+++ b/pystac/collection.py
@@ -1,3 +1,4 @@
+import collections.abc
 from datetime import datetime
 import dateutil.parser
 from dateutil import tz
@@ -305,6 +306,12 @@ class SpatialExtent:
             2D Collection with only one bbox would be [[xmin, ymin, xmax, ymax]]
     """
     def __init__(self, bboxes):
+        # A common mistake is to pass in a single bbox instead of a list of bboxes.
+        # Account for this by transforming the input in that case.
+        if isinstance(bboxes, collections.abc.Sequence):
+            if not isinstance(bboxes[0], collections.abc.Sequence):
+                bboxes = [bboxes]
+
         self.bboxes = bboxes
 
     def to_dict(self):
@@ -388,6 +395,13 @@ class TemporalExtent:
         Datetimes are required to be in UTC.
     """
     def __init__(self, intervals):
+        # A common mistake is to pass in a single interval instead of a
+        # list of intervals. Account for this by transforming the input
+        # in that case.
+        if isinstance(intervals, collections.abc.Sequence):
+            if not isinstance(intervals[0], collections.abc.Sequence):
+                intervals = [intervals]
+
         for i in intervals:
             if i[0] is None and i[1] is None:
                 raise STACError('TemporalExtent interval must have either '
--- a/pystac/collection.py
+++ b/pystac/collection.py
@@ -1,4 +1,4 @@
-import collections.abc
+from collections import abc
 from datetime import datetime
 import dateutil.parser
 from dateutil import tz
@@ -308,8 +308,8 @@ class SpatialExtent:
     def __init__(self, bboxes):
         # A common mistake is to pass in a single bbox instead of a list of bboxes.
         # Account for this by transforming the input in that case.
-        if isinstance(bboxes, collections.abc.Sequence):
-            if not isinstance(bboxes[0], collections.abc.Sequence):
+        if isinstance(bboxes, abc.Sequence):
+            if not isinstance(bboxes[0], abc.Sequence):
                 bboxes = [bboxes]
 
         self.bboxes = bboxes
@@ -398,8 +398,8 @@ class TemporalExtent:
         # A common mistake is to pass in a single interval instead of a
         # list of intervals. Account for this by transforming the input
         # in that case.
-        if isinstance(intervals, collections.abc.Sequence):
-            if not isinstance(intervals[0], collections.abc.Sequence):
+        if isinstance(intervals, abc.Sequence):
+            if not isinstance(intervals[0], abc.Sequence):
                 intervals = [intervals]
 
         for i in intervals:
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,3 +1,4 @@
+ipython==7.16.1
 jsonschema==3.2.0
 pylint==1.9.1
 Sphinx==1.8.0
@@ -6,5 +7,5 @@ sphinxcontrib-fulltoc==1.2.0
 sphinxcontrib-napoleon==0.7
 flake8==3.8.*
 yapf==0.28.*
-nbsphinx==0.4.3
+nbsphinx==0.7.1
 coverage==5.2.*
--- a/tests/test_link.py
+++ b/tests/test_link.py
@@ -1,0 +1,24 @@
+from datetime import datetime
+import unittest
+
+import pystac
+from tests.utils import (RANDOM_BBOX, RANDOM_GEOM)
+
+
+class LinkTest(unittest.TestCase):
+    def test_link_doest_fail_if_href_is_none(self):
+        """Tests to cover a bug that was uncovered where a non-None HREF
+        was supposed"""
+        catalog = pystac.Catalog(id='test', description='test')
+        item = pystac.Item(id='test-item',
+                           geometry=RANDOM_GEOM,
+                           bbox=RANDOM_BBOX,
+                           datetime=datetime.utcnow(),
+                           properties={'key': 'one'})
+        catalog.add_item(item)
+        catalog.set_self_href('/some/href')
+        catalog.make_all_links_relative()
+
+        for link in catalog.links:
+            if link.rel == 'item':
+                self.assertIsNone(link.get_href())
--- a/tests/test_link.py
+++ b/tests/test_link.py
@@ -4,21 +4,21 @@ import unittest
 import pystac
 from tests.utils import (RANDOM_BBOX, RANDOM_GEOM)
 
+TEST_DATETIME = datetime(2020, 3, 14, 16, 32)
+
 
 class LinkTest(unittest.TestCase):
-    def test_link_doest_fail_if_href_is_none(self):
-        """Tests to cover a bug that was uncovered where a non-None HREF
-        was supposed"""
-        catalog = pystac.Catalog(id='test', description='test')
+    def test_link_does_not_fail_if_href_is_none(self):
+        """Test to ensure get_href does not fail when the href is None"""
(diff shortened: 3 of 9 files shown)


tests/test_link.py
```diff
@@ -0,0 +1,24 @@
+from datetime import datetime
+import unittest
+
+import pystac
+from tests.utils import (RANDOM_BBOX, RANDOM_GEOM)
+
+
+class LinkTest(unittest.TestCase):
+    def test_link_doest_fail_if_href_is_none(self):
+        """Tests to cover a bug that was uncovered where a non-None HREF
```
schwehr, line 10: Rephrase.  `cover` and  `uncovered` in the same sentence.

tests/test_link.py
```diff
@@ -0,0 +1,24 @@
...
+        was supposed"""
+        catalog = pystac.Catalog(id='test', description='test')
```
schwehr, line 12: As above, use different strings for `id` and `description`

tests/test_link.py
```diff
@@ -0,0 +1,24 @@
...
+        item = pystac.Item(id='test-item',
+                           geometry=RANDOM_GEOM,
+                           bbox=RANDOM_BBOX,
+                           datetime=datetime.utcnow(),
+                           properties={'key': 'one'})
```
schwehr, line 17: Does the test work if the `properties` arg is {}?  If so, go with simpler args.

tests/test_link.py
```diff
@@ -0,0 +1,24 @@
...
+
+class LinkTest(unittest.TestCase):
+    def test_link_doest_fail_if_href_is_none(self):
+        """Tests to cover a bug that was uncovered where a non-None HREF
+        was supposed"""
+        catalog = pystac.Catalog(id='test', description='test')
+        item = pystac.Item(id='test-item',
+                           geometry=RANDOM_GEOM,
+                           bbox=RANDOM_BBOX,
+                           datetime=datetime.utcnow(),
```
schwehr, line 16: As above, use a fixed date.

tests/test_link.py
```diff
@@ -0,0 +1,24 @@
...
+                           properties={'key': 'one'})
+        catalog.add_item(item)
+        catalog.set_self_href('/some/href')
+        catalog.make_all_links_relative()
+
+        for link in catalog.links:
```
schwehr, line 22: Is there a way to ask for just the link under test?  If there was (for some bizarre reason... like me being a beginner hacking on things) no link with re == 'item', there would never be an assert checked.

tests/validation/test_validate.py
```diff
@@ -108,3 +109,25 @@ def test_validate_all(self):
...
+    def test_validates_geojson_with_tuple_coordinates(self):
+        """This unit tests guards against a bug where if a geometry
+        dict has tuples instead of lists for the coordinate sequence,
+        which can be produced by shapely, then the geometry still passses
+        validation.
+        """
+        geom = {
+            'type':
+            'Polygon',
+            'coordinates': (((-115.3057626, 36.1265426997), (-115.3057626, 36.1282976997),
```
schwehr, line 122: Go with fewer decimal places on the coords unless this is important for the test.  e.g. use 1 or 2 digit ints if possible.  If you need decimal places or negatives for the test, add a comment as to why.

tests/validation/test_validate.py
```diff
@@ -108,3 +109,25 @@ def test_validate_all(self):
...
+                             (-115.3075176, 36.1282976997), (-115.3075176, 36.1265426997),
+                             (-115.3057626, 36.1265426997)), )
```
schwehr, line 124: Style question: I don't know with pystac if you want or don't want the trailing `,`

tests/validation/test_validate.py
```diff
@@ -108,3 +109,25 @@ def test_validate_all(self):
...
+        }
+
+        item = pystac.Item(id='test-item',
+                           geometry=geom,
+                           bbox=[-115.308, 36.126, -115.305, 36.129],
+                           datetime=datetime.utcnow(),
+                           properties={'key': 'one'})
```
schwehr, line 131: As above, can you use `properties={}` for a simpler test?

schwehr (changes requested):
I'm not sure best to review for pystac.  I tried to err side of too many comments, questions, and suggestions.

pystac/collection.py
```diff
@@ -1,3 +1,4 @@
+import collections.abc
```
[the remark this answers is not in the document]
lossyrob: Done.

For 3.6 support - Yes, I'm a believer that we need to move on from Python 2...Python 3 was released 12 years ago and I think it's fair to expect that's enough time to upgrade :-) 

The top level description in the README says that it's Python 3 based. For versions <3.6, I'm not sure what the common practice is there, but I've mostly seen things supported 3.6 and on. Feel free to PR the README if you think more clarity on that is warranted, I'm all for clarification if it's needed!

requirements-dev.txt
```diff
@@ -1,3 +1,4 @@
+ipython==7.16.1
```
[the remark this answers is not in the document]
lossyrob: The Sphinx docs use nbsphinx to build doc versions of the tutorial notebooks. This requirement was added to avoid a long sequence of warning that were produced with the newer version of nbsphinx.

tests/test_link.py
```diff
@@ -0,0 +1,24 @@
+from datetime import datetime
+import unittest
+
+import pystac
+from tests.utils import (RANDOM_BBOX, RANDOM_GEOM)
```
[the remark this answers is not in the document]
lossyrob: good point. Can you open an issue to track that?

pystac/collection.py
```diff
@@ -305,6 +306,12 @@ class SpatialExtent:
             2D Collection with only one bbox would be [[xmin, ymin, xmax, ymax]]
     """
     def __init__(self, bboxes):
+        # A common mistake is to pass in a single bbox instead of a list of bboxes.
+        # Account for this by transforming the input in that case.
+        if isinstance(bboxes, collections.abc.Sequence):
+            if not isinstance(bboxes[0], collections.abc.Sequence):
```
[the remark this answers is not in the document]
lossyrob: > doc changes

I see your point, but I actually don't want to encourage anyone to pass a bare list, as this is against the spec. This change is really to catch a common error.

> IndexError...raises

Yes. I also want to be light on constructor validation, because if someone is trying to read invalid STAC, I think the stance PySTAC would want to take is to be able to read it in, validate it to discover errors, and correct things in the context of PySTAC. There are some places in the codebase that we validate in the constructor, and I'm thinking that actually we should move away from that entirely for consistency, and force the user to explicitly validate if they want to check if everything is valid.

pystac/collection.py
```diff
@@ -388,6 +395,13 @@ class TemporalExtent:
         Datetimes are required to be in UTC.
     """
     def __init__(self, intervals):
+        # A common mistake is to pass in a single interval instead of a
+        # list of intervals. Account for this by transforming the input
+        # in that case.
+        if isinstance(intervals, collections.abc.Sequence):
+            if not isinstance(intervals[0], collections.abc.Sequence):
```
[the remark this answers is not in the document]
lossyrob: Same comment as above.

tests/validation/test_validate.py
lossyrob: This is actually required, as it ensures tuple creation (or else it's interpreted as parans).

e.g.

```python
> x = (4), print(x)
4
> x = (4,); print(x)
(4,)

tests/test_link.py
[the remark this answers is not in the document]
schwehr: Created #205

schwehr (approved):

Issue #33, opened 2019-11-01
Implement Extension: SAR


lossyrob:
Implement the SAR extension, for Items containing SAR data.

https://github.com/radiantearth/stac-spec/tree/v0.8.0/extensions/sar

lossyrob:
This was implemented in #203

PR #203, opened 2020-10-21
Implement SAR extension.  #33

schwehr:
I'm new to type hints, so they aren't totally perfect.  The `classmethod`s are likely incorrect.  It would be great to get help to make them be correct or we can drop them. Type hints tested with pytype.

It seems like there should be a subset of `sar` available for collections.  e.g. this, which is definitely not correct.

- https://earthengine-stac.storage.googleapis.com/catalog/COPERNICUS_S1_GRD.json
- https://console.cloud.google.com/storage/browser/_details/earthengine-stac/catalog/COPERNICUS_S1_GRD.json

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -291,6 +291,19 @@ TimestampsItemExt
    :undoc-members:
    :show-inheritance:
 
+SAR Extension
+-------------
+
+Implements the `SAR Extension <https://github.com/radiantearth/stac-spec/tree/v1.0.0-beta.2/extensions/sar>`_.
+
+SarItemExt
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. autoclass:: pystac.extensions.sar.SarItemExt
+   :members:
+   :undoc-members:
+   :show-inheritance:
+
 Single File STAC Extension
 --------------------------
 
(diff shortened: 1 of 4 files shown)


lossyrob:
There is a TODO for documentation, but a review request. Is this still in draft state?

schwehr:
Yes, still a draft.  I was hoping you'd take a look at the non-doc part.

But hold off on that and I'll give it another look over myself and add the docs.

Sorry I'm having so much trouble with the github workflow.  The tooling feels very different than the reviews I do internally at Google.

lossyrob:
Gotcha - no worries. I just wasn't sure the state of things. If there's something particular you'd like me to take a look at, happy to, otherwise I think it'll shorten the review cycles if the review can happen all at once when the PR is out of draft phase.

lossyrob (approved):
Looks great, and good to merge. Few things:

- I added a commit to put the API docs in
- I created #230 to track making our usage of optional properties being set as None, and how that affects the `apply` method.
- I created #231 to track making our usage of Enums consistent. You're using enums here, but they are not used in throughout the rest of the library. I think that's good for now, but we want to eventually be consistent on our usage. If you have opinions about that, please comment in the issue!

PR #206, opened 2020-10-22
Add codespell to CI.

### LINKED ISSUE #200: Integrate codespell into CI process.
I, for one, am an admittedly horrible speller. My editor doesn't have spell check and I'm constantly letting bad spellings through.

One option to combat this for myself and others is to implement a `codespell` check on all *.rst, *.md and *.py files in the project. Codespell has an interactive mode and an ignore dictionary, so we could commit the ignore dict if there's any intentional spellings that block CI from passing.

lossyrob:
This PR adds running codespell to the scripts/test script, which is run by CI. This will fail PRs that have spelling mistakes. A currently empty ignore file is put into place in case codespell fails on any intentional spellings.

Fixes #200

### DIFF
--- a/scripts/test
+++ b/scripts/test
@@ -23,6 +23,16 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
         # Code formatting
         yapf -dpr pystac tests
 
+        # Code spelling
+        codespell -I .codespellignore -f \
+            pystac/*.py pystac/**/*.py \
+            tests/*.py tests/**/*.py \
+            docs/*.rst docs/**/*.rst \
+            docs/*.ipynb docs/**/*.ipynb \
+            scripts/* \
+            *.py \
+            *.md
+
         # Test suite with coverage enabled
         coverage run --source=pystac/ -m unittest discover tests/
         coverage xml
--- a/README.md
+++ b/README.md
@@ -92,7 +92,16 @@ To format code:
 
 Note that you may have to use `yapf3` explicitly depending on your environment.
 
-You can also run the `./scripts/test` script to check flake8 and yapf.
+To check for spelling mistakes in modified files:
+
+```
+> git diff --name-only | xargs codespell -I .codespellignore -f
+
+You can also run the `./scripts/test` script to check for linting, spelling, and run unit tests.
+
+### Continuous Integration
+
+CI will run the `scripts/test` script to check for code quality. If you have a Pull Request that fails CI, make sure to fix any linting, spelling or test issues reported by `scripts/test`.
 
 ### Documentation
 
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,3 +1,4 @@
+codespell==1.17.1
 ipython==7.16.1
 jsonschema==3.2.0
 pylint==1.9.1


scripts/test
```diff
@@ -23,6 +23,16 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
         # Code formatting
         yapf -dpr pystac tests
 
+        # Code spelling
+        codespell -I .codespellignore -f \
+            pystac/*.py pystac/**/*.py \
```
schwehr, line 28: Why not just do dirs?  It looks like you are leaving out the test data.  There are only a few things needed for the codespellignore for the test data.

```shell
codespell -I .codespellignore -f docs pystac scripts tests *.py *.md
```

schwehr (changes requested):
I can't figure out how to comment on an empty file... 

For `.codespellignore`, I think you'll need:

```
ACI
ALOS
NED
UE
```

scripts/test
lossyrob: Trying to just specify directories was including things like `docs/_build/html/_static/jquery.js` and `.ipynb_checkpoints` and `__pycache__` files for me, and throwing a lot of warnings about binary files.

Does that command cleanly for you?

schwehr: codespell on my machine doesn't seem to pay attention to `.codespellignore`, but if you add `--skip _build,__pycache__`, that seems to work for me minus the two examples files.

```
codespell --skip _build,__pycache__ -I .codespellignore -f docs pystac scripts tests *.py *.md
tests/data-files/examples/gee-0.6.2/catalog.json:18: ACI ==> ACPI
tests/data-files/examples/gee-0.6.2/catalog.json:438: ALOS ==> ALSO
tests/data-files/examples/gee-0.6.2/catalog.json:443: ALOS ==> ALSO
tests/data-files/examples/gee-0.6.2/catalog.json:448: ALOS ==> ALSO
tests/data-files/examples/gee-0.6.2/catalog.json:453: ALOS ==> ALSO
tests/data-files/examples/gee-0.6.2/catalog.json:2218: NED ==> NEED
tests/data-files/examples/sentinel-0.6.0/sentinel-2-l1c/9/V/catalog.json:1: UE ==> USE, DUE
```

lossyrob: Skipping the directories mentioned worked for avoiding those files, but it ended up picking up my venv files. Also I'll note that I don't think spell checking the data-files is necessary; misspellings from other folk's STAC is harmless and wont' affect tests. I think approach this as being explicit about what we want to include rather than doing everything but specific excludes will make for a better experience.

schwehr: Really either way is fine and gets the project better coverage than now.  Personally, I'd check the data files too.  Errors cause friction in unexpected ways for in different circumstances and tend to propagate via copy paste, but these are unlikely to cause too much trouble.

lossyrob: Makes sense. I think keeping it to code and documentation works for now, and we can expand it if we see we're running into situations where we wish we were spell checking stacs that are brought in. 

Can you move your review to Approved if this is resolved?

schwehr (approved):

PR #208, opened 2020-10-22
Allow for unnormalized path prefixes like /vsitar/

### LINKED ISSUE #180: PyStac Sometimes Removes Characters from URIs
Hello,

I have noticed that when getting absolute hrefs from PyStac (via `catalog.make_all_asset_hrefs_absolute()` or similar) PyStac sometimes loses key parts of non-local-fileystem paths.

In particular, I noticed this when using a catalog root of `/vsitar//tmp/catalog.tar/catalog.json`.  The absolute URIs that are returned are of the form `/vsitar/tmp/catalog.tar...` with the critical second `/` between `/vsitar` and `/tmp` missing (it is not a valid GDAL path with the slash removed).  This behavior seems to be generated by [this line](https://github.com/stac-utils/pystac/blob/cf804c370b79b4d2d1e6faf689eb915a9d7ed32c/pystac/utils.py#L110).

This behavior is possible to work around, but I wonder if it is possible to change this?

Thank you.

lossyrob:
The logic in make_absolute_href uses os.path.abspath, which normalizes paths. This lets us translate joined paths like /home/user/../catalog.json into proper paths. However, normalizing the path will also modify the root path being joined to in ways that may not be desirable. Specifically in the case of a /vsitar/ prefix, the double slashes that are required for proper GDAL reading are removed by normalization. This commit modifies the logic in make_absolute_href to detect when the original start path is modified by a call to abspath, and if so replaces the modified version with the original version.

Fixes #180

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -79,6 +79,13 @@ class UtilsTest(unittest.TestCase):
             actual = make_absolute_href(source_href, start_href)
             self.assertEqual(actual, expected)
 
+    def test_make_absolute_href_on_vsitar(self):
+        rel_path = 'some/item.json'
+        cat_path = '/vsitar//tmp/catalog.tar/catalog.json'
+        expected = '/vsitar//tmp/catalog.tar/some/item.json'
+
+        self.assertEqual(expected, make_absolute_href(rel_path, cat_path))
+
     def test_make_absolute_href_windows(self):
         utils._pathlib = ntpath
         try:
--- a/pystac/utils.py
+++ b/pystac/utils.py
@@ -108,6 +108,13 @@ def make_absolute_href(source_href, start_href=None, start_is_dir=False):
             else:
                 start_dir = _pathlib.dirname(parsed_start.path)
             abs_path = _pathlib.abspath(_join(is_path, start_dir, parsed_source.path))
+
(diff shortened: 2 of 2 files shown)


tests/test_utils.py
```diff
@@ -79,6 +79,13 @@ def test_make_absolute_href(self):
             actual = make_absolute_href(source_href, start_href)
             self.assertEqual(actual, expected)
 
+    def test_make_absolute_href_on_vsitar(self):
+        cat_path = '/vsitar//tmp/catalog.tar/catalog.json'
+        rel_path = 'some/item.json'
+        result = make_absolute_href(rel_path, cat_path)
```
schwehr, line 85: Minor suggestion for clarity:

```
rel_path = 'some/item.json'
cat_path = '/vsitar//tmp/catalog.tar/catalog.json'
expected = '/vsitar//tmp/catalog.tar/some/item.json'

self.assertEqual(expected, make_absolute_href(rel_path, cat_path))
```

lossyrob: Ok, thanks. That is minor, but I made the change.

Can you mark PRs that have minor suggestions as approved, and use 'request changes' if there's specific changes or discussion left? Thanks!

I believe this falls into the former category, so I'll merge once CI passes.

PR #209, opened 2020-10-23
version ext: Add docs, add type hints, and use pystac.Item & pystac.Collection

schwehr:


lossyrob (approved):
Thanks - should have caught the lack of docstrings in the original PR :-)

PR #221, opened 2020-10-28
Fix #186 and related issues

### LINKED ISSUE #186: Improve error handling when accidentally importing a Collection with Catalog
Today I accidentally used `pystac.Catalog.from_file` to read a `collection.json` instead of `pystac.Collection.from_file` and got this error:
```
TypeError: 'Catalog' object is not subscriptable
```

I could see a number of people making this mistake or it being a difficult to handle bug in some automated process that reads catalogs. It would be nice if this error was handled such that a specific pystac exception is thrown when this happens or it automatically switches to the collection reader when a collection is detected.

lossyrob:
This PR fixes issues uncovered while investigating #186.

- The merge_common_properties logic was executing in the case of a 0.9.0 item that did not have the 'commons' extension implemented. This was due to a bug in the logic that didn't account for the case where `stac_extensions` was not declared in the Item.
- When the merge_common_properties logic does execute, if the collection that is referenced by the item produces a cached value that is not a Collection (for example if a collection was read in as a Catalog as described in #186), throw a more clear error than was previously being thrown.

The two other changes below relate to the caching changes made in #214:
- The logic in Catalog and Collection `from_dict` was causing root objects to be cached twice - once with it's ID key and once when the HREF was set on the object. This changes the logic to ensure that if an object is read in with an HREF, it's cached only by its HREF. 
- `set_self_href` on Collection was promoted to be the implementation for all STACObjects. This method on Collection should have been deleted in #214, but is done so here.

Fixes #186

### DIFF
--- a/pystac/catalog.py
+++ b/pystac/catalog.py
@@ -88,12 +88,13 @@ class Catalog(STACObject):
             self.extra_fields = extra_fields
 
         self._resolved_objects = ResolvedObjectCache()
-        self._resolved_objects.cache(self)
 
         self.add_link(Link.root(self))
 
         if href is not None:
             self.set_self_href(href)
+
+        self._resolved_objects.cache(self)
 
     def __repr__(self):
         return '<Catalog id={}>'.format(self.id)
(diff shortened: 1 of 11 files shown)


lossyrob:
Code coverage decreases because code removal; all additions tested.

PR #232, opened 2020-10-30
Handle setting new versions correctly in version extension.

### LINKED ISSUE #210: version extension link setters do not check for existing links
Looking at the review comments in #199 for the scientific extension, I got derailed by version.  The mutability of a version extension item is broken.  If you set `item0.latest = item1` or set it from `apply`, and then later set `item0.lastest = item2`, item0 will end up with two SUCCESSOR links.  I don't think that is what anyone expects.

I plan to address this after #209.

Sketch of a test:

```python
    def test_set_links_twice(self):
        latest1 = make_item(2013)
        predecessor1 = make_item(2010)
        successor 1= make_item(2012)
        self.item.ext.version.apply(self.version, deprecated, latest1, predecessor1, successor1)

        year = 2015
        latest2 = make_item(year)
        expected_href = URL_TEMPLATE % year
        self.item.ext.version.latest = latest2
        links = self.item.get_links(version.LATEST)
        self.assertEqual(1, len(links))
        self.assertEqual(expected_href, link.get_href())

        year = 2009
        predecessor2 = make_item(year)
        expected_href = URL_TEMPLATE % year
        self.item.ext.version.latest = latest2
        links = self.item.get_links(version.PREDECESSOR)
        self.assertEqual(1, len(links))
        self.assertEqual(expected_href, link.get_href())

        year = 2014
        successor2 = make_item(year)
        expected_href = URL_TEMPLATE % year
        self.item.ext.version.latest = latest2
        links = self.item.get_links(version.SUCCESSOR)
        self.assertEqual(1, len(links))
        self.assertEqual(expected_href, link.get_href())
```

lossyrob:
Prior to this change, setting the 'latest', 'successor', or 'predecessor' would add links, but not clear any old links if there were already versions set. This ensures that the links are cleared before setting any new version of the object.

It also ensures that setting a version to 'None' will clear the links, allowing users to remove a version from an item or collection.

Fixes #210

### DIFF
--- a/tests/extensions/test_version.py
+++ b/tests/extensions/test_version.py
@@ -140,6 +140,59 @@ class VersionItemExtTest(unittest.TestCase):
         self.assertIs(successor, item1_copy)
         self.assertIs(latest, item1_copy)
 
+    def test_setting_none_clears_link(self):
+        deprecated = False
+        latest = make_item(2013)
+        predecessor = make_item(2010)
+        successor = make_item(2012)
+        self.item.ext.version.apply(self.version, deprecated, latest, predecessor, successor)
+
+        self.item.ext.version.latest = None
+        links = self.item.get_links(version.LATEST)
+        self.assertEqual(0, len(links))
+        self.assertIsNone(self.item.ext.version.latest)
+
+        self.item.ext.version.predecessor = None
+        links = self.item.get_links(version.PREDECESSOR)
+        self.assertEqual(0, len(links))
+        self.assertIsNone(self.item.ext.version.predecessor)
+
+        self.item.ext.version.successor = None
+        links = self.item.get_links(version.SUCCESSOR)
+        self.assertEqual(0, len(links))
+        self.assertIsNone(self.item.ext.version.successor)
+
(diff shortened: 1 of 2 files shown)


tests/extensions/test_version.py
```diff
@@ -140,6 +140,59 @@ def test_full_copy(self):
         self.assertIs(successor, item1_copy)
         self.assertIs(latest, item1_copy)
 
+    def test_setting_none_clears_link(self):
+        deprecated = False
+        latest1 = make_item(2013)
```
schwehr, line 145: Why `latest1`?  Why not use just `latest` here and below for latest, predecessor, and successor in ` test_setting_none_clears_link` for Item and Collection tests?  It makes sense in `test_multiple_link_setting` but had me looking for something I had missed in this test.

schwehr (changes requested):

tests/extensions/test_version.py
lossyrob: Copy pasta mistake. Good catch!

schwehr (approved):

PR #238, opened 2020-11-05
Compatiblity with CFTime

rabernat:
I am trying to catalog climate model data stored in Zarr format on Google Cloud / AWS using STAC. My approach is to create a STAC collection for each Zarr dataset and link to the Zarr data using collection-level assets (see related discussion @ https://github.com/radiantearth/stac-spec/issues/781).

I have hit a tiny snag with datetimes. I need to populate `temporal_extent` for the collections. I get this by reading the dataset's time range. The times are encoded using the [CF Conventions for time coordinates](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.8/cf-conventions.html#time-coordinate). In python, CF times are implemented by the [cftime](https://unidata.github.io/cftime/index.html) package. cftime provides a (mostly) duck-type compatible with datetime. However, cftime has no `.tzinfo` attribute.

This tiny change allows pystac to work with cftime datetimes.

I have not added a test because I didn't want to add cftime to the dependencies. One could possibly use a mock to test this feature. Or we could merge this tiny change with no test. I'll follow whatever path the devs recommend.

cc @charlesbluca

### DIFF
--- a/pystac/utils.py
+++ b/pystac/utils.py
@@ -150,7 +150,7 @@ def datetime_to_str(dt):
     Returns:
         str: The ISO8601 formatted string representing the datetime.
     """
-    
+
     # workaround for CFTime datetimes
     if hasattr(dt, 'tzinfo') and dt.tzinfo is None:
         dt = dt.replace(tzinfo=timezone.utc)
--- a/pystac/utils.py
+++ b/pystac/utils.py
@@ -150,8 +150,10 @@ def datetime_to_str(dt):
     Returns:
         str: The ISO8601 formatted string representing the datetime.
     """
-    if dt.tzinfo is None:
-        dt = dt.replace(tzinfo=timezone.utc)
+    
+    if hasattr(dt, 'tzinfo'):
+        if dt.tzinfo is None:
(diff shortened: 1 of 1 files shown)


rabernat:
Could the maintainers (maybe @lossyrob) let me know whether this PR might be acceptable. I'd be happy to make any changes required.

lossyrob (changes requested):
@rabernat apologies for the delay in review.

The change looks great, however can you provide a unit test that demostrates the failure case when this change is not in place/succeeds with this change? Thanks!

rabernat:
> The change looks great, however can you provide a unit test that demostrates the failure case when this change is not in place/succeeds with this change?

Sure. Would you prefer I add cftime as a test dependency or that I use some kind of mock / monkeypatch to datetime?

lossyrob:
I think the mock/monkeypatch version would work, with a comment pointing at the specific cftime class it is mimicking in case API changes down the road. Thanks!

rabernat:
After thinking this through, I decided it made more sense to fix this in cftime, rather than pystac. See https://github.com/Unidata/cftime/pull/209. But thanks for considering this.

PR #241, opened 2020-11-19
Fix unexpected behaviour of `generate_subcatalogs`

### LINKED ISSUE #240: Unexpected behaviour of `generate_subcatalogs`
Hello, first of all thanks so much for this fantastic library - it's really great! I observed an unexpected behaviour of the Catalog `generate_subcatalogs` method - I thought it might be worth reporting. 

I  am setting up a local catalog with some items organised in a sub-catalog structure. I would like to add items at different stages (e.g. whenever new assets become available). I would have expected that calling the `generate_subcatalogs` method every time I add new items to the root catalog would do the job. However, every time the method is called, the full sub-catalog structure is replicated. Here's a minimal working example:  

```python
import pystac 
from datetime import datetime

# initialise my catalog
catalog = pystac.Catalog(id='my-catalog', description='My Awesome Catalog')

properties = dict(first=1, second=2, third=3)

# add first set of items 
for id in ['A', 'B', 'C']:
    item = pystac.Item(id, None, None, datetime.utcnow(), properties)
    catalog.add_item(item)

# generate sub-catalog structure
catalog.generate_subcatalogs(template='${first}/${second}/${third}')
catalog.describe()
# * <Catalog id=my-catalog>
#     * <Catalog id=1>
#         * <Catalog id=2>
#             * <Catalog id=3>
#               * <Item id=A>
#               * <Item id=B>
#               * <Item id=C>

# add more items
for id in ['D', 'E', 'F']:
    item = pystac.Item(id, None, None, datetime.utcnow(), properties)
    catalog.add_item(item)
catalog.generate_subcatalogs(template='${first}/${second}/${third}')
catalog.describe()
# * <Catalog id=my-catalog>
#     * <Catalog id=1>
#         * <Catalog id=2>
#             * <Catalog id=3>
#                 * <Catalog id=1>
#                     * <Catalog id=2>
#                         * <Catalog id=3>
#                           * <Item id=A>
#                           * <Item id=B>
#                           * <Item id=C>
#     * <Catalog id=1>
#         * <Catalog id=2>
#             * <Catalog id=3>
#               * <Item id=D>
#               * <Item id=E>
#               * <Item id=F>
``` 

while I would like to obtain the following:

```python
# * <Catalog id=my-catalog>
#     * <Catalog id=1>
#         * <Catalog id=2>
#             * <Catalog id=3>
#               * <Item id=A>
#               * <Item id=B>
#               * <Item id=C>
#               * <Item id=D>
#               * <Item id=E>
#               * <Item id=F>
```

If you agree this is unexpected, one could initialise the following sub-catalog dictionary:
https://github.com/stac-utils/pystac/blob/2a0c850cb2f9dfde628c0e5c1973e77230ef46a6/pystac/catalog.py#L545
with all the sub-catalogs up to the current level:
```python
subcat_id_to_cat = {}
curr_parent = self
while curr_parent is not None:
    subcat_id_to_cat[curr_parent.id] = curr_parent
    curr_parent = curr_parent.get_parent()
```
In this way, sub-catalog levels that are already present are not replicated..
Thanks in advance!

fnattino:
Fixes #240

### DIFF
--- a/pystac/catalog.py
+++ b/pystac/catalog.py
@@ -520,7 +520,7 @@ class Catalog(STACObject):
 
         return self
 
-    def generate_subcatalogs(self, template, defaults=None, **kwargs):
+    def generate_subcatalogs(self, template, defaults=None, parent_ids=None, **kwargs):
         """Walks through the catalog and generates subcatalogs
         for items based on the template string. See :class:`~pystac.layout.LayoutTemplate`
         for details on the construction of template strings. This template string
@@ -533,35 +533,37 @@ class Catalog(STACObject):
             defaults (dict):  Default values for the template variables
                 that will be used if the property cannot be found on
                 the item.
+            parent_ids (List[str]): Optional list of the parent catalogs'
+                identifiers. If the bottom-most subcatalags already match the
+                template, no subcatalog is added.
 
         Returns:
             [catalog]: List of new catalogs created
         """
         result = []
+        parent_ids = parent_ids or list()
+        parent_ids.append(self.id)
         for child in self.get_children():
-            result.extend(child.generate_subcatalogs(template, defaults=defaults))
+            result.extend(child.generate_subcatalogs(template, defaults=defaults,
+                                                     parent_ids=parent_ids.copy()))
 
         layout_template = LayoutTemplate(template, defaults=defaults)
-
-        subcat_id_to_cat = {}
-        curr_parent = self
-        while curr_parent is not None:
-            subcat_id_to_cat[curr_parent.id] = curr_parent
-            curr_parent = curr_parent.get_parent()
 
         items = list(self.get_items())
         for item in items:
             item_parts = layout_template.get_template_values(item)
+            id_iter = reversed(parent_ids)
+            if all([id == next(id_iter, None) for id in reversed(item_parts.values())]):
+                continue
             curr_parent = self
             for k, v in item_parts.items():
                 subcat_id = '{}'.format(v)
-                subcat = subcat_id_to_cat.get(subcat_id)
+                subcat = curr_parent.get_child(subcat_id)
                 if subcat is None:
                     subcat_desc = 'Catalog of items from {} with {} of {}'.format(
                         curr_parent.id, k, v)
                     subcat = pystac.Catalog(id=subcat_id, description=subcat_desc)
                     curr_parent.add_child(subcat)
-                    subcat_id_to_cat[subcat_id] = subcat
                     result.append(subcat)
                 curr_parent = subcat
             self.remove_item(item.id)
--- a/pystac/catalog.py
+++ b/pystac/catalog.py
@@ -555,7 +555,8 @@ class Catalog(STACObject):
         for item in items:
             item_parts = layout_template.get_template_values(item)
             id_iter = reversed(parent_ids)
-            if all([id == next(id_iter, None) for id in reversed(item_parts.values())]):
+            if all(['{}'.format(id) == next(id_iter, None)
+                    for id in reversed(item_parts.values())]):
                 continue
             curr_parent = self
             for k, v in item_parts.items():
--- a/pystac/catalog.py
+++ b/pystac/catalog.py
@@ -557,6 +557,9 @@ class Catalog(STACObject):
             id_iter = reversed(parent_ids)
             if all(['{}'.format(id) == next(id_iter, None)
                     for id in reversed(item_parts.values())]):
+                # Skip items for which the sub-catalog structure already
+                # matches the template. The list of parent IDs can include more
+                # elements on the root side, so compare the reversed sequences.
                 continue
             curr_parent = self
             for k, v in item_parts.items():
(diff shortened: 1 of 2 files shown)


lossyrob (changes requested):
This is looking good - thank you! A couple of thoughts:

The way this works now, the `generate_subcatalogs` recurses down to the child catalogs and then crawls up it's parent tree to cache parent catalog IDs. This ends up iterating over the catalog multiple times - both down the tree with `for child in self.get_children():` and during the while loop you introduce. Can you structure it in such a way that it reduced the iterations? I think passing forward the `subcat_id_to_cat` as an optional parameter might do the trick.

Also, could you provide one or more unit tests that would cover the case mentioned in #240? Might be worth trying to test any edge cases that may arise around catalog merging - e.g. what happens if two subcatalogs have the same ID? This is often the case for subcatalogs that are based on date - so a subcatalog of day "15" might exist for a subcatalog of month "1", but also for month "2". I'm not sure how this code will behave in this scenario.

fnattino:
Thanks a lot for having a look! I have followed your suggestion and removed the redundant iteration over the catalog by passing over the parent catalog IDs. I have changed the subsequent loop over items slightly: I use the list of parent IDs to check whether subcatalogs need to be added, but I crawl down the template levels using the subcatalogs' `get_child` method. I think this should solve most of the problems with the same ID appearing in different sub-catalog levels, as this should naturally follow the catalog's branches. 

I will add unit tests with something like #240 and other cases!

fnattino:
I have added unit tests for the following edge cases:
- if the sub-catalog structure already match the template, `generate_subcatalogs` should do nothing (or, in other words, it should be possible to call `generate_subcatalogs` multiple times without side effects);
- A consistent sub-catalog structure should be obtained if items are added to a catalog at multiple stages and `generate_subcatalogs` is called every times items are added (case of #240);
- `generate_subcatalogs` should behave for catalog structures where the same sub-catalog ID appears in different catalog branches (as in the month/day example above); 
- `generate_subcatalogs` should also work for structures where the same sub-catalog ID appears at different levels in the same branch, as for the template `${property1}/${property2}` where both `property1` and `property2` can assume the same value. I am thinking e.g. to the `eo:row` and `eo:column` properties in the  AWS Landsat catalog..

pystac/catalog.py
```diff
@@ -533,29 +533,40 @@ def generate_subcatalogs(self, template, defaults=None, **kwargs):
...
         layout_template = LayoutTemplate(template, defaults=defaults)
-        subcat_id_to_cat = {}
+
         items = list(self.get_items())
         for item in items:
             item_parts = layout_template.get_template_values(item)
+            id_iter = reversed(parent_ids)
+            if all(['{}'.format(id) == next(id_iter, None)
+                    for id in reversed(item_parts.values())]):
+                continue
```
lossyrob, line 560: The `'{}'.format(id)` here is redundant, and could just be `id`, yeah?

Is the reversing to try and reduce iterations as the likely case that the more root-leaning parent IDs would match? If so the iteration of the `reverse` between the id_iter and item_parts would cancel this out I think - and I believe this is equivalent to the forward-iterating case, so perhaps the reverse's can be dropped?

lossyrob (changes requested):
Nice work, well tested. Just a couple of questions on a bit of code and I think this is good to go!

pystac/catalog.py
fnattino: > The '{}'.format(id) here is redundant, and could just be id, yeah?

Actually the values in the dictionary returned by `layout_template.get_template_values(item)` are not necessarily strings (they have the actual property type), so they need to be converted in order to compare to the IDs in the catalog structure. I have used the same syntax as further below (line 551 in the original version of the code) for consistency.

> Is the reversing to try and reduce iterations as the likely case that the more root-leaning parent IDs would match? If so the iteration of the reverse between the id_iter and item_parts would cancel this out I think - and I believe this is equivalent to the forward-iterating case, so perhaps the reverse's can be dropped?

Here I want to to check whether the template sub-catalog structure matches the actual one, but on the item-leaning side. Reversing allows me to align the current and the desired structures using the item position as a reference. Suppose the items are in the catalog `/my-catalog/my-collection/2020/12/01` and the template is `${year}/${month}/${day}` (i.e. `parent_ids = ['my-catalog', 'my-collection', '2020', '12', '01']` and `item_parts.values() = ['2020', '12', '01']`): by reversing both the structure and the template I can verify that they match on the outermost side, thus allowing me to skip to the next element.

lossyrob: > Actually the values in the dictionary returned by layout_template.get_template_values(item) are not necessarily strings

Ah! Of course. My mistake.

> Here I want to to check whether the template sub-catalog...

Ok, that makes sense now. The parent IDs can be longer than the item_parts, and you're only exhausting the item parts - that was the part I was missing. Would you mind adding some comments to the effect of the comment above, as it'll make things a bit easier to parse for future readers?

fnattino: Good point, sorry for the cryptic code :)

lossyrob (approved):
Thank you!

PR #243, opened 2020-11-30
Get eo bands defined in assets only

emmanuelmathot:
**Context**
As defined in the [EO extension](https://github.com/radiantearth/stac-spec/tree/master/extensions/eo) specification, the `eo:bands` properties are allowed in assets without being declared at the item level

**Issue**

The `get_bands` method in `EOItemExt` does not return bands defined in asset level if no `asset` parameter in specified.

**Proposed Solution**

The `get_bands` method in `EOItemExt` return aggregated list of bands defined in asset level if no `asset` parameter in specified.

**PR Checklist:**

- [X] This PR is made against the dev branch (all proposed changes except releases should be against dev, not master).
- [X] This PR has **no** breaking changes.

### DIFF
--- a/pystac/extensions/eo.py
+++ b/pystac/extensions/eo.py
@@ -55,7 +55,8 @@ class EOItemExt(ItemExtension):
         """Gets an Item or an Asset bands.
 
         If an Asset is supplied and the bands property exists on the Asset,
-        returns the Asset's value. Otherwise returns the Item's value
+        returns the Asset's value. Otherwise returns the Item's value or 
+        all the asset's eo bands
 
         Returns:
             List[Band]
@@ -64,6 +65,13 @@ class EOItemExt(ItemExtension):
             bands = asset.properties.get('eo:bands')
         else:
             bands = self.item.properties.get('eo:bands')
+
+        # get assets with eo:bands even if not in item
+        if bands is None:
+            bands = []
+            for (value) in self.item.get_assets().items():
+                if 'eo:bands' in value.properties:
+                    bands.extend(value.properties.get('eo:bands'))
 
         if bands is not None:
             bands = [Band(b) for b in bands]
--- a/pystac/extensions/eo.py
+++ b/pystac/extensions/eo.py
@@ -69,7 +69,7 @@ class EOItemExt(ItemExtension):
         # get assets with eo:bands even if not in item
         if bands is None:
             bands = []
-            for (value) in self.item.get_assets().items():
+            for (key, value) in self.item.get_assets().items():
                 if 'eo:bands' in value.properties:
                     bands.extend(value.properties.get('eo:bands'))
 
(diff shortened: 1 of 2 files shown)


emmanuelmathot:
I join an STAC item example for validation
[K3_20201112193439_45302_18521139_L1G.json.txt](https://github.com/stac-utils/pystac/files/5615092/K3_20201112193439_45302_18521139_L1G.json.txt)

pystac/extensions/eo.py
```diff
@@ -65,6 +66,13 @@ def get_bands(self, asset=None):
         else:
             bands = self.item.properties.get('eo:bands')
 
+        # get assets with eo:bands even if not in item
+        if asset is None and bands is None:
+            bands = []
+            for (key, value) in self.item.get_assets().items():
+                if 'eo:bands' in value.properties:
+                    bands.extend(value.properties.get('eo:bands'))
```
lossyrob, line 74: What happens if multiple assets have the same bands? This could be the case if you have for example RGB and RGBIR imagery in the same Item.

lossyrob (changes requested):
This brings up some general questions for me - is the general rule that, for a property that _could_ be on an item (that is an array) that instead has values across assets, should the item always check asset properties? Or is this particular to eo:bands?

Despite the general rule not being clear, the practical instance of this - where if you ask an Item what it's eo:bands are, and those are defined in the assets, you'd want the right answer - seems like a good addition, especially if you've hit on a use case needing it.

I put a comment around a case where assets could have the same band information - we shouldn't repeat band data in that case. Besides handling and testing that case, this looks good to me!

pystac/extensions/eo.py
emmanuelmathot: I see your point. IMO, the `eo:bands` property should be limited to assets because there are strongly linked. Furthermore, the results of get_bands should be a dictionary key/value with key as the asset key and value the list of `eo:bands`. My proposal here is to resolve a case that is allowed by the specification.
I would keep it that way even if leading to 'duplicates' in the function returned list. It simply returns what is specified in the item. If I extend your example, the red band of the RGB could be different of the RGBIR because their composition could be made differently (e.g. different center_wavelength based on optical calibration) and thus that would be 2 different bands.

lossyrob: It's a good point about the bands being potentially different but with the same name - band.name isn't unique according to the spec, so really you'd have to check all properties in order to de-duplicate. I do think this is probably a less common case though.

@matthewhanson you've worked with the eo extension a bunch, do you have an opinion here? This is for a pystac.Item returning bands from the extension call `item.ext.eo.get_bands()`, returning either the bands specified in the Item or a aggregation of bands specified in the assets - the potential for duplication happening in the latter case.

lossyrob: In lieu of further input and to move this along, I'm in favor of your suggestion to keep it as is, and we can modify if this seems like the incorrect choice down the road.

lossyrob (approved):

PR #247, opened 2020-12-22
Collection TemporalExtent can be open ended

### LINKED ISSUE #246: Can TemporalExtent be open on both sides?
If a `Collection` tries to declare

```
extent: { temporal: {interval: [[null, null]]}, ...}
```

pystac v0.5.3 raises 

```
pystac.STACError: TemporalExtent interval must have either a start or an end time, or both
```

[Per the spec, temporal extents may be open ended](https://github.com/radiantearth/stac-spec/blob/v1.0.0-beta.2/collection-spec/collection-spec.md#temporal-extent-object), with no restriction on both sides being open ended.

vpipkt:
Fixes #246

### DIFF
--- a/pystac/collection.py
+++ b/pystac/collection.py
@@ -3,7 +3,7 @@ from datetime import datetime
 import dateutil.parser
 from dateutil import tz
 from copy import (copy, deepcopy)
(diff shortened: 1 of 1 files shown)


vpipkt:
@lossyrob 🙇

Issue #258, opened 2021-02-09
STAC item validation failure when using view extension apply


tyler-c2s:
Currently getting validation errors when using the view extensions `.apply()` method. 

Example:
```python
view_dict = {'off_nadir': 0, 'sun_azimuth': 133.076479108908, 'sun_elevation': 33.4042972979731}
item.ext.view.apply(**view_dict)
item.validate()
```
results in:
```
ValidationError: None is not of type 'number'

Failed validating 'type' in schema['allOf'][1]['properties']['properties']['properties']['view:incidence_angle']:
    {'maximum': 90,
     'minimum': 0,
     'title': 'Center incidence angle',
     'type': 'number'}

On instance['properties']['view:incidence_angle']:
    None
```

this is not the case when directly adding each property:
```python
"""
view_dict = {'off_nadir': 0, 'sun_azimuth': 133.076479108908, 'sun_elevation': 33.4042972979731}
item.ext.view.off_nadir = view_dict["off_nadir"]
item.ext.view.sun_azimuth = view_dict["sun_azimuth"]
item.ext.view.sun_elevation = view_dict["sun_elevation"]
item.validate()
```

the issue appears to be that when using the apply function, any non-included parameters are set to `None`, causing issues in validation

PR #259, opened 2021-02-09
handle optional arguments in View Extension

tyler-c2s:
Resolves https://github.com/stac-utils/pystac/issues/258, in which parameters that were not passed to the `.apply()` method in the view extension would result in validation issues. 

Updated:
- Only sets property is passed is not None
- Raise error if all parameters are None
- Added tests for added functionality

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [unreleased]
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


lossyrob:
Thanks! Would you mind updating the [CHANGELOG.md]() with this change? This will be the first unreleased addition, so if you could add a `## [unreleased]` section that'd be great.

Issue #231, opened 2020-10-30
Refactor to use Enums consistently in PySTAC


lossyrob:
Currently, things like `CatalogType`, `LinkType` etc are not enums but constants held in a class. In other places like the SAR extension, we use Enums.

We should be consistent about our usage of enums across the library.

Note that moving from constants to enums would be a pretty big API break, potentially requiring a lot of client code changes. If we were to make that decision, it might be best to provide upgrade scripts to ease some of that API breakage pain.

If you have an opinion on whether or not Enums should be used across the library, please use this issue to voice it!

kylebarron:
As mentioned in the [linked issue](https://github.com/stac-utils/stactools/pull/61#discussion_r577031459), I'm a fan of using `Enum`. One nice thing is that it lets you iterate over the enum easily. For example, say you were making a CLI where you wanted to provide a media type option. Currently you'd have to make the option list manually and keep it up to date. When subclassing from enum, it's much simpler: `click.Choice([e.name for e in MediaType], case_sensitive=False))`.

> potentially requiring a lot of client code changes

Are you referring to code changes inside pystac or in third party libraries? I wouldn't expect many code changes, and if you're using existing classes as de-facto enums, you might not need any...

If you have a class like
```py
class MediaType:
    COG = 'image/tiff; application=geotiff; profile=cloud-optimized'
    GEOJSON = 'application/geo+json'
    GEOPACKAGE = 'application/geopackage+sqlite3'
```
and you detect media type in code with 
```py
if media_type == MediaType.COG
```
then you need no changes there between the above and 
```py
class MediaType(str, Enum):
```

I'm guessing by backwards incompatible changes, you're referring to the string representation of the enum object? E.g. the current implementation prints the value of the enum

```py
print(MediaType.COG)
# image/tiff; application=geotiff; profile=cloud-optimized
```

whereas changing to subclassing from `Enum` by default prints differently:
```py
print(MediaType.COG)
# MediaType.COG
```

however this is easy to get around by overriding the `__str__` method:
```py
class MediaType(str, Enum):
    def __str__(self):
        return str(self.value)

    COG = 'image/tiff; application=geotiff; profile=cloud-optimized'
    GEOJSON = 'application/geo+json'
    GEOPACKAGE = 'application/geopackage+sqlite3'

print(MediaType.COG)
# image/tiff; application=geotiff; profile=cloud-optimized
```

Any other things that you think wouldn't be backwards-compatible?

[...]

PR #261, opened 2021-02-17
Subclass relevant classes from Enum

kylebarron:
For https://github.com/stac-utils/pystac/issues/231. I'm guessing you only need the `__str__` override for the `MediaType` class, since that's the only one that's likely to actually get printed?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Fix handling of optional properties when using apply on view extension ([#259](https://github.com/stac-utils/pystac/pull/259))
 
+### Changed
+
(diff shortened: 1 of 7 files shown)


lossyrob:
Nice, thanks! Two other Enums to hit:

- [Extensions](https://github.com/stac-utils/pystac/blob/cbf22cf88cee58320e310bd293ceaa4d16f1bd68/pystac/extensions/__init__.py#L10)
- [LabelType](https://github.com/stac-utils/pystac/blob/cbf22cf88cee58320e310bd293ceaa4d16f1bd68/pystac/extensions/label.py#L10)

lossyrob:
@kylebarron good to go on this - but could you add a CHANGELOG.md entry? Still have to add that step to a PR template...

kylebarron:
Give me a minute to check on one thing before merging

kylebarron:
I think this should be ready to merge, thanks

Issue #178, opened 2020-08-25
Generate collection summaries field automatically from items


lossyrob:
Generate a collection's `summaries` property based on the items it contains. See [the summaries detail section of the collection spec](https://github.com/radiantearth/stac-spec/blob/master/collection-spec/collection-spec.md#collection-fields).

Feature request by @m-mohr

lossyrob:
@m-mohr Would this take as input the field names of the item's `properties` that you want to generate summaries for? What other inputs would be required?

m-mohr:
Having that would be my favorite feature in PySTAC! 🤩

I think by default it would just summarize all properties found, maybe except some fields that usually don't make sense (complex objects/too many array elements) or are available top-level like datetime (already available in in the extent, but the extents could also get populated automatically), license, title, description, ...

A good idea in addition is to let the user choose, indeed. We can also discuss this further in the future.

cholmes:
I was thinking about this more as I wrote up https://github.com/radiantearth/stac-spec/pull/985

I do think there should be a default to summarize all properties 'that make sense'. And then some mechanism to let a user choose. 

For the first I think that's:

* Any property that is a number or a date/time - it's just a min and max range, so easy to do.
* Any string that has less than X number of occurrences. I'd default to like 10-ish, but then give users an option to use a different value here, so they could choose to summarize all fields that have less than 40 if they want, or less than 5. 

Should be sure to capture simple arrays as well. And then like @m-mohr says, ignore datetime, license, provider, title, description, etc. 

For letting a user choose, you probably need to give them options to list the fields to either deny or allow. 

Note that @volaya may have some Planet cycles to look into this, so @lossyrob and @m-mohr - do sound in soon if you have ideas.

m-mohr:
That all makes a lot of sense.

I'm currently implementing stac-fields: https://github.com/stac-utils/stac-fields
I think we could use the fields.json in the package to figure out details about the summaries, e.g. whether the arrays should be merged (e.g. eo:bands) or not (e.g. cube:dimensions). I could also indicate there which fields should not be summarized and other information that may be required to generate meaningful summaries. Then you have a machine-readable source of truth for summaries across all libraries.

cholmes:
> That all makes a lot of sense. A default limit of 10 might be a bit too low and it can also apply to objects, not just strings. For example, a limit of 10 would not summarize the bands for Sentinel-2, which has 10+ bands. I'd rather go for 20 or 25 or so.

[...]

PR #264, opened 2021-02-21
Added support for summaries

volaya:
[PR not to be merged yet]

This PR implements the mechanism for generating summaries automatically for Collection objects.

It uses the fields.json file mentioned in https://github.com/stac-utils/stactools/issues/45 to know which attributes should be summarized.

However, that file doesnt provide enough information, and it should be improved before merging this PR. Summaries generated with the current code in this branch are not optimal.

For instance, for numeric values, there should be some information about whether to calculate stats or just collect all existing values (as it happens with epsg codes)


pystac/collection.py
```diff
@@ -9,6 +12,16 @@
 from pystac.utils import datetime_to_str
 
 
+fieldsfilename = os.path.join(os.path.dirname(__file__),
+                              "resources", "fields-normalized.json")
```
m-mohr, line 16: Would it make sense to load it from a CDN by default? That file is expected to change pretty often (mostly to add new fields)...

volaya:
The proposal made in stac-fields looks good to me. That is, to not just have true or false for the summary property, but some details about how to create the summary.

About how to handle the fields.json file, I think we can easily implement some flexible approach like this:

- If the user does not provide a path to a custom fields.json file, a remote common one is used (or a local one shipped with the library, as a fallback option in case there is no connection)
- If a path to a json.file is provided, that is used instead, so there is a way to easily have a custom way of summarizing values
- We can provide some predefined files with subgroups of parameters (like those 'core', 'extended' and 'all' that @cholmes mentioned), and have a simple way of using them (in the command-line tool, that can be by defining some specific argument for those hardcoded files)

Let me sketch something like that in this same branch, so we can discuss based on it.

volaya:
I adapted the code that I had done before (which basically computes and updates the summaries when an item is added to a collection), to the recent changes in the collection module. The code that takes care of summaries is now in its own module.

It accepts a file with field definitions (for now, those fields are not assumed to contain information on how to summarize their values, but just whether or not to do it, as it happened before). If no file is passed, it picks the code from the CDN, or uses a local file if that's not available.

CI fails, but gives no information about the test that is failing. I am not seeing that issue when running tests locally, so I will try to figure out what is going on.

lossyrob:
@volya are you not able to see the github action logs? Clicking on the "Details" for one of the fails of the checks brings me to the GitHub Actions build logs, where I can see the "Execute linters and test suite" failed, and the bottom of the logs have:

```
+ echo ' -- CHECKING TYPES WITH PYRIGHT --'
+ echo
+ scripts/pyright pystac tests
Checking node version...
Checking node_modules dir...
Checking pyright exists...
...installing pyright
/usr/local/bin/pyright -> /usr/local/lib/node_modules/pyright/index.js
/usr/local/bin/pyright-langserver -> /usr/local/lib/node_modules/pyright/langserver.index.js
+ pyright@1.1.138
added 1 package from 1 contributor in 0.886s
done.
No configuration file found.
No pyproject.toml file found.
stubPath /home/runner/work/pystac/pystac/typings is not a valid directory.
Assuming Python platform Linux
Searching for source files
Found 73 source files
/home/runner/work/pystac/pystac/pystac/summaries.py
  /home/runner/work/pystac/pystac/pystac/summaries.py:22:24 - error: No overloads for "min" match the provided arguments
    Argument types: (T@RangeSummary, T@RangeSummary) (reportGeneralTypeIssues)
  /home/runner/work/pystac/pystac/pystac/summaries.py:23:24 - error: No overloads for "min" match the provided arguments
    Argument types: (T@RangeSummary, T@RangeSummary) (reportGeneralTypeIssues)
2 errors, 0 warnings, 0 infos 
Completed in 6.079sec
Error: Process completed with exit code 1.
```

Are you running `scripts/test` locally, which runs pyright for type checking? I would assume this should also fail locally.

volaya:
Yes, I can see the log, but it doesn't tell me which test is failing or the full error stack trace. And I cannot figure out how a RangeSummary object might arrive to that line...

pystac/collection.py
```diff
@@ -598,6 +518,14 @@ def __init__(
     def __repr__(self) -> str:
         return "<Collection id={}>".format(self.id)
 
+    def create_summaries(self, summarizer: Summarizer = None):
```
lossyrob, line 521:
```suggestion
    def create_summaries(self, summarizer: Optional[Summarizer] = None) -> None:
```

pystac/summaries.py
```diff
@@ -0,0 +1,170 @@
...
+                    if v not in listsummary:
+                        listsummary.append(v)
+                    summaries.add(k, listsummary)
+
+
+class Summaries:
+    def __init__(self, summaries: Dict[str, Any], summarizer: Summarizer = None) -> None:
+        self._summaries = summaries
+
+        self.summarizer = summarizer or Summarizer()
```
lossyrob, line 109: There's currently no caching for the properties json so this would fetch a new document for every collection, which has performance implications.

lossyrob (commented):
One high level comment is that I'm not sure Summarizer should be a concept that either Collection or Summaries know about - the default `create_summaries` is destructive to any manual summaries that may have been added, and it's not clear how to flexibly use the summarizer with methods against the objects. There's also the issue with multiple transparent fetches mentioned below.

I think generating summaries from the properties JSON is a specific enough task that having some independent flexible API around it might be in order. I.e. having the way to generate Summaries is to use a Summarizer directly, instead of having that functionality wrapped in methods.

e.g. an API that is used like:

```python
import pystac
from pystac.summaries import Summarizer

collection: pystac.Collection = ...
new_items: List[pytsac.Item] = ...

summerizer = Summarizer()

# Example 1: Create from strach
collection.summaries = summerizer.summerize(collection)

# Example 2: Create summaries and merge (generated have precedent)
collection.summaires.update(summarizer.summerize(collection)

# Example 3: Create summaries and merge (existing have precedent)
collection.summaries = summarizer.summerize(collection).update(collection.summaries)

# Example 4: Update summaries with new items before adding to collection
new_summaries = summarizer.summerize(new_items)
collection.summaries.merge(new_summaries)
collection.add_items(new_items)
```

Here summarizer has a `summarize` method:
```
def summerize(source: Union[pystac.Collection, Iterable[pystac.Item]) -> Summaries:
   ...
```

There's a difference between `update` (where the summary for each summary key is overriding any existing summary) and `merge` (which expands the summary to include the information from the incoming summaries for overlapping keys). I'm not sure these would be the best name, but identifying the different use cases to me means the API is complex enough that we might want to keep the generation of summaries and the update/merging of summaries as independent steps (Summarizer only generates summaries, summaries know how to update themselves or merge in new summaries). 

I think also breaking out the Summaries merge/update methods to be independent of the `stac-fields` based summary generation logic would allow those methods to be used independently, if a user has their own method of creating summaries that they want to merge together.

volaya:
Yes, you are right, better to have something independent for summaries. My idea was to have some method of automatically creating summaries when adding elements to a collection, but it might be a better approach to let the user choose how to do that, and just provide the functionality to do it.

Let me update the PR with your ideas and we can continue the discussion. Thanks for the comments!

lossyrob:
@volaya the failure is not on a unit test, but on a static type checking. There's linting and type checking in `scripts/test`, it doesn't only run unit tests.

```
 /home/runner/work/pystac/pystac/pystac/summaries.py:22:24 - error: No overloads for "min" match the provided arguments
    Argument types: (T@RangeSummary, T@RangeSummary) (reportGeneralTypeIssues)
  /home/runner/work/pystac/pystac/pystac/summaries.py:23:24 - error: No overloads for "min" match the provided arguments
    Argument types: (T@RangeSummary, T@RangeSummary) (reportGeneralTypeIssues)
```

In summaries.py on line 22, there's a call to `min(...)` that deals with values of type `T`. Since `T` is generic, there's no guarantee that the type of v or self.minimum would be something you could put into `min`. Solving this would require putting a type constraint on `T` that ensures the types it allows would have the ability to be ordered/compared. 

Being able to specify that generic types are sortable is apparently been a pain; see https://github.com/python/typing/issues/59 and [this stackoverflow answer](https://stackoverflow.com/a/65224102/841563) for a solution

volaya:
Ah, ok. Thanks for that hint. I didn't know that there were type checks, and I was a bit puzzled with that error message...

volaya:
@lossyrob I have Updated the PR with your ideas.

Collections do not know about summaries, and the updated/combine operation is done by the Summaries class. Creating a summary is done by the Summarizer class.

One comment about that: I think the Summarizer should take care of performing the merge/combine, since I believe that those operations (or at least the combine one) must depend on the stac-fields json. Otherwise, you can combine summaries that do not have the same type of summary for a given value (for instance one in which a numerical field is summarized with a range, and another one where it is summarized with more detailed stats). If that's ok, then I guess we need some to add some checking in the corresponding methods. But probably it's better to move them to the Summarizer class. Without knowing how to summarize a given field (which is at the moment known only by the Summarizer class, thanks to the json file with filed definitions), one cannot know how to combine existing summaries for that field.

Also, notice that, in the current design of the Summaries class, you can use the same id for summaries of different type (let's say, for a range and for a list). That might happen if trying to update/combine two summaries that have summarized a given field in different ways.

Maybe it can be a good idea to not store lists, ranges, schemas and other in separate dicts as instance variables of the Summaries class, but just keep a single dictionary of summaries and have several classes for them (like we have the RangeSummary class now).

Let me know what you think and I can start adding some unit tests, once we have the approach for this classes more or less ready.

volaya:
CI still has issues with the typing. The solution you linked to doesn't seem to work (however, I tried locally with mypy, and that fix makes it happy...) I will see if I can manage to make it work somehow

volaya:
I also added the idea mentioned in https://github.com/stac-utils/pystac/issues/178 about having a limit for the number of elements in list summaries

m-mohr:
@volaya stac-fields will return more details on summaries once the CDN picks up the changes (usually within 24hours), see also https://github.com/stac-utils/stac-fields/issues/6

pystac/summaries.py
```diff
@@ -62,6 +64,14 @@ def from_dict(cls, d: Dict[str, Any], typ: Type[T] = Any) -> "RangeSummary[T]":
                                       "fields-normalized.json")
 
 
+class SummaryStrategy(Enum):
+    ARRAY = "v"
+    RANGE = "r"
+    SCHEMA = "s"
+    DONT_SUMMARIZE = False
+    UNDEFINED = True
```
m-mohr, line 72:
```suggestion
    UNDEFINED = True
```

I think "default" or "detect" would be better choices as name. There are rules defined in the fields json readme for the case it's not set, which depends on the data type mostly.

volaya:
@m-mohr I have implemented here the ideas that you added in https://github.com/stac-utils/stac-fields/issues/6

pystac/summaries.py
volaya: changed

pystac/summaries.py
```diff
@@ -61,7 +61,7 @@ def from_dict(cls, d: Dict[str, Any], typ: Type[T] = Any) -> "RangeSummary[T]":
 FIELDS_JSON_URL = "https://cdn.jsdelivr.net/npm/@radiantearth/stac-fields/fields.json"
 
 FIELDS_JSON_LOCAL_PATH = os.path.join(os.path.dirname(__file__), "resources",
-                                      "fields-normalized.json")
```
m-mohr: fields-normalized.json is the better choice for non-JS components as all dependencies have been resolved...

pystac/summaries.py
```diff
@@ -0,0 +1,262 @@
...
+                    jsonfields = json.load(f)
+                self._set_field_definitions(jsonfields)
+            except:
+                self._set_default_field_definitions()
+
+    def _set_default_field_definitions(self) -> None:
+        if not Summarizer.__default_field_definitions:
+            try:
+                with urllib.request.urlopen(FIELDS_JSON_URL) as url:
+                    type(self).__default_field_definitions = json.loads(url.read().decode())
```
lossyrob, line 105: Curious about the use of `type(self)` here and `Summarizer.__default_field_definitions` below. Using `Summarizer.__default_field_definitions` would be more consistent with other places in the lib; also we've generally done single underscore for private vars and not double underscore

pystac/summaries.py
```diff
@@ -0,0 +1,262 @@
...
+    '''
+
+    __default_field_definitions: dict = {}
+
+    def __init__(self, fields: Optional[str] = None):
+        if fields is None:
+            self._set_default_field_definitions()
+        else:
+            try:
+                with open(fields) as f:
```
lossyrob, line 98: Should use `pystac.StacIO.default().read_json(fields)` here

pystac/summaries.py
```diff
@@ -0,0 +1,262 @@
...
+                    jsonfields = json.load(f)
+                self._set_field_definitions(jsonfields)
+            except:
+                self._set_default_field_definitions()
```
lossyrob, line 102: I feel like this should throw the error and not result in a silent fallback to the default - if someone were to pass in fields and have expectations that they would be used, this could be a very tricky thing to debug from results downstream

pystac/summaries.py
```diff
@@ -0,0 +1,262 @@
...
+    def _set_default_field_definitions(self) -> None:
+        if not Summarizer.__default_field_definitions:
+            try:
+                with urllib.request.urlopen(FIELDS_JSON_URL) as url:
+                    type(self).__default_field_definitions = json.loads(url.read().decode())
+            except:
+                pass
+        if not Summarizer.__default_field_definitions:
+            with open(FIELDS_JSON_LOCAL_PATH) as f:
+                type(self).__default_field_definitions = json.load(f)
```
lossyrob, line 115: Other places in the library rely on reading remote URLs (e.g. validation); I'm not sure it's worth including PySTAC's own copy of fields for the case where the remote URL fails. If users want to use a local copy they can pass it in manually themselves. I'd remove this conditional and remove the local copy of fields from the resources.

lossyrob (commented):
Added some comments; pushing up some commits to solve the CI issues

lossyrob:
Added some commits, but the clock on my WSL vm was out of whack so they appear older than they are. The commits I added start with #a7e1747

volaya:
@lossyrob Good suggestions. I added the changes that you proposed.

lossyrob:
Looks like [Protocol](https://docs.python.org/3/library/typing.html#typing.Protocol) is new in 3.8 . Not sure how the windows CI builds are passing for pre-3.8 builds. I'll look into another solution that works across python versions for the `Comparable` issue

lossyrob:
I just pushed a commit that will conditionally install `typing_extensions`, which allows the import of `Protocol` in pre-3.8 versions of Python. This adds an extra dependency in those situations, but does not add any new dependencies for Python 3.8 or later.

lossyrob:
Merged in upstream/main, and made some changes to pass the mypy CI check that was added. This included removing the `typ` parameter from `RangeSummary.from_dict`.

lossyrob (approved):

lossyrob:
Changes approved, but this does need a CHANGELOG entry @volaya. Should probably change the PR title and description too to reflect that it's ready. Other then that I think this is good to merge!

pystac/summaries.py
```diff
@@ -0,0 +1,270 @@
...
+        self.maximum = max(self.maximum, v)
+
+    @classmethod
+    def from_dict(cls, d: Dict[str, Any]) -> "RangeSummary[T]":
+        minimum: T = get_required(d.get("minimum"), "RangeSummary", "minimum")
+        maximum: T = get_required(d.get("maximum"), "RangeSummary", "maximum")
+        return cls(minimum=minimum, maximum=maximum)
+
+
+FIELDS_JSON_URL = "https://cdn.jsdelivr.net/npm/@radiantearth/stac-fields/fields.json"
```
m-mohr, line 65:
```suggestion
FIELDS_JSON_URL = "https://cdn.jsdelivr.net/npm/@radiantearth/stac-fields/fields-normalized.json"
```

This must be the normalized version, otherwise, you'll run into bugs here as some fields may not be resolved yet.

pystac/summaries.py
```diff
@@ -0,0 +1,270 @@
...
+    """
+
+    _default_field_definitions: Dict[str, Any] = {}
+
+    def __init__(self, fields: Optional[str] = None):
+        if fields is None:
+            self._set_default_field_definitions()
+        else:
+            jsonfields = pystac.StacIO.default().read_json(fields)
+            self._set_field_definitions(jsonfields)
```
duckontheweb, line 96: Since this is going to get called any time a `Summarizer` is instantiated without a `fields` argument, maybe it makes sense to move the logic for fetching the JSON in `_set_default_field_definitions` into a separate function and use `functools.lru_cache` to cache this across all instances. 

It seems like we could re-use that cached function for both of these cases (with or without `fields` argument), which would help with performance in the case where they are providing a custom URL and instantiating the `Summarizer` multiple times.

Something like:

```python
@lru_cache
def _get_fields_json(url: str) -> Dict[str, Any]:
    return pystac.StacIO.default().read_json(url)
```

pystac/summaries.py
```diff
@@ -0,0 +1,270 @@
...
+        self, source: Union["Collection_Type", Iterable["Item_Type"]]
+    ) -> "Summaries":
+        """Creates summaries from items"""
+        summaries = Summaries.empty()
+        if hasattr(source, "get_all_items"):
+            for item in source.get_all_items():  # type: ignore[union-attr]
+                self._update_with_item(summaries, item)
+        else:
+            for item in source:  # type: ignore[union-attr]
+                self._update_with_item(summaries, item)
```
duckontheweb, line 169: This satisfies the type checker without needing to ignore the error. 

```suggestion
        if isinstance(source, Collection):
            for item in source.get_all_items():
                self._update_with_item(summaries, item)
        else:
            for item in source:
                self._update_with_item(summaries, item)
```

pystac/summaries.py
volaya: Actually, the case of the default fields definition being downloaded multiple time was already avoided, since the variable storing that was a class variable, and fetching the definition file was only done if that variable was none. But I think that your solution is much better and will also cache non-default files, so I have changed the mechanism.

I have also simplified things a bit, using only the `read_json `method from `StacIO`

pystac/summaries.py
volaya: done

volaya: Thinking about this...what happens if we are dealing with a Catalog instead of a Collection? It also has the `get_all_items' method. Is that a valid type of element for this? Should we extend that `isinstance` statement to check for both classes?

pystac/summaries.py
duckontheweb: > Actually, the case of the default fields definition being downloaded multiple time was already avoided, since the variable storing that was a class variable, and fetching the definition file was only done if that variable was none.

Right! Sorry I missed that.

pystac/summaries.py
duckontheweb: It looks like [as of 1.0.0-rc.4](https://github.com/radiantearth/stac-spec/blob/master/CHANGELOG.md#v100-rc4---2021-05-11) Catalogs no longer support summaries, so I think it makes sense to restrict this to only Collections.

duckontheweb:
@volaya The CI failures are a little perplexing since it's failing on a file that doesn't even exist in your branch 🤷 . I've merged in the latest from `main` and fixed those issues.

duckontheweb (commented):
Thanks again for this. I think the only remaining items are:

- [ ] Update the PR title to reflect the new status
- [ ] Add a CHANGELOG entry

duckontheweb (approved):
Awesome, thanks again. This is will be a big addition to the next release!

Issue #230, opened 2020-10-30
Handle setting None consistently in extensions.


lossyrob:
In extension classes there is an `apply` method for setting all the properties of the extension at once. These will take optional parameters for optional properties, normally defaulting to `None`.

We should be consistent across all the extensions about whether or not to use the None value as an indication that the optional property should be unset.

We should also ensure that any optional property setters will take None as a value and unset the property accordingly.

lossyrob:
Fixed by #272

PR #270, opened 2021-03-03
added support for STAC file extension

volaya:
It adds support for data_type, nodata, checksum and size

The tests only cover the support for values in assets, since I am not sure how to apply (or even if it makes sense to ) those parameters can be applied outside of the asset level...

### DIFF
--- a/pystac/extensions/file.py
+++ b/pystac/extensions/file.py
@@ -65,6 +65,13 @@ class FileItemExt(ItemExtension):
         self.nodata = nodata
         self.checksum = checksum
 
+    def _set_property(self, key, value, asset):
+        target = self.item.properties if asset is None else asset.properties
+        if value is None:
+            target.pop(key, None)
+        else:
+            target[key] = value
+
     @property
     def data_type(self):
         """Get or sets the data_type of the file.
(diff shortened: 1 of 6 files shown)


pystac/extensions/file.py
```diff
@@ -0,0 +1,235 @@
...
+        else:
+            return asset.properties.get('file:size')
+
+    def set_size(self, size, asset=None):
+        """Set an Item or an Asset size.
+
+        If an Asset is supplied, sets the property on the Asset.
+        Otherwise sets the Item's value.
+        """
+        if asset is None:
```
lossyrob, line 142: We should handle setting optional values to `None` by popping the dict keys - I've fixed an issue in https://github.com/stac-utils/pystac/pull/269/files that has a pattern for this which you could reuse here.

volaya: @lossyrob Instead of replacing it in all the setters, I have extracted the mechanism to a separate method `_set_property`. Maybe it would be worth to have that in the parent Item class? Other extensions might have the same issue (I actually used the `eo` extension as the base for the `file ` one, so that one for sure has the same problem with None values in setters.

Let me know if you think it's a good idea and I can take care of that bit of refactoring

lossyrob:
Looks like `codespell` is complaining about "FileTest" - you can add it to the [codespell ingore file](https://github.com/stac-utils/pystac/blob/develop/.codespellignore) (will be the first entry!).

Afterwards I suggest making sure `scripts/test` pass; this is the same script that CI is running

pystac/extensions/file.py
lossyrob: Sounds like a great idea!

lossyrob: Seems like I captured the idea of doing this in https://github.com/stac-utils/pystac/issues/230, so this refactor will close that issue.

lossyrob (approved):

Issue #215, opened 2020-10-27
Don't use _pathlib in utils.py


schwehr:
Don't do this in `utils.py`:

```python

import os

# [SNIP]

# (i.e. testing Windows path manipulation on non-Windows systems)
_pathlib = os.path
```

instead do: 

```python
import os
```

And in testing, use `@mock.patch('os.patch')`...

```python
import mock

# [SNIP]
    @mock.patch('os.path')
    def test_make_relative_href_windows(self, mock_path):
        # TODO: set behavior of mock_path.
```

jsignell:
Just a note that when I started working on #619 in #1002 I removed `_pathlib`. I'm not sure we even really need to patch the os patch since we can just rely on CI to run the tests on the correct architecture.

PR #271, opened 2021-03-03
Ensure correct url format from local files

volaya:
Not sure if others are having the same problem or I am missing something, but without this fix all tests fail in Windows.

The reason for it is in here:

https://github.com/stac-utils/pystac/blob/develop/pystac/stac_io.py#L19

The urlopen function does not understand a windows path such as `c:\folder\file.ext`, and raises an exception.

### DIFF
--- a/pystac/stac_object.py
+++ b/pystac/stac_object.py
@@ -1,5 +1,6 @@
 from abc import (ABC, abstractmethod)
 from enum import Enum
+from urllib.request import pathname2url
 
 import pystac
 from pystac import STACError
@@ -521,6 +522,7 @@ class STACObject(LinkMixin, ABC):
         """
         if not is_absolute_href(href):
             href = make_absolute_href(href)
+        href = f"file:{pathname2url(href)}"
         d = STAC_IO.read_json(href)
 
         if cls == STACObject:


pystac/stac_object.py
```diff
@@ -521,6 +522,7 @@ def from_file(cls, href):
         """
         if not is_absolute_href(href):
             href = make_absolute_href(href)
+        href = f"file:{pathname2url(href)}"
```
lossyrob, line 525: This breaks tests - perhaps a better approach would be to put this line directly in the STAC_IO method you point to in the body of the PR?

lossyrob:
Hm, we've had issues with Windows paths in the past, but I thought we got past that. I don't believe this is broken for all cases of windows paths; maybe it's possible to put up a unit test that shows the problem, even when running on non-windows OS's? We use [ntpath](https://github.com/stac-utils/pystac/blob/develop/tests/test_utils.py#L36) to test in unit tests, although there's an issue around [using a different technique](https://github.com/stac-utils/pystac/issues/215) which isn't implemented yet.

pystac/stac_object.py
volaya: Let me try to work out a different way of doing this. In my case, all test pass with this fix, so I guess the behaviour in the CI and in my machine (or any windows machine, for that matter), might be different.

volaya:
I don't think that npath is the problem or affects this. The issue is with the path starting with the drive name and a colon (C:). With that, urlopen (in the line that i pointed out) understands that the drive name is the protocol of the url, and raises the exception. So the paths are correctly formed, that works fine, but without the "file://" prefix urlopen does not work.

lossyrob:
Gotcha. Perhaps the case with a drive prefix can be handled prior to the https://github.com/stac-utils/pystac/blob/develop/pystac/stac_io.py#L19 line instead of how it's currently implemented to avoid the unit test failure.

duckontheweb:
@volaya I got the Windows CI builds running properly in #361 and noticed this issue as well. I solved it there by using `pystac.utils._urlparse` instead of `urllib.parse.urlparse`, since that method handles the Windows schemes. See d4e93acce8b6c6b26681698d4c629a9df3992b8d for the change. 

There still appear to be a number of other path-related issues on Windows, so I opened up #362 to track fixing the rest of those.

duckontheweb:
@volaya I'm going to close this, as it's superseded by #361. Feel free to open an issue if you continued to see problems with any of the tests on Windows.

PR #272, opened 2021-03-04
Extract method to correctly handle setting properties in Item/Asset for ItemExtensions

volaya:
This PR adds a method to the ItemExtension class, that makes it easier and more robust for derived classes to set properties in Assest and/or Items. 

Existing extensions needing that mechanism have been refactored to rely on the parent method instead

This was discussed in https://github.com/stac-utils/pystac/pull/270

If this PR is merged, https://github.com/stac-utils/pystac/pull/230 will be no longer needed


alkamin (approved):
This is a great improvement.

lossyrob (approved):
🎉

PR #273, opened 2021-03-10
Remove unnecessary deepcopies

moradology:
Calls to `deepcopy` are fairly expensive when working with large number of STACs. In `to_dict` methods, a dictionary `d` is instantiated only to have `deepcopy` applied to it for the `return` value. This seems to be unnecessary.

Here's `snakeviz` output of some `cProfile` timings.
Before:
image

After:
image

It should be noted that the after includes also the use of `orjson` to optimize ser/de operations. This optimization can be left out of `pystac` because it requires an external dependency and because it can easily be plugged in through a `STAC_IO` subclass


matthewhanson:
nice catch @moradology , this looks good to me, there doesn't seem to be any reason to do a deepcopy of a locally defined variable before returning, 
I'm also interested in improving the performance of PySTAC with large catalogs and interested in any optimizations you've tried.  Will open up new issue.

matthewhanson (approved):

PR #288, opened 2021-03-17
added type property for catalogs and collections

volaya:
**Related Issue(s):** #
 https://github.com/radiantearth/stac-spec/pull/971

**Description:**
Adds tpe property for catalogs and collections

**PR Checklist:**

### DIFF
--- a/pystac/catalog.py
+++ b/pystac/catalog.py
@@ -366,6 +366,7 @@ class Catalog(STACObject):
             links = filter(lambda l: l.rel != 'self', links)
 
         d = {
+            'type': self.STAC_OBJECT_TYPE.value.title(),
             'id': self.id,
             'stac_version': pystac.get_stac_version(),
             'description': self.description,
--- a/pystac/serialization/identify.py
+++ b/pystac/serialization/identify.py
@@ -264,18 +264,25 @@ def identify_stac_object_type(json_dict):
     """
     object_type = None
 
-    # Identify pre-1.0 ITEMCOLLECTION (since removed)
-    if 'type' in json_dict and 'assets' not in json_dict:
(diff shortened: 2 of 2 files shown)


volaya:
Before merging this, one thing to consider:

Pystac uses the 'extent' property to identify collections (see https://github.com/stac-utils/pystac/blob/develop/pystac/serialization/identify.py#L273). That should be changed to now use the type property, but doing so breaks a large number of tests, mainly because sample json files in tests do not have the type property. Those files should be updated, but before working on that, I would like a second opinion about it, to know if it is worth or not. For what I understand, the type property is now mandatory, so sample test catalogs and collections should have it.

matthewhanson:
@volaya That's correct, this is a required property now, and Pystac should be updated to use the type to identify Collections.  I think all the tests should be updated as well.  I think I'd combine the identify changes and example changes into this PR.

lossyrob:
The identify functionality identifies older version of STAC as well, so we don't want to get rid of that. It can start with a check for the property `type` and make decisions based on that, but if something is pre-1.0.0-RC1, it will need to fall back to it's current logic.

The STAC objects that should be updated are the main test data, but all the `examples` test data that are for a specific version shouldn't be changed - those test PySTAC's ability to migrate older versions to new STAC. 
r

volaya:
I added a new commit with changes to the identify function, which now uses the approach suggested by @lossyrob (first try to identify based on the type property, and if that's not possible, use the previous approach).

This also fixes a problem not mentioned in https://github.com/stac-utils/pystac/pull/289 With that PR, Collections can now have assets, and the identify method uses the presence of the 'assets' property to identify Item objects, which will no longer be a valid strategy. No need to change anything in that PR if this one is merged.

lossyrob (approved):

PR #289, opened 2021-03-17
Implement collection assets as a core property of collections

volaya:
This is still a work in progress. I have to add more tests and check some failures in preexisting ones

**Related Issue(s):** 

#266 

**Description:**

This PR adds assets as core properties of collections, according to the STAC 1.0 specification

**PR Checklist:**


pystac/asset.py
```diff
@@ -0,0 +1,244 @@
+from copy import copy
+
+from pystac.stac_object import STACObject
+from pystac import STACError
+from pystac.utils import (is_absolute_href, make_absolute_href, make_relative_href)
+
+
+class STACObjectWithAssets(STACObject):
```
lossyrob, line 8: Is there a reason to use this type of inheritance instead of a Mixin, similar to the pattern that [LinkMixin](https://github.com/stac-utils/pystac/blob/develop/pystac/stac_object.py#L94) uses?

volaya: Good catch. I will change it to that approach

volaya:
@lossyrob You were right in that the problem was in how the missing extensions were identified. It was not correctly figuring out that part, and the procedure to modify the assets property was not being executed. Please check how I solved it, to ensure that it is correct.

pystac/asset.py
volaya: done

pystac/asset.py
```diff
@@ -0,0 +1,248 @@
...
+        """
+        asset.set_owner(self)
+        self.assets[key] = asset
+        return self
+
+    def make_asset_hrefs_relative(self):
+        """Modify each asset's HREF to be relative to this object's self HREF.
+
+        Returns:
+            STACObjectWithAssets: self
```
lossyrob, line 36: docstring references to removed type

lossyrob (changes requested):
Looks good, I merged in a change from develop that affected the AssetMixin. One note about a docstring having an old class, but otherwise I think this is good to go.

I may try to cut a final release of 1.0.0-beta.2 PySTAC with a couple of bug fixes before merging this in.

pystac/asset.py
volaya: I removed all references to the STACObjectWithAssets class

lossyrob:
These changes were integrated into #309

Issue #284, opened 2021-03-16
Normalize URLs and set link types for newly added entities


matthewhanson:
When you save a catalog it walks the tree and saves only those objects which are resolved...i.e., they are stored in memory. Which makes sense, no need to read the file in only to save it again.

However, when calling `normalize_hrefs` or `normalize_and_save`, all links are resolved and function mutates the entire tree.

This makes updates to a catalog difficult. If I have an existing catalog and add subcatalogs/items to it, it's hard to save those. I need to use `save_object`, but then have to include the code for generating the proper new self hrefs.  Also, when combined with `generate_subcatalogs` this becomes pretty tricky and I'm still trying to figure out the best way to do it.  I kept having trouble trying to set the self hrefs and link type to relative on a set of subcatalogs.

It seems like if I add a new entity to an existing catalog, part of that add should be to set the self_href, and the link type that points to it (relative or absolute), so that I can then later call just the `save` function on my catalog.   Since the only resolved item links are the new items I added this efficiently saves just the newly added items and ignores everything else. 

Somewhat related to this is would be nice to be able to easily "unresolve" entities, which would allow me to save periodically - I could save the catalog, add some items, save again, add more items, etc.

matthewhanson:
The description above conflates a few issues so to keep this focused on the normalization of URLs for STAC Objects in a catalog.

When you normalize hrefs it resolves all [`children`](https://github.com/stac-utils/pystac/blob/5a492ca37ef6d8b0b2ea10db35150e72a709a398/pystac/catalog.py#L499) and [`items`](https://github.com/stac-utils/pystac/blob/5a492ca37ef6d8b0b2ea10db35150e72a709a398/pystac/catalog.py#L487) over the whole tree (or at least the branch of the tree you are own). This is unnecessary and can become a problem when an existing catalog is large and only a few items are added.

The self link for all added objects should automatically be set in the [`add_child`](https://github.com/stac-utils/pystac/blob/5a492ca37ef6d8b0b2ea10db35150e72a709a398/pystac/catalog.py#L151) and [`add_item`](https://github.com/stac-utils/pystac/blob/5a492ca37ef6d8b0b2ea10db35150e72a709a398/pystac/catalog.py#L180) functions. An existing catalog already uses the same base URL, accessible via the root catalog.

matthewhanson:
@lossyrob 
I think I might be missing something here, probably something with the caching, but I'm wondering why normalization needs to walk the catalog at all.

[...]

PR #290, opened 2021-03-18
Remove link type

### LINKED ISSUE #285: Saving a branch of a catalog with Normalized hrefs saves a self link
When saving a catalog with relative links `self` links aren't included, except for the root catalog.

It's useful to call save() on a collection, or sub-catalog, in order to save just that branch (for instance using multiple processes where each one is adding items to collections and saving the collection). But when save() is called, the `self` absolute link is added.

I think that when save is called, it should check if it's the root catalog before it adds the self link

### LINKED ISSUE #286: Remove link_types
The save function always loads the entire catalog before saving anything new in order to set [absolute or relative URLs](https://github.com/stac-utils/pystac/blob/5a492ca37ef6d8b0b2ea10db35150e72a709a398/pystac/catalog.py#L613). 

For hierarchical links (root, child, parent, item, items), `link_type` will always be the same, so it seems unnecessary to crawl the catalog to set those link types.

Most of the occurrences of `link_type` in the code are for passing it around and setting it. The only place it's used is for [`Link.get_href`](https://github.com/stac-utils/pystac/blob/5a492ca37ef6d8b0b2ea10db35150e72a709a398/pystac/link.py#L106).

I think link_type as a Link field can be entirely removed and replaced with a heuristic. Some assumptions:

- The root catalog, and only the root catalog should determine if member objects should use relative or absolute links
- Relative links should only be used for hierarchical links (root, parent, child, item, items, self)
- Relative self links are just an identity and can be left out
- Non-hierarchical links should always use absolute links
- If a STAC object does not have any hierarchy (it has no parents, root) relative links make no sense so they should be absolute
- If the STAC object does have hierarchical links there will always be a root catalog

If these assumptions hold true then if the link has a rel type of ['root', 'parent', 'child', 'item', 'items', 'self'] and the object has a root catalog and the root catalog uses relative links then return relative link, otherwise return an absolute link.

matthewhanson:
**Related Issue(s):** #
#286 , also resolves #285

**Description:**
This PR removes the link_type field and the LinkType class, along with the need to loop through all links to set them as RELATIVE or ABSOLUTE.
Instead, when a link href is requested it returns an absolute link unless all the following are true:

- The link is a hierarchical link ('root', 'child', 'parent', 'item')
- The link has an owner (i.e., it is part of a STAC object and not a standalong Link instance)
- The owner has a root catalog
- The root catalog.catalog_type is CatalogType.RELATIVE_PUBLISHED

When all are true the link relative to the owner is returned.

There are some additional implications and related changes:
- Catalog.catalog_type is now always defined, never None. It defaults to ABSOLUTE_PUBLISHED
- A user can just set `Catalog.catalog_type = Catalog.Type.X`...subsequent saves will generate relative links for all saved objects
- Specifying CatalogType to `normalize_and_save` and `save` is no longer needed, in fact it should be deprecated.  It's been left in for backward compatibility. The catalog_type of the root should be favored.
- An implication/assumption is that unless you are working with a stand-alone Item or Collection, there will always be a root catalog. 
- Fixed #286 so that now you can seamlessly save any node in a tree which will save it from that point down

I've not yet updated tests, instead I'm putting this WIP PR up for discussion since it's a big internal change.  Afterward I can update tests

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,19 @@
 
 ### Added
 
+- HIERARCHICAL_LINKS array constant of all the types of hierarchical links (self is not included)
+
 ### Fixed
 
 - Fixed error when accessing the statistics attribute of the pointcloud extension when no statistics were defined ([#282](https://github.com/stac-utils/pystac/pull/282))
 
 ### Changed
 
+- Link behavior - link URLs can be either relative or absolute. Hierarchical (e.g., parent, child) links are made relative or absolute based on the value of the root catalog's `catalog_type` field
+
 ### Removed
+
+- Removed LinkType class and the `link_type` field from links
 
 ## [v0.5.5]
 
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -155,14 +155,17 @@ Relative vs Absolute Link HREFs
 
 Absolute links point to their file locations in a fully described way. Relative links
 are relative to the linking object's file location. For example, if a catalog at
-``/some/location/catalog.json`` has a link to an item that has an HREF set to ``item-id/item-id.json``, then that link should resolve to the absolute path ``/some/location/item-id/item-id.json``.
+``/some/location/catalog.json`` has a link to an item that has an HREF set to ``item-id/item-id.json``,
+then that link should resolve to the absolute path ``/some/location/item-id/item-id.json``.
 
-The implementation of :class:`~pystac.Link` in PySTAC allows for the link to be marked as
-``link_type=LinkType.ABSOLUTE`` or ``link_type=LinkType.RELATIVE``. This means that,
-even if the stored HREF of the link is absolute, if the link is marked as relative, serializing
-the link will produce a relative link, based on the self link of the parent object.
+Links are set as absolute or relative HREFs at save time, as determine by the root catalog's catalog_type
+:attribute:`~pystac.Catalog.catalog_type`. This means that, even if the stored HREF of the link is absolute,
+if the root ``catalog_type=CatalogType.RELATIVE_PUBLISHED`` or ``catalog_type=CatalogType.SELF_CONTAINED``
+and subsequent serializing of the any links in the catalog will produce a relative link,
+based on the self link of the parent object.
 
-You can make all the links of a catalog relative or absolute using the :func:`Catalog.make_all_links_relative <pystac.Catalog.make_all_links_relative>` and :func:`Catalog.make_all_links_absolute <pystac.Catalog.make_all_links_absolute>` methods.
+You can make all the links of a catalog relative or absolute by setting the :func:`Catalog.catalog_type` field
+then resaving the entire catalog.
 
 .. _rel vs abs asset:
 
--- a/docs/tutorials/adding-new-and-custom-extensions.ipynb
+++ b/docs/tutorials/adding-new-and-custom-extensions.ipynb
@@ -181,7 +181,7 @@
    "source": [
     "The `from_item` class method simply returns a new instance of the item extension given an item.\n",
     "\n",
-    "The `_object_links` class method returns the `rel` string for any links that point to STAC objects like Catalogs, Collections or Items. PySTAC needs to know which links point to STAC objects because it needs to consider them when fully resolving a STAC into in-memory objects. In a lot of cases, extensions don't add new links to STAC objects, so this is normally an empty list; however, if the extension does do this (like the `source` link in the [Label Extension](https://github.com/radiantearth/stac-spec/tree/v1.0.0-beta.2/extensions/label#links-source-imagery)), make sure to return the correct value (like the LabelItemExt is doing [here](https://github.com/azavea/pystac/blob/v0.5.0/pystac/extensions/label.py#L291-L293))."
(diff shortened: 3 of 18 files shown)


lossyrob:
Discussed this with Matt in a video chat; thinking through these and related changes, I think this will be a big improvement and the assumptions are sound. 👍

matthewhanson:
In updating the tests I came across a test case that is using a relative link for a hierarchical link, which would no longer be supported:
https://github.com/stac-utils/pystac/blob/develop/tests/data-files/label/label-example-1.json#L97

While moving assets along with the metadata might be a common case (and relative links for assets still work), I don't think using relative hrefs in a link should be a recommended practice.  Relative links for an asset that's in the same location as the metadata makes sense, but a source link for labels points to another Item which could be in some random part of the tree and having a relative link to that is rather awkward.

I'm fine with not supporting this use case right now.  Alternately we could just keep all non-hierarchical links as is when renormalizing....since a renormalization is typically going to be over the whole catalog and the relative link will still be valid. 
I want to avoid trying to programmatically update relative links when renormalizing, for one it means we'll need to move back to using a link_type, but also because I think there's too many things that could go wrong that in practice this is not a valuable feature.

@lossyrob

lossyrob:
@matthewhanson using relative links to go to different parts of a Catalog tree is necessary for self-contained ML datasets. Otherwise you'll have to use absolute links to link a Label Item to it's source imagery, and then you can no longer copy the catalog around and have it be valid. I think this is a use case we need to support.

This is the sort of use case that causes cycles in the graph that is otherwise a tree - an annoying one to support but one I think is worth utilizing if we consider STAC to be something that can allow referencing of Items in a deeper way then parent->child (e.g., a derived from link).

Is there a way to avoid Link Type and also allow for these types of relative links? Where does the problem? I don't follow where in the normalization process this breaks things.

matthewhanson:
Good point, basically we don't want to convert relative non-hierarchical links to absolute URLs.

If we just leave relative links the way they are, if you normalize_hrefs those relative links will still be valid.

It's a problem if you wanted to move the item from one place in the catalog to another, but that wouldn't work now would it? You would use `remove_item` and `add_item`, but unless you manually update the relative link it won't be correct in it's new location.

Or if you remove the item linked to or move it, but again, that also would cause the same issue currently.

matthewhanson:
Leaving the links alone actually is the same way that relative assets are treated, so I think we should keep them consistent.

lossyrob:
If you are normalizing HREFs, it will resolve links recursively via [resolve_links](https://github.com/stac-utils/pystac/blob/5a492ca37ef6d8b0b2ea10db35150e72a709a398/pystac/stac_object.py#L462), which uses [_object_links](https://github.com/stac-utils/pystac/blob/5a492ca37ef6d8b0b2ea10db35150e72a709a398/pystac/catalog.py#L683), which can be contributed to by extensions, a la the [label extension](https://github.com/stac-utils/pystac/blob/develop/pystac/extensions/label.py#L297) specifying that the "source" link is an object link that should be resolved. So for example you are normalizing a catalog `ROOT` that has two subcatalogs `IMAGES` and `LABELS`, and `LABELS` has a single item `LabelItem` that has the "label" extension, and `IMAGES` has a single item `SourceItem` that is just an image.

One way the traversal could happen is with `LABELS` going first:
- ROOT calls LABELS.normalize_hrefs
- LABELS resolves the child item link calls LabelItem.normalize_hrefs
- This causes the "source" link to get resolved
- ROOT calls IMAGES.normalize_hrefs
- IMAGES resolves the child item link for SourceItem - which was previously resolved and so pulls from the cache. This item and the "source" linked item are the same item. SourceItem.normalize_href is called, which modifies the source item HREF, and that is reflected in both the LabelItem's link and the SourceItem's self href.

Another traversal reads IMAGES first, but similarly resolves SourceItem, and when the "source" link of the LabelItem is read, it resolves to the cached item, and they are still refering to the same object.

It seems like your changes continue to resolve the object links, so I think the above would still apply here? The link for any of the _object_link links should not be updated by the linking object, because those are pending modification (or are already resolved and modified) by the normalize_href call in another part of the catalog.

I might be confusing myself, and others with this, but at least I think the "objects can get resolved via non-hierarchical links" is applicable to your idea of using resovled-ness as a flag determining whether or not to save something

matthewhanson:
I've updated this to handle all links defined by extensions `_object_links`. 
This appears to emulate the previous behavior, except that CatalogType is not required to be specified on save. Instead the root catalog of any STAC object determines relative vs absolute based on it's `catalog_type` field.

lossyrob (approved):
Looks good!

I made a change to a tutorial for adding a custom extension just to add a note that the `_object_links()` for an extension also influence if those links will be relative or absolute based on catalog type.

Merge away!

:tada:

lossyrob:
We should probably cut a new release with this change and the coupe of fixes - the last 1.0.0-beta.2 release. Want to set up a time to pair on that so there's two of us who can release?

matthewhanson:
@lossyrob Let's plan for Monday to release, I've got another PR (much smaller) to add self hrefs to `add_items` and `add_child` #284 and possibly another to optimize generate_subcatalogs

Issue #332, opened 2021-05-05
Stricter type hint linting


l0b0:
Would you be interested in stricter type hint linting? I've implemented these additional mypy settings in [another project](https://github.com/linz/geospatial-data-lake/blob/eeee270c29cc6bd8bfe812576538c3a8e6bbd5c6/setup.cfg) already:

- [check_untyped_defs](https://mypy.readthedocs.io/en/stable/config_file.html#confval-check_untyped_defs) "Type-checks the interior of functions without type annotations."
- [disallow_any_generics](https://mypy.readthedocs.io/en/stable/config_file.html#confval-disallow_any_generics) "Disallows usage of generic types that do not specify explicit type parameters."
- [no_implicit_optional](https://mypy.readthedocs.io/en/stable/config_file.html#confval-no_implicit_optional) "Changes the treatment of arguments with a default value of `None` by not implicitly making their type `Optional`." Basically, declaring a parameter `foo: str = None` becomes an error, rather than being implicitly treated as `Optional[str]`.
- [show_error_codes](https://mypy.readthedocs.io/en/stable/config_file.html#confval-show_error_codes) "Shows error codes in error messages." This makes it easier to write `type: ignore[SPECIFIER]` comments (which are more specific than the big hammer, `type: ignore`), since the `SPECIFIER` part of that is part of the error message.
- [warn_redundant_casts](https://mypy.readthedocs.io/en/stable/config_file.html#confval-warn_redundant_casts) "Warns about casting an expression to its inferred type."
- [warn_return_any](https://mypy.readthedocs.io/en/stable/config_file.html#confval-warn_return_any) "Shows a warning when returning a value with type `Any` from a function declared with a non-`Any` return type."
- [warn_unused_ignores](https://mypy.readthedocs.io/en/stable/config_file.html#confval-warn_unused_ignores) "Warns about unneeded `# type: ignore` comments."

I've also avoided `ignore_missing_imports` simply by marking only a handful of specific imports as ignored.

lossyrob:
@l0b0 in #309 I actually moved to Pyright as the type checker, as mypy was having trouble inferring types and allowing some of the techniques I deployed there. I'd be interested if mypy could work with the current typing approach with stricter configuration options without significant changes to the approach. I forget exactly what mypy was erroring on, but if you run default mypy against #309 you'll see some errors that I believe would take some deeper changes to how we're structuring types, or creating a lot more lines of code, where Pyright is able to type check successfully.

[...]

PR #337, opened 2021-05-09
Fix mypy support

l0b0:
**Related Issue(s):** #332

**Description:** Fixes most typing issues which mypy complain about in current main branch. I'm unsure about the remaining issue:

```console
$ mypy pystac/
pystac/extensions/base.py:29: error: Incompatible default for argument "typ" (default has type "object", argument has type "Type[P]")  [assignment]
pystac/extensions/base.py:30: error: Variable "typ" is not valid as a type  [valid-type]
pystac/extensions/base.py:30: note: See https://mypy.readthedocs.io/en/latest/common_issues.html#variables-vs-type-aliases
Found 2 errors in 1 file (checked 37 source files)
```

Since types aren't enforced at runtime, and I don't believe mypy will assert the type of `result` at the various call sites (verified by changing `return self._get_property(CLOUD_COVER_PROP, float)` to `return self._get_property(CLOUD_COVER_PROP, str)` and seeing no change in the mypy output), the `_get_property` `typ` argument should probably be removed, unless it *is* enforced by Pyright. I'd like to hear your opinion on this before going forward with that change though.

**PR Checklist:**

### DIFF
--- a/pystac/collection.py
+++ b/pystac/collection.py
@@ -431,9 +431,9 @@ class RangeSummary(Generic[T]):
         return {"minimum": self.minimum, "maximum": self.maximum}
 
     @classmethod
-    def from_dict(cls, d: Dict[str, Any], typ: Type[T] = Any) -> "RangeSummary[T]":
-        minimum: Optional[T] = get_required(d.get("minimum"), "RangeSummary", "minimum")
-        maximum: Optional[T] = get_required(d.get("maximum"), "RangeSummary", "maximum")
+    def from_dict(cls, d: Dict[str, Any]) -> "RangeSummary[T]":
+        minimum: T = get_required(d.get("minimum"), "RangeSummary", "minimum")
+        maximum: T = get_required(d.get("maximum"), "RangeSummary", "maximum")
         return cls(minimum=minimum, maximum=maximum)
 
 
@@ -483,7 +483,7 @@ class Summaries:
         self.schemas.pop(prop_key, None)
         self.other.pop(prop_key, None)
 
-    def is_empty(self):
+    def is_empty(self) -> bool:
         return not (
             any(self.lists) or any(self.ranges) or any(self.schemas) or any(self.other)
         )
--- a/pystac/extensions/eo.py
+++ b/pystac/extensions/eo.py
@@ -272,16 +272,16 @@ class EOExtension(
         """
         return self._get_bands()
 
-    def _get_bands(self) -> Optional[List[Band]]:
-        return map_opt(
-            lambda bands: [Band(b) for b in bands],
-            self._get_property(BANDS_PROP, List[Dict[str, Any]]),
-        )
-
     @bands.setter
     def bands(self, v: Optional[List[Band]]) -> None:
         self._set_property(
             BANDS_PROP, map_opt(lambda bands: [b.to_dict() for b in bands], v)
+        )
+
+    def _get_bands(self) -> Optional[List[Band]]:
+        return map_opt(
+            lambda bands: [Band(b) for b in bands],
+            self._get_property(BANDS_PROP, List[Dict[str, Any]]),
         )
 
     @property
@@ -365,6 +365,7 @@ class SummariesEOExtension(SummariesExtension):
         """Get or sets a list of :class:`~pystac.Band` objects that represent
         the available bands.
         """
+
         return map_opt(
             lambda bands: [Band(b) for b in bands],
             self.summaries.get_list(BANDS_PROP, Dict[str, Any]),
--- a/pystac/extensions/file.py
+++ b/pystac/extensions/file.py
@@ -188,6 +188,7 @@ class SummariesFileExtension(SummariesExtension):
         Returns:
             FileDataType
         """
+
         return map_opt(
             lambda x: [FileDataType(t) for t in x],
             self.summaries.get_list(DATA_TYPE_PROP, str),
@@ -261,4 +262,4 @@ class FileExtensionHooks(ExtensionHooks):
                     obj["assets"][key][CHECKSUM_PROP] = old_checksum[key]
 
 
-FILE_EXTENSION_HOOKS = FileExtensionHooks()
+FILE_EXTENSION_HOOKS: ExtensionHooks = FileExtensionHooks()
--- a/pystac/extensions/sar.py
+++ b/pystac/extensions/sar.py
@@ -8,8 +8,7 @@ from typing import Any, Dict, Generic, List, Optional, Set, TypeVar, cast
 
 import pystac
 from pystac.serialization.identify import STACJSONDescription, STACVersionID
-from pystac.extensions.base import ExtensionManagementMixin
-from pystac.extensions.projection import ProjectionExtension
+from pystac.extensions.base import ExtensionManagementMixin, PropertiesExtension
 from pystac.extensions.hooks import ExtensionHooks
 from pystac.utils import get_required, map_opt
 
@@ -59,7 +58,7 @@ class ObservationDirection(enum.Enum):
 
 
 class SarExtension(
-    Generic[T], ProjectionExtension[T], ExtensionManagementMixin[pystac.Item]
+    Generic[T], PropertiesExtension, ExtensionManagementMixin[pystac.Item]
 ):
     """SarItemExt extends Item to add sar properties to a STAC Item.
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -15,6 +15,8 @@
 import os
 import sys
 import subprocess
+from typing import Any, Dict
+
 sys.path.insert(0, os.path.abspath('.'))
 sys.path.insert(0, os.path.abspath('../'))
 from pystac.version import __version__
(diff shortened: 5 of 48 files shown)


lossyrob:
The `typ` argument _is_ recognized by Pyright; passing in the type as an argument forces Pyright (and Pylance) to recognize it as the return type of the method, and can for instance catch errors when the type one passes into _get_property as a declared type is not the same as the return type of the property method that uses it. I'd label this as mildly useful; otherwise you have to assume that the Optional[T] returned from _get_property is Any, and there's no way to input a type restriction on that return type.

The mypy compliance changes are fine, but are mostly working around mypy not being able to infer types that shouldn't necessarily be stated (e.g. all of the ExtensionHooks additions). I was however suprised to see that Pyright missed some lack of return statements in the item_assets extension. I wonder if there's some settings to fiddle around with for Pyright to not let something like that pass.

I don't feel super strongly about the need to pass in types, and am mildly against having to satisfy mypy's lack of ability to infer types or handle passing runtime types; however seeing that pyright missed something pretty obvious it does feel like having the addition of mypy may be worth having to work around these things.

If we want to move forward, then I'll ask that you add mypy to the requirements-dev.txt and to the `scripts/test` script so that it becomes part of the CI testing.

duckontheweb:
Based on the [`mypy` docs for the`has-type` error](https://mypy.readthedocs.io/en/stable/error_code_list.html?highlight=%22has-type%22#check-that-type-of-target-is-known-has-type) it seems like the need for type annotations on things like `DATACUBE_EXTENSION_HOOKS` might be due to an import cycle. I also couldn't find anything in the `pyright` docs about stopping `pyright` from [inferring a `NoReturn` type](https://github.com/microsoft/pyright/blob/main/docs/type-inference.md#noreturn-return-type), so I think if we want to continue to catch those errors we will need to use `mypy`. I agree that it's less than ideal to have to add those annotations where they should be inferred, but given that `mypy` caught some other issues that `pyright` missed, I think the trade-off is worth it.

As for the issue regarding the `typ` argument to `_get_property`, it looks like we can resolve the `mypy` errors and retain the type checking on the return value by making two changes:

1) Remove the default from the `typ` parameter (we never take advantage of this default in the code)
2) Remove the variable type annotation on `result`

```diff
- def _get_property(self, prop_name: str, typ: Type[P] = Type[Any]) -> Optional[P]:
-     result: Optional[typ] = self.properties.get(prop_name)
+ def _get_property(self, prop_name: str, typ: Type[P]) -> Optional[P]:
+     result = self.properties.get(prop_name)
```

I tested this by passing a `str` type instead of `float` to the [`_get_property` call](https://github.com/stac-utils/pystac/blob/5158a4c622db0e58c42faf109dbd35d2ad0c7abf/pystac/extensions/eo.py#L295) in the `eo` extension and `mypy` correctly raised a `[return-value]` error (and all tests continue to pass).
   
I would support @lossyrob 's suggestion that we move forward by adding `mypy` to `requirements-dev.txt` and the test script .

lossyrob:
👍 @l0b0 do you want to add mypy to the CI and move this from RFC to ready to merge? Thanks!

l0b0:
Thank you @duckontheweb, that worked!

l0b0:
@lossyrob I tried using the non-deprecated `importlib` rather than `imp`, but ran into [an issue with the "vanilla" job](https://github.com/stac-utils/pystac/runs/2651643026?check_suite_focus=true). Not sure what to do about that, so I've just ignored the setup.py type issue.

lossyrob:
Not type checking setup.py is totally reasonable to me - I think we'll probably want to change that import in the future anyway.

lossyrob:
There was a conflict with a recent PR around test_pointcloud, resolved and merged that via the GitHub UI

duckontheweb:
The CI failures are due to type errors in the pointcloud tests introduced from the merge. It looks like a couple of the functions just need their return types annotated.

The successful builds on Windows are due to a problem with the CI. The CI script doesn't seem to be running at all, so it isn't returning a non-zero exit code. I'll fix that in a separate PR.

lossyrob:
Yep, the merge conflict didn't account for some of the missing return types; I just added them and pushed that change so this should be ready to go soon 🤞

pystac/collection.py
```diff
@@ -151,16 +151,10 @@ class TemporalExtent:
...
-        self, intervals: Union[List[List[Optional[Datetime]]], List[Optional[Datetime]]]
-    ):
-        # A common mistake is to pass in a single interval instead of a
-        # list of intervals. Account for this by transforming the input
-        # in that case.
-        if isinstance(intervals, list) and isinstance(intervals[0], Datetime):
-            self.intervals = [cast(List[Optional[Datetime]], intervals)]
-        else:
-            self.intervals = cast(List[List[Optional[Datetime]]], intervals)
+    def __init__(self, intervals: List[List[Optional[Datetime]]]):
```
lossyrob, line 154: This changes behavior; I'd prefer to keep this PR to the type changes. If we want to remove this nicety, which was put in place based on user requests, we can discuss in a follow up PR or issue.

pystac/extensions/file.py
```diff
@@ -188,10 +188,11 @@ def data_type(self) -> Optional[List[FileDataType]]:
         Returns:
             FileDataType
         """
-        return map_opt(
-            lambda x: [FileDataType(t) for t in x],
```
lossyrob, line 192: Is it the case that mypy doesn't allow lambdas? Is there a flag for this? I find this style of short lambda useful and more readable

pystac/extensions/sar.py
```diff
@@ -59,7 +58,7 @@ class ObservationDirection(enum.Enum):
 
 
 class SarExtension(
-    Generic[T], ProjectionExtension[T], ExtensionManagementMixin[pystac.Item]
+    Generic[T], PropertiesExtension, ExtensionManagementMixin[pystac.Item]
```
lossyrob, line 61: Nice catch

lossyrob (changes requested):
Looking good. I have one request to avoid changing library behavior, and a question around the possibility maintaining the use of lambda.

pystac/extensions/file.py
l0b0: No, it allows lambdas; it's just simpler to type a function. I'm not sure you can type lambdas, but I expect it'd be a mess.

l0b0:
How's this @lossyrob?

pystac/extensions/file.py
lossyrob: Ah right, I think Pyright is OK with inferring types in lambdas without annotations where mypy isn't...a bit of inconvenience but worth it.

lossyrob (approved):

pystac/extensions/eo.py
```diff
@@ -365,8 +365,12 @@ def bands(self) -> Optional[List[Band]]:
         """Get or sets a list of :class:`~pystac.Band` objects that represent
         the available bands.
         """
+
+        def construct_bands(bands: List[Dict[str, Any]]) -> List[Band]:
+            return [Band(b) for b in bands]
+
         return map_opt(
-            lambda bands: [Band(b) for b in bands],
+            construct_bands,
```
duckontheweb, line 373: It looks like the reason that mypy can infer a type for the lambda in the `_get_bands` method above but not here is due to the way the 2nd argument to `map_opt` is typed.

In this case, we didn't have a type annotation on `SummariesExtension` for the `summaries` property. Adding the following type annotation in `pystac.extensions.base.SummariesExtension` allows us to keep an untyped lambda here:

```python
from pystac.collection import Collection, Summaries

...

class SummariesExtension:
    summaries: Summaries
```

Since it seems like `mypy` is actually catching cases where we haven't adequately typed things, this might be a better solution that creating a typed function.

pystac/extensions/file.py
duckontheweb: I think this may actually be an indicator that we're not adequately typing something in the second argument to `map_opt`. I left a comment on `pystac.extensions.eo` to illustrate this in more detail, but I think we may want to keep the untyped lambdas here and track down the typing issues in more detail.

I think the solution from the other comment will also solve this case and allow us to keep the untyped lambda here.

duckontheweb (changes requested):
@l0b0 Thanks so much for this PR, it's uncovered some subtle typing issues that we never would have caught otherwise.

My only requested changes are related to the change from untyped lambdas to defined functions that @lossyrob commented on, other than that this looks great and thanks for bearing with all of the changes.

.gitignore
```diff
@@ -14,6 +14,7 @@ stdout*
 .idea/
 docs/_build/
 coverage.xml
+venv/
```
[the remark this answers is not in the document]
l0b0: I'll remove it, but I consider using the global .gitignore for anything is a bad idea. You never know when it will conflict with the needs of some project, at which point you need a custom `!foo` line in the project .gitignore or `!/path/to/project/foo` in the global .gitignore.

l0b0:
I think this latest version should cover your comments, @duckontheweb. Thanks for working out how to keep the original lambda code!

duckontheweb (approved):
Looks great, thanks again for contributing this!

PR #347, opened 2021-05-18
Version 1.0.0-beta.1

lossyrob:
This PR updates to STAC 1.0.0-rc.4 and changes the PySTAC library version number to 1.0.0-beta.1

We will be cutting a 1.0.0-beta.1 release of PySTAC so that users can use the current codebase to update their catalogs to STAC 1.0.0-rc.4 in preparation for the upcoming full STAC 1.0 release.


lossyrob:
@duckontheweb this is the version upgrade. Once this is in master I'll:
- Push a new release branch `1.0`. I would normally just make a pull request from `main` to the release branch, but this is the first of the series. This will also avoid the need for a PR review on that which I think would be necessary for future releases into existing release branches
- Create a GitHub release, which will create a v1.0.0-beta.1 tag. CI will then publish the new release to pypi. I have started putting the CHANGELOG text for that release into the GitHub release description, so I'll do that as well.

That should be it for this beta release, and should be the same process for when we want to move to an RC release

CHANGELOG.md
```diff
@@ -287,7 +295,8 @@ use `Band.create`
 
 Initial release.
 
-[Unreleased]: <https://github.com/stac-utils/pystac/compare/v0.5.6...main>
+[Unreleased]: <https://github.com/stac-utils/pystac/compare/v1.0.0-beta.1...main>
+[v1.0.0-beta.1]: <https://github.com/stac-utils/pystac/compare/v0.5.6...v1.0.0-beta.1>
```
duckontheweb, line 299: Looks like these links are inconsistent across versions. Some use `...` (3 dots), which just lists the commits, and some use `..` which shows the file diffs. My preference would be for all of them to use `..` (2 dots) and show the file diffs, but either way we should make them consistent throughout.

```suggestion
[Unreleased]: <https://github.com/stac-utils/pystac/compare/v1.0.0-beta.1..main>
[v1.0.0-beta.1]: <https://github.com/stac-utils/pystac/compare/v0.5.6..v1.0.0-beta.1>
```

Also the `## unreleased` header should probably be changed to `## [Unreleased]` so it matches the link here.

duckontheweb (approved):
There are a couple of small things that should probably be fixed in the CHANGELOG, but other than that it looks good.

CHANGELOG.md
lossyrob: Good call, will make that change. Thanks!

duckontheweb: > Looks like these links are inconsistent across versions. Some use `...` (3 dots), which just lists the commits, and some use `..` which shows the file diffs. My preference would be for all of them to use `..` (2 dots) and show the file diffs, but either way we should make them consistent throughout.
> 
> Also the `## unreleased` header should probably be changed to `## [Unreleased]` so it matches the link here.

Actually, scratch that. Now that I've seen how many example files change between versions I think we should go with `...` and just show the list of commits 😄

duckontheweb: 👍

Issue #328, opened 2021-05-04
Add 'via' and 'canonical' rel types are now options in Items.


lossyrob:
From stac-spec changelog for PR at https://github.com/radiantearth/stac-spec/pull/884

duckontheweb:
Closed via #351

PR #351, opened 2021-05-20
[WIP] Add staticmethod for canonical links

duckontheweb:
**Related Issue(s):** #

#328 

**Description:**

* Adds `Link.canonical` static method for creating links with a "canonical" rel type

    Since the "via" links can refer to arbitrary non-STAC records I didn't add an equivalent static method for those.

**PR Checklist:**

### DIFF
--- a/pystac/link.py
+++ b/pystac/link.py
@@ -14,7 +14,7 @@ HIERARCHICAL_LINKS = ["root", "child", "parent", "collection", "item", "items"]
 
 
 class Link:
-    """A link is connects a :class:`~pystac.STACObject` to another entity.
+    """A link connects a :class:`~pystac.STACObject` to another entity.
 
     The target of a link can be either another STACObject, or
     an HREF. When serialized, links always refer to the HREF of the target.
@@ -325,3 +325,13 @@ class Link:
     def item(item: "Item_Type", title: Optional[str] = None) -> "Link":
         """Creates a link to an Item."""
         return Link("item", item, title=title, media_type="application/json")
+
+    @staticmethod
+    def canonical(
+        item_or_collection: Union["Item_Type", "Collection_Type"],
+        title: Optional[str] = None,
+    ) -> "Link":
+        """Creates a canonical link to an Item or Collection."""
+        return Link(
+            "canonical", item_or_collection, title=title, media_type="application/json"
+        )
--- a/tests/test_link.py
+++ b/tests/test_link.py
@@ -82,6 +82,19 @@ class LinkTest(unittest.TestCase):
 
 
 class StaticLinkTest(unittest.TestCase):
+    def setUp(self):
+        self.item = pystac.Item(
+            id="test-item",
+            geometry=None,
+            bbox=None,
+            datetime=TEST_DATETIME,
+            properties={},
+        )
+
+        self.collection = pystac.Collection(
+            "collection id", "desc", extent=ARBITRARY_EXTENT
+        )
+
     def test_from_dict_round_trip(self):
         test_cases = [
             {"rel": "", "href": ""},  # Not valid, but works.
@@ -101,13 +114,21 @@ class StaticLinkTest(unittest.TestCase):
                 pystac.Link.from_dict(d)
 
     def test_collection(self):
-        c = pystac.Collection("collection id", "desc", extent=ARBITRARY_EXTENT)
-        link = pystac.Link.collection(c)
+        link = pystac.Link.collection(self.collection)
         expected = {"rel": "collection", "href": None, "type": "application/json"}
         self.assertEqual(expected, link.to_dict())
 
     def test_child(self):
-        c = pystac.Collection("collection id", "desc", extent=ARBITRARY_EXTENT)
-        link = pystac.Link.child(c)
+        link = pystac.Link.child(self.collection)
         expected = {"rel": "child", "href": None, "type": "application/json"}
+        self.assertEqual(expected, link.to_dict())
+
+    def test_canonical_item(self):
+        link = pystac.Link.canonical(self.item)
+        expected = {"rel": "canonical", "href": None, "type": "application/json"}
+        self.assertEqual(expected, link.to_dict())
+
+    def test_canonical_collection(self):
+        link = pystac.Link.canonical(self.collection)
+        expected = {"rel": "canonical", "href": None, "type": "application/json"}
         self.assertEqual(expected, link.to_dict())


duckontheweb:
@lossyrob Aside from the static methods, did you have any other changes in mind to support these rel types?

lossyrob:
Hm, I actually misremembered and thought we had a Enum similar to MediaType that encapsulated common rel types. Now I see that we only have some of them included as static methods.

Perhaps we should a RelType enum? We'd could (similar to MediaType) change the signatures for e.g. the Link __init__ to take a `Union[str, RelType]`, as we want to indicate that a RelType would be prefered but not block other strings to be passed in.

The rel types would include any that are on the item, catalog or collection spec pages, as well as the [best practices](https://github.com/radiantearth/stac-spec/blob/master/best-practices.md#using-relation-types)

Would this be worth it to add? I think it might help users decide what rel types to use for their links. If not then I'd say this PR is good to go

duckontheweb:
Yeah, when I first read the issue I assumed there was an Enum as well. I'm +1 for adding a RelType enum and changing any callables that have a `rel` argument to accept `Union[str, RelType]`. This would also allow us to use checks like `link.rel == RelType.ROOT` instead of `link.rel == "root"`, which is probably better practice.

I can add this to the PR and allow others to comment on that change in the review process.

pystac/extensions/version.py
```diff
@@ -110,41 +102,49 @@ def deprecated(self, v: Optional[bool]) -> None:
...
         )
 
     @predecessor.setter
     def predecessor(self, item: Optional[T]) -> None:
-        self.obj.clear_links(PREDECESSOR)
+        self.obj.clear_links(pystac.RelType.PREDECESSOR)
         if item is not None:
-            self.obj.add_link(pystac.Link(PREDECESSOR, item, MEDIA_TYPE))
+            self.obj.add_link(
+                pystac.Link(pystac.RelType.PREDECESSOR, item, pystac.MediaType.JSON)
```
lossyrob, line 130: Good catch on MEDIA_TYPE -> pystac.MediaType.JSON

pystac/rel_type.py
```diff
@@ -0,0 +1,33 @@
+from enum import Enum
+
+
+class RelType(str, Enum):
+    """A list of common rel types that can be used in STAC Link metadata."""
```
lossyrob, line 5: Perhaps we should link to the best practices docs here so users who want to learn more about how to use the reltypes have an easy find?

pystac/rel_type.py
```diff
@@ -0,0 +1,33 @@
...
+    SELF = "self"
+    VIA = "via"
+
+    # Label Extension
+    SOURCE = "source"
+
+    # Version Extension
+    LATEST = "latest-version"
+    PREDECESSOR = "predecessor-version"
+    SUCCESSOR = "successor-version"
```
lossyrob, line 33: Including extension reltypes makes sense to me, though I could see a situation where someone is confused about using a reltype outside of the extension context or expecting to find the reltype in the extension. I'd say I'm +0 on this approach.

pystac/extensions/version.py
```diff
@@ -22,14 +22,6 @@
 VERSION: str = "version"
 DEPRECATED: str = "deprecated"
 
-# Link "rel" attribute values.
-LATEST: str = "latest-version"
-PREDECESSOR: str = "predecessor-version"
-SUCCESSOR: str = "successor-version"
```
lossyrob, line 28: Maybe it makes sense to add extension-specific relation types to a comment or docstring somewhere, to make it easier for users looking at the version source for the first time to see what RelTypes this extension adds? The constants-on-top approach the extensions have been nice for serving as a documentation of the spec - you see it adds these reltypes and you end up groking a bit more about the extension spec. Moving the extension rel types to RelType makes sense to me, as it keeps it all accessible in one place, and some source file docs would balance out the only negative effect I see of that change.

lossyrob (approved):
Looks good, added some things to consider but nothing that I think is a blocker.

pystac/rel_type.py
duckontheweb: Yeah, that's a good idea. I'll add a reference to those docs.

pystac/rel_type.py
duckontheweb: To be honest, the more I sit with this approach the less I like it. I think it might be clearer to have rel types defined by extensions sitting with that extension code. 

Here's an alternative proposal:

1) Create a base `RelType` enum:

    ```python
    class RelType(str, Enum):
        def __str__(self) -> str:
            return str(self.value)
    ``` 

2) Have a `CoreRelType` enum that actually defines the values from the core spec and best practices:

    ```python
    class CoreRelType(RelType):
        ALTERNATE = "alternate"
        ...
    ```

3) Extension define their own sub-class as needed:

    ```python
    class LabelExtensionRelType(RelType):
        SOURCE = "source"
    ```

The type annotations for all the functions/classes that use rel types would stay as `rel: Union[str, RelType]` and extended rel types would be more clearly associated with the right extension. The downside would be that you wouldn't have a single place to enumerate all STAC rel types, you would need to specifically import something like `pystac.extensions.label.LabelExtensionRelType`, but maybe that explicit import is better anyway...

lossyrob: Yeah that makes sense to me. A little more work to use but a clearer organization.

I do think that having `RelType` hold all of the core rel types instead of making a separate `CoreRelType` would make sense; that would make LabelExtensionRelType also hold those core reltypes which is weird but it's probably more convenient to have the base RelType have the core instead of having to figure out there's a "core". Not sure though.

`LabelExtension` is getting a bit long, maybe `LabelRelType` or `LabelExtRelType`? Or perhaps it's best to have the full name for clarity, I usually don't mind additional typing if it makes code more consistent and clear. I do sometimes get accused of writing enterprise-y looking code though with long names 😆

duckontheweb: Python [only allows you to subclass an enum with no members](https://docs.python.org/3/library/enum.html#restricted-enum-subclassing), so we wouldn't be able to have `LabelExtRelType` inherit from `RelType` if it included the core members.

It looks like we don't need to, though. Since it inherits from `str`, `LabelExtRelType` will pass a type check against `str` in any function arguments, so it doesn't need to inherit from `RelType`. As you mentioned in an earlier comment, we're really just using the `Union[str, RelType]` annotation to let users know they should use `RelType` if possible. 

Given all of that, I would be for removing the extension-specific rel types from `pystac.rel_type.RelType` and putting them into their own enums in the extension modules.

duckontheweb: I implemented this in 2d53f0b1b4f5bca32bc3b3e96facb38f8a66c706 so we can see how it feels. I'll revert it if we decide to go a different route.

duckontheweb:
@lossyrob Based on the discussion above I moved any rel types defined by extensions into the extension code. If you're okay with that approach I think we're ready to merge.

lossyrob:
👍

Issue #353, opened 2021-05-21
Optional EO bands in apply method


duckontheweb:
The `bands` property is optional in the EO Extension spec and in the [`bands` setter](https://github.com/stac-utils/pystac/blob/61f62b1559ef3864a934982600cbc2e4770f136b/pystac/extensions/eo.py#L282) in the `EOExtension` class, but is required in [`EOExtension.apply`](https://github.com/stac-utils/pystac/blob/61f62b1559ef3864a934982600cbc2e4770f136b/pystac/extensions/eo.py#L255). It seems like this should be optional as well.

duckontheweb:
Closed by 7d341b21d1011dd160b5c1eaf46b1260c1272807

Issue #326, opened 2021-05-03
Update extension pydoc strings to match new extension architecture


lossyrob:
Extensions were refactored as part of #309; however, ensuring docstrings were up to date and matching the new architecture was left as a TODO.

This issue covers each extension besides item-assets and datacube, which were implemented as part of that PR and are tracked in #325 and #324 .

One approach to this work is to take on individual extensions, and work through the docstrings and test to make sure they are completely up to date, as well as check the sphinx docs to make sure they are correct.

This should take into the account the work to remove type info from docstrings encapsulated in #323 

Extensions covered by this issue:
- [x] eo (#355)
- [x] file (#442)
- [x] label (#442)
- [x] pointcloud (#558)
- [x] projection (#372)
- [x] sar (#556)
- [x] sat (#509)
- [x] scientific (#372)
- [x] timestamps (#513)
- [x] version (#442)
- [x] view (#372)

lossyrob:
Also we may want to consider doing the work for #327 as part of focusing on each extension.

duckontheweb:
Closed by all PRs listed in description.

PR #355, opened 2021-05-21
Update EO extension docs

duckontheweb:
**Related Issue(s):** #

#326 
#353 

**Description:**

* Adds docstrings for the extension base classes
* Updates docstrings for the new EO extension classes
* Updates the "Extensions" section of the "Concepts" docs to reflect the new extension structure
* Other fixes to the docs to make the above changes work

I plan on submitting 1-2 more PRs to address the remaining extension documentation work for #326, but I wanted to get feedback on this first since I'll likely use it as a template for the other extensions.

**PR Checklist:**

### DIFF
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -309,87 +309,85 @@ The :class:`~pystac.validation.JsonSchemaSTACValidator` takes a :class:`~pystac.
 Extensions
 ==========
 
-Accessing Extension functionality
+From the documentation on `STAC Spec Extensions 
+<https://github.com/radiantearth/stac-spec/tree/master/extensions>`__:
+
+   Extensions to the core STAC specification provide additional JSON fields that can be
+   used to better describe the data. Most tend to be about describing a particular 
+   domain or type of data, but some imply functionality.
+
+This library makes an effort to support all extensions that are part of the
+`stac-extensions GitHub org
+<https://stac-extensions.github.io/#extensions-in-stac-extensions-organization>`__, and
+we are committed to supporting all STAC Extensions at the "Candidate" maturity level or
+above (see the `Extension Maturity
+<https://stac-extensions.github.io/#extension-maturity>`__ documentation for details).
+
+Accessing Extension Functionality
 ---------------------------------
 
-All STAC objects are accessed through ``Catalog``, ``Collection`` and ``Item``, and all extension functionality
-is accessed through the ``ext`` property on those objects. For instance, to access the band information
-from the ``eo`` extension for an item that implements the extension, you use:
+Extension functionality is encapsulated in classes that are specific to the STAC
+Extension (e.g. Electro-Optical, Projection, etc.) and STAC Object
+(:class:`~pystac.Collection`, :class:`pystac.Item`, or :class:`pystac.Asset`). All
+classes that extend these objects inherit from
+:class:`pystac.extensions.base.PropertiesExtension`, and you can use the
+``ext`` method on these classes to extend an object.
+
+For instance, to extend an item with the :stac-ext:`Electro-Optical Extension <eo>`
+you would use :meth:`EOExtension.ext <pystac.extensions.eo.EOExtension.ext>` as
+follows:
 
 .. code-block:: python
 
-   # All of the below are equivalent:
-   item.ext['eo'].bands
-   item.ext[pystac.Extensions.EO].bands
-   item.ext.eo.bands
+   import pystac
+   from pystac.extensions.eo import EOExtension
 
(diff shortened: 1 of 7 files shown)


docs/concepts.rst
```diff
@@ -309,87 +309,85 @@ The :class:`~pystac.validation.JsonSchemaSTACValidator` takes a :class:`~pystac.
...
-required as part of the extension specification and have `None` default values for optional parameters:
-
-.. code-block:: python
+   eo_ext.bands
+   eo_ext.cloud_cover
 
-   eo_ext = item.ext.eo
-   eo_ext.apply(0.5, bands, cloud_cover=None) # Do not have to specify cloud_cover
+See the documentation for each extension implementation for details on the supported
+properties and other functionality.
```
lossyrob, line 356: Another topic worth discussing in this section or elsewhere is the mutability of the underlying STACObject when using extensions. E.g. in the above example, if I set `eo_ext.cloud_cover = 0.1`, then the properties of the wrapped Item is modified. This includes extensions objects that are generated from the extensions; if I were to say do

```
eo_ext.bands[0].common_name = "purple"
```

then the underlying Item or Collection properties would be modified to change the common name of the first band. Users should be aware and careful about the highly mutable nature of PySTAC in general and extensions in particular

pystac/extensions/base.py
```diff
@@ -23,8 +33,21 @@ def _set_summary(
...
+    """Abstract base class for extending the properties of an :class:`~pystac.Item`
+    to include properties defined by a STAC Extension.
+
+    This class should not be instantiated directly. Instead, create an
+    extension-specific class that inherits from this class and instantiate that. See
+    :class:`~pystac.extensions.eo.PropertiesEOExtension` for an example.
+    """
+
     properties: Dict[str, Any]
+    """Additional properties associated with this extension."""
```
lossyrob, line 45:
```suggestion
    """The properties that this extension wraps. 
    
    The extension which implements PropertiesExtension can use _get_property and _set_property to get and set values on this instance. Note that _set_properties mutate the properties directly."""
```

pystac/extensions/base.py
```diff
@@ -23,8 +33,21 @@ def _set_summary(
...
+
     additional_read_properties: Optional[Iterable[Dict[str, Any]]] = None
+    """Additional read-only properties accessible from the extended object. These are
+    used when extending a :class:`~pystac.Asset` to give access to the properties of
+    the owning :class:`~pystac.Item`."""
```
lossyrob, line 50:
```suggestion
    """Additional read-only properties accessible from the extended object. 
    
    These are used when extending a :class:`~pystac.Asset` to give access to the properties of
    the owning :class:`~pystac.Item`. If a property exists in both additional_read_properties and
    properties, the value in additional_read_properties will take precedence. 
    """
```

pystac/extensions/base.py
```diff
@@ -50,27 +73,44 @@ def _set_property(
 
 
 class ExtensionManagementMixin(Generic[S], ABC):
+    """Abstract base class with tools for adding and removing extensions from STAC
```
lossyrob, line 76:
```suggestion
    """Abstract base class with methods for adding and removing extensions from STAC
```

docs/concepts.rst
```diff
@@ -309,87 +309,85 @@ The :class:`~pystac.validation.JsonSchemaSTACValidator` takes a :class:`~pystac.
...
-   item.ext.eo.bands
+This library makes an effort to support all extensions that are part of the
+`stac-extensions GitHub org
+<https://stac-extensions.github.io/#extensions-in-stac-extensions-organization>`__, and
+we are committed to supporting all STAC Extensions at the "Candidate" maturity level or
+above (see the `Extension Maturity
+<https://stac-extensions.github.io/#extension-maturity>`__ documentation for details).
 
-Notice the ``eo`` property on ``ext`` - this utilizes the `__getattr__ <https://docs.python.org/3/reference/datamodel.html#object.__getattr__>`_ method to delegate the property name to the ``__getitem__`` method, so we can access any registered extension as if it were a property on ``ext``.
+Accessing Extension Functionality
```
lossyrob, line 326: Should include a section on adding extensions to STACObjects e.g. `EOExtensions.add_to(item)`

lossyrob (changes requested):
Looking great. I think this approach to extensions works generally. Added some suggested changes, and a request for a section around enabling/disabling extensions on stac objects

docs/concepts.rst
duckontheweb: Good call. I added a note to this section as well as an example of that mutability.

docs/concepts.rst
duckontheweb: Good point. Added this in 6d297b346e42086db00d1a7e30564dc3611fbca7

duckontheweb:
@lossyrob I think I got all of the requested changes covered, let me know what you think.

PR #342, opened 2021-05-17
Removed type information from docstrings

volaya:
**Related Issue(s):** #

fixes #323

**Description:**

**PR Checklist:**

- [ ] Code is formatted (run `scripts/format`)
- [ ] Tests pass (run `scripts/test`)
- [ ] This PR maintains or improves overall codebase code coverage.
- [ ] Changes are added to the [CHANGELOG](https://github.com/stac-utils/pystac/blob/develop/CHANGELOG.md). See [the docs](https://pystac.readthedocs.io/en/latest/contributing.html#changelog) for information about adding to the changelog.

PR #357, opened 2021-05-24
v1.0.0 beta.2 main -> release branch

duckontheweb:
Merges changes from main into the 1.0 release branch for a v1.0.0-beta.2 release

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -3,17 +3,19 @@ name: CI
 on:
   push:
     branches:
-      - develop
-      - master
+      - main
   pull_request:
 
 jobs:
   build:
     name: build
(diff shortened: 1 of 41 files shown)


lossyrob (approved):
Should https://github.com/stac-utils/pystac/pull/342 have a changelog entry? It's a docs-only change, but could be a useful note for anyone who's looking to see why the docs shifted. I'm +0 on that - otherwise gtg

duckontheweb:
> Should #342 have a changelog entry? It's a docs-only change, but could be a useful note for anyone who's looking to see why the docs shifted. I'm +0 on that - otherwise gtg

I'm +1. It might not be obvious to folks who are already familiar with the docs where the type information went. I'll add an entry.

gadomski (approved):

PR #361, opened 2021-05-25
Fix Windows CI builds

duckontheweb:
**Related Issues:**

* #362 

**Description:**

* Fix Windows CI jobs
    Uses `shell: bash` to ensure we're running a `bash` shell, even on Windows.
* Fix dependency caching in CI jobs
    The cache keys are now specific to Python version and OS and point to the correct cache directories for each OS.
* Create temp directories in current working directory
    In the GitHub Actions Windows runner the default `TMPDIR` directory is on a different drive (`C:\\`) than the code and test data files (`D:\\`). This was causing failures in `os.path.relpath` on Windows.
* Separates lint and test jobs
    The lint and type checking jobs only need to run once per Python version since they are not platform-specific.
* Use safe joins and parsing for all paths
    In `pystac.utils` renames `_urlparse` -> `safe_urlparse` and `_join` -> `join_path_or_url` and uses these functions throughout the codebase and tests to ensure we are joining paths properly. There were a number of places where `os.path.join` was being used to join URL paths, which caused failures on Windows. Also changes the signature of `join_path_or_url` to take a `JoinType` enum as the first argument to make the usage more clear.
* Preserves the case of the drive in Windows paths. 
    When using functions like `pystac.utils.make_absolute_href`, the drive letter was being converted to lowercase. The `safe_urlparse` function now preserves the original casing from the start path (see the changes to the tests in 8e9455 for an example)
* Uses explicit `utf-8` encoding for all file I/O operations
    The default encoding on Windows is not necessarily UTF-8, which was causing decoding errors when trying to read test data files. This uses `encoding="utf-8"` for all file read/write operations to avoid conflicts
* Creates separate functions for handling URL v. path cases in `make_absolute_href` and `make_relative_href`
    When debugging the Windows issues, it was pretty difficult to reason through how a URL would be handled in those functions vs. a local path. This PR creates distinct internal functions to handle each case. The result is some redundancy in the code, but I think the trade-off in clarity is worth it.

**UPDATE 2021-06-03**
* Adds `setup.py` to linting and auto-formatting
* Moves version config from `setup.py` to `setup.cfg`. This means we no longer have to use the deprecated `imp` module.

**PR Checklist:**

- [x] Tests pass (run `scripts/test`) (*see comments above*)


duckontheweb:
> There are still a number of path-related failures on Windows. I opened #362 to track fixes to those rather than try to fix them as part of this PR.

Actually, I think I have some fixes to most of these. I'll update the PR shortly.

duckontheweb:
@volaya I would be interested in your review of this since I think it addresses some of the issue you were finding in #271 .

@l0b0 I would also be interested in your review since you've been thinking about pre-commit hooks and CI in #366 .

.github/workflows/continuous-integration.yml
```diff
@@ -24,15 +31,38 @@ jobs:
...
+      - name: Install dependencies
+        run: |
+          pip install -r requirements-test.txt
+          pip install -e ".[validation]"
 
-      - name: Execute linters and test suites
-        run: ./scripts/cibuild
+      - name: Execute test suite
+        run: ./scripts/test
+        shell: bash
```
l0b0, line 65: The script already has a Bash shebang line, so this should be unnecessary.

scripts/lint
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+set -e
+
+echo
+echo " -- CHECKING TYPES WITH PYRIGHT --"
+echo
+
+scripts/pyright pystac tests
```
l0b0, line 15: I'd recommend running this and other lints on setup.py as well. Another approach would be to gather all Python files which are version controlled (`git ls-files '*.py'`) and list each of them, like this:

```
mapfile -d '' python_files < <(git ls-files -z '*.py)
scripts/pyright "${python_files[@]}"
```

scripts/lint
```diff
@@ -0,0 +1,36 @@
...
+black --check pystac
+black --check tests
+
+echo
+echo " -- CHECKING SPELLING WITH CODESPELL --"
+echo
+
+# Code spelling
+codespell -I .codespellignore -f \
+    pystac/*.py pystac/**/*.py \
```
l0b0, line 36: Does `codespell` do its own glob expansion? If not you probably want to run `shopt -s globstar` before this line or use the `git ls-files` approach outlined above.

.github/workflows/continuous-integration.yml
```diff
@@ -42,14 +72,51 @@ jobs:
...
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          # Cache based on OS, Python version, and dependency hash
+          key: lint-${{ runner.os }}-python${{ matrix.python-version }}-${{ hashFiles('requirements-test.txt') }}
+      
+      - name: Install dependencies
+        run: |
+          pip install -r requirements-test.txt
+          pip install -e .
```
l0b0, line 104: Do you need to install the current project to lint the code?

.github/workflows/continuous-integration.yml
```diff
@@ -24,15 +31,38 @@ jobs:
...
 
-      - name: Cache dependencies
+      - name: Cache dependencies (Linux)
+        if: startsWith(runner.os, 'Linux')
         uses: actions/cache@v2
         with:
           path: ~/.cache/pip
           # Cache based on OS, Python version, and dependency hash
-          key: pip-${{ runner.os }}-python${{ matrix.python-version }}-${{ hashFiles('requirements-dev.txt') }}
+          key: test-${{ runner.os }}-python${{ matrix.python-version }}-${{ hashFiles('requirements-test.txt') }}
```
l0b0, line 40: I personally would prefer the original "pip" prefix, since you might need more than one type of cache later ([example](https://github.com/linz/geostore/blob/83a29363f1a6bf74f2dddfbd19e4d68b2c4d36b0/.github/workflows/test.yml#L69-L97)), for example in a multi-stage transition to another Python package manager, and then it's good to be specific.

.github/workflows/continuous-integration.yml
duckontheweb: That's what I thought, too, but leaving this out results in the script not being run at all and that job appearing to succeed. Looking at the logs for the Windows runners, it looks like they execute the script using the following command, which means the shebang doesn't get used:

```
C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
```

.github/workflows/continuous-integration.yml
duckontheweb: That's a good point, I'll update this.

duckontheweb: Fixed in a858c4cf2deeffa8eba2d5540d6b7864e26da996

.github/workflows/continuous-integration.yml
duckontheweb: Fixed in a858c4cf2deeffa8eba2d5540d6b7864e26da996

scripts/lint
duckontheweb: > I'd recommend running this and the other linters on setup.py as well.

Are you suggesting making the linting part of the install process for this package or that we allow people to run the tests using `setup.py`? 

If the first option, it seems like that would required adding the linting requirements (`mypy`, `flake8`, etc.) to the build dependencies and would make the install process take longer. I'm not sure I see the advantage given that we are running this as part of the CI for every PR and every time we publish to PyPi. 

If the second, the `test` command [was deprecated](https://github.com/pypa/setuptools/issues/1684) from setuptools, so I don't think we want to try to support that, unless there is another supported way of running those from setup.py.

scripts/lint
duckontheweb: I actually just copied this over from the `scripts/test` file, so I'm not totally sure. @lossyrob Do you have any insight on this?

scripts/lint
l0b0: Neither, I was suggesting that setup.py should also be linted. :)

duckontheweb: Hah, yes, good suggestion!

duckontheweb: > Another approach would be to gather all Python files which are version controlled (`git ls-files '*.py'`) and list each of them

Okay, this makes more sense to me now in the context of the other part of the comment. I'll experiment with this and see how it works.

l0b0: [Related](https://github.com/stac-utils/pystac/issues/390)

scripts/lint
l0b0: [This commit](https://github.com/stac-utils/pystac/pull/401/commits/97bed90378e32ebf08e3a0c80f004cf20d1d8d09) should solve this issue.

scripts/lint
duckontheweb: Since this is going to change based on #401 anyway, I just added `setup.py` to the list of files being linted (and to the autoformatting in `scripts/format`).

scripts/lint
duckontheweb: Sounds good. I'm going to keep this as-is in this PR and leave it to #401 to introduce the fix.

pystac/utils.py
```diff
@@ -31,18 +37,89 @@ def _urlparse(href: str) -> URLParseResult:
...
+
+    # Get the relative path
+    rel_url = posixpath.relpath(source_path, start_dir)
+
+    # Ensure we retain a trailing slash from the original source path
+    if parsed_source.path.endswith("/"):
+        rel_url += "/"
+
+    # Prepend the "./"
+    if not rel_url.startswith("."):
```
l0b0, line 92: I believe this would break if any of the filenames start with a dot:

```
>>> posixpath.relpath("/foo/.bar", "/foo")
'.bar'
```

Could you do `if rel_url != "./" and not rel_url.startswith("../"):`? I think that should handle the two relevant cases, by making sure the string match includes a slash (which can't be part of a filename).

pystac/utils.py
```diff
@@ -11,7 +12,7 @@
 _pathlib = os.path
```
l0b0, line 12: This threw me for a loop - I thought it was actual [pathlib](https://docs.python.org/3/library/pathlib.html) for a while.

pystac/utils.py
```diff
@@ -61,27 +138,78 @@ def make_relative_href(
...
+
+def _make_absolute_href_url(
+    parsed_source: URLParseResult,
+    parsed_start: URLParseResult,
+    start_is_dir: bool = False,
+) -> str:
+
+    # If the source is already absolute, just return it
+    if parsed_source.scheme != "":
+        return urlunparse((parsed_source))
```
l0b0, line 163: There's an extra pair of brackets here. Ditto in at least one other place below.

pystac/utils.py
```diff
@@ -61,27 +138,78 @@ def make_relative_href(
...
+
+    # If the start path is not a directory, get the parent directory
+    start_dir = (
+        parsed_start.path if start_is_dir else parsed_start.path.rsplit("/", 1)[0]
+    )
+    # Ensure the directory has a trailing slash so urljoin works properly
+    start_dir = start_dir.rstrip("/") + "/"
```
l0b0, line 170: Are you sure you want to do this? It works with \*nix filenames, but in HTTP URLs `foo//bar` and `foo/bar` are semantically distinct. For example, some websites use path separators as some kind of argument separator, as in `https://example.org/api/endpoint/arg1/arg2/…`. It's a terrible practice, but it does happen.

If you want to keep this as is, you might want to `rstrip("/\\")` since this happens before `replace("\\", "/")`. Otherwise `\foo\bar\\` would (I think) end up as `/foo/bar//`, exactly what this is trying to prevent.

setup.cfg
```diff
@@ -0,0 +1,2 @@
+[metadata]
+version = attr: pystac.version.__version__
```
l0b0, line 2: Would you be OK to add a trailing newline to files without them? Git and all the \*nix tools expect it.

requirements-dev.txt
```diff
@@ -1,20 +1,10 @@
...
 nbsphinx==0.7.1
-coverage==5.2.*
-doc8==0.8.1
 
 # https://github.com/pydata/xarray/issues/5299#issuecomment-840730954
 jinja2<3.0
 
-# optional dependencies
-orjson==3.5.2
+-r ./requirements-test.txt
```
duckontheweb, line 10: @l0b0 Just a note that this will conflict with #407

l0b0: Thanks for the heads up; I'll do the rebase when the time comes.

pystac/utils.py
lossyrob: There's an issue around removing this with a suggestion to use mocks for testing. The suggestion would require more test code to implement the patches, but may point to a better way of doing this.

pystac/utils.py
```diff
@@ -31,18 +37,89 @@ def _urlparse(href: str) -> URLParseResult:
...
+    def from_parsed_uri(parsed_uri: URLParseResult) -> "JoinType":
+        """Determines the appropriate join type based on the scheme of the parsed
+        result."""
+        if parsed_uri.scheme == "":
+            return JoinType.PATH
+        else:
+            return JoinType.URL
+
+    PATH = "path"
+    URL = "url"
```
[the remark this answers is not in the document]
lossyrob: Seems like more code than it's worth? Unless we make a base class for all the PySTAC enums that can take advantage of this...to me using straight strings is more clear. If I had to read the source code from scratch I'd spend a few less seconds scratching my head with the explicit strings defined.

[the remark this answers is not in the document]
duckontheweb: Yeah, I agree that it doesn't seem worth it unless we do this throughout the code base. Might be worth a separate PR to see how it feels in practice.

pystac/utils.py
```diff
@@ -31,18 +37,89 @@ def _urlparse(href: str) -> URLParseResult:
...
 
+def _make_relative_href_url(
+    parsed_source: URLParseResult,
+    parsed_start: URLParseResult,
+    start_is_dir: bool = False,
+) -> str:
+
+    # If the start path is not a directory, get the parent directory
+    start_dir = (
+        parsed_start.path if start_is_dir else parsed_start.path.rsplit("/", 1)[0]
```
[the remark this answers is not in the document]
duckontheweb: I thought using `dirname` was causing problems on Windows, but I may have lumped it in with some other issues. I'll try this out and see if the Windows tests continue to pass.

pystac/utils.py
duckontheweb: Thanks for catching this. I'll add a test case for this as well.

lossyrob: Tested this case in https://github.com/duckontheweb/pystac/pull/1, which does not fail. Perhaps because of the above logic on the parsed startdir this isn't an issue?

lossyrob (approved):
LGTM. Added a couple of test cases I was curios about in https://github.com/duckontheweb/pystac/pull/1, these passed.

pystac/utils.py
duckontheweb: We need to do this in the case where we've used `.rsplit("/", 1)[0]` (or `dirname` if that ends up working) for `urljoin` to treat it as a directory, but you're right that we shouldn't do it if we are just using `parsed.start.path`.

duckontheweb: So, it looks like `urljoin` doesn't handle the case that you brought up with empty path arguments in the way we expect:

```python
>>> urljoin("/a/b/x//", "thing")
'/a/b/x/thing'
>>> urljoin("/a/b/x///", "thing")
'/a/b/x/thing'
>>> urljoin("https://a/b/x///", "thing")
'https://a/b/x/thing'
```

Given that this is a pretty standard way of joining URLs in Python I'm inclined to not to try to get around it, but I think I will change the code to the following so that we're at least confident that we are preserving the path properly:

```python
# If the start path is not a directory, get the parent directory
if start_is_dir:
    start_dir = parsed_start.path
else:
    # Ensure the directory has a trailing slash so urljoin works properly
    start_dir = parsed_start.path.rsplit("/", 1)[0] + "/"
```

pystac/utils.py
duckontheweb: Added a test case and updated the logic in cff9606a056ab9107a71da9cc7128a5b13a8fcd1

pystac/utils.py
duckontheweb: Fixed in c48c15bf8537eb81dbf28003eb225806bf55d3ff

setup.cfg
duckontheweb: Fixed in c48c15bf8537eb81dbf28003eb225806bf55d3ff

pystac/utils.py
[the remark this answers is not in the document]
duckontheweb: Works just fine! Fixed in 3467b4c97e8471a5e3e7f832b6e78ab04c5cf026

tests/data-files/get_examples.py
```diff
@@ -50,7 +50,7 @@ def remove_bad_collection(js: Dict[str, Any]) -> Dict[str, Any]:
 
     examples_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), "examples"))
 
-    with TemporaryDirectory() as tmp_dir:
+    with TemporaryDirectory(dir=os.getcwd()) as tmp_dir:
```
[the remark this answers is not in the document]
duckontheweb: I added this in 1ed631f50e0fec8475fc5d0fee31fc6e97f927ef, but ran into some typing issues that I couldn't make sense of, so I just added an ignore statement.

tests/utils/__init__.py
```diff
@@ -41,3 +43,18 @@ def _parse_times(a_dict: Dict[str, Any]) -> None:
...
+# raises a runtime exception:
+#  TypeError: 'type' object is not subscriptable
+class TemporaryDirectory(tempfile.TemporaryDirectory):  # type: ignore [type-arg]
+    def __init__(self, suffix: Optional[str] = None, prefix: Optional[str] = None):
+        """In the GitHub Actions Windows runner the default TMPDIR directory is on a
+        different drive (C:\\) than the code and test data files (D:\\). This was causing
+        failures in os.path.relpath on Windows, so we put the temp directories in the
+        current working directory instead. There os a "tmp*" line in the .gitignore file
+        that ignores these directories."""
+        super().__init__(suffix=suffix, prefix=prefix, dir=os.getcwd())
```
duckontheweb, line 60: @l0b0 @lossyrob I couldn't find a way around ignoring this issue (see the comment for the error I was running into). If anyone has an idea for fixing this the right way I'm happy to include it.

duckontheweb:
@l0b0 I think I've addressed all of your comments. Let me know what you think.

tests/utils/__init__.py
lossyrob: You could define a method that returns a `TemporaryDirectory` instance e.g.

```python
import os
from tempfile import TemporaryDirectory

def get_tmp_dir() -> TemporaryDirectory:
  return TemporaryDirectory(dir=os.getcwd())

with get_tmp_dir() as tmp_dir:
  print(tmp_dir)
```

duckontheweb: Yeah, I tried that first, but it ran into the same typing issue.

lossyrob: Huh, very strange. I would say using the method for simplicity and adding `# type:ignore` with a note about why is the best workaround.

l0b0:
> @l0b0 I think I've addressed all of your comments. Let me know what you think.

LGTM, cheers!

l0b0 (approved):

tests/utils/__init__.py
duckontheweb: Nvm, I was able to fix the typing issue with help from [this comment](https://github.com/python/mypy/issues/5264#issuecomment-399407428). Apparently it's a known issue with Python standard library classes.

duckontheweb: Replaced the sub-class with a function and fixed the type issue in d23be0a2dc2459d308eb332a2fda849936f2cfce

PR #364, opened 2021-05-26
Raster extension

emmanuelmathot:
**Description:**

Implements the raster extension

**PR Checklist:**

- [X] Code is formatted (run `scripts/format`)
- [X] Changes are added to the [CHANGELOG](https://github.com/stac-utils/pystac/blob/develop/CHANGELOG.md). See [the docs](https://pystac.readthedocs.io/en/latest/contributing.html#changelog) for information about adding to the changelog.


pystac/extensions/raster.py
```diff
@@ -0,0 +1,715 @@
...
+    @nodata.setter
+    def nodata(self, v: Optional[float]) -> None:
+        if v is not None:
+            self.properties["nodata"] = v
+        else:
+            self.properties.pop("nodata", None)
+
+    @property
+    def sampling(self) -> Optional[Sampling]:
+        """Get or sets the Indicates whether a pixel value should be assumed
```
duckontheweb, line 470:
```suggestion
        """Get or sets the property indicating whether a pixel value should be assumed
```

pystac/extensions/raster.py
```diff
@@ -0,0 +1,715 @@
...
+    @sampling.setter
+    def sampling(self, v: Optional[Sampling]) -> None:
+        if v is not None:
+            self.properties["sampling"] = v
+        else:
+            self.properties.pop("sampling", None)
+
+    @property
+    def data_type(self) -> Optional[DataType]:
+        """Get or sets the The data type of the band.
```
duckontheweb, line 488:
```suggestion
        """Get or sets the data type of the band.
```

pystac/extensions/raster.py
```diff
@@ -0,0 +1,715 @@
...
+    @bits_per_sample.setter
+    def bits_per_sample(self, v: Optional[float]) -> None:
+        if v is not None:
+            self.properties["bits_per_sample"] = v
+        else:
+            self.properties.pop("bits_per_sample", None)
+
+    @property
+    def spatial_resolution(self) -> Optional[float]:
+        """Get or sets Average spatial resolution (in meters) of the pixels in the band.
```
duckontheweb, line 520:
```suggestion
        """Get or sets the average spatial resolution (in meters) of the pixels in the band.
```

pystac/extensions/raster.py
```diff
@@ -0,0 +1,715 @@
...
+    @spatial_resolution.setter
+    def spatial_resolution(self, v: Optional[float]) -> None:
+        if v is not None:
+            self.properties["spatial_resolution"] = v
+        else:
+            self.properties.pop("spatial_resolution", None)
+
+    @property
+    def statistics(self) -> Optional[Statistics]:
+        """Get or sets Average spatial resolution (in meters) of the pixels in the band.
```
duckontheweb, line 536:
```suggestion
        """Get or sets the average spatial resolution (in meters) of the pixels in the band.
```

pystac/extensions/raster.py
```diff
@@ -0,0 +1,715 @@
...
+    @offset.setter
+    def offset(self, v: Optional[float]) -> None:
+        if v is not None:
+            self.properties["offset"] = v
+        else:
+            self.properties.pop("offset", None)
+
+    @property
+    def histogram(self) -> Optional[Histogram]:
+        """Get or sets the Histogram distribution information of the pixels values in the band
```
duckontheweb, line 602:
```suggestion
        """Get or sets the histogram distribution information of the pixels values in the band
```

pystac/extensions/raster.py
```diff
@@ -0,0 +1,715 @@
...
+    Args:
+        item : The item to be extended.
+
+    Attributes:
+        item : The Item that is being extended.
+
+    Note:
+        Using RasterItemExt to directly wrap an item will add
+        the 'raster' extension ID to the item's stac_extensions.
+    """
```
duckontheweb, line 642: Sorry, I've been doing some work in other PRs to update these docstrings and I think this may have been copied over from an earlier version.

```suggestion
    """An abstract class that can be used to extend the properties of an
    :class:`~pystac.Item` or :class:`~pystac.Asset` with properties from 
    the :stac-ext:`Raster Extension <raster>`. This class is generic over 
    the type of STAC Object to be extended (e.g. :class:`~pystac.Item`, 
    :class:`~pystac.Asset`).

    This class will generally not be used directly. Instead, use the concrete
    implementation associated with the STAC Object you want to extend (e.g.
    :class:`~ItemRasterExtension` to extend an :class:`~pystac.Item`).
    """
```

pystac/extensions/raster.py
```diff
@@ -0,0 +1,715 @@
...
+
+    def apply(self, bands: List[RasterBand]) -> None:
+        """Applies label extension properties to the extended Item.
```
duckontheweb, line 645:
```suggestion
        """Applies raster extension properties to the extended :class:`pystac.Item` or 
        :class:`pystac.Asset`.
```

pystac/extensions/raster.py
```diff
@@ -0,0 +1,715 @@
...
+            raise pystac.ExtensionTypeError(
+                f"Raster extension does not apply to type {type(obj)}"
+            )
+
+    @staticmethod
+    def summaries(obj: pystac.Collection) -> "SummariesRasterExtension":
+        return SummariesRasterExtension(obj)
+
+
+class AssetRasterExtension(RasterExtension[pystac.Asset]):
```
duckontheweb, line 691:
```suggestion
class AssetRasterExtension(RasterExtension[pystac.Asset]):
    """A concrete implementation of :class:`RasterExtension` on an :class:`~pystac.Asset`
    that extends the Asset fields to include properties defined in the
    :stac-ext:`Raster Extension <raster>`.

    This class should generally not be instantiated directly. Instead, call
    :meth:`RasterExtension.ext` on an :class:`~pystac.Asset` to extend it.
    """
    
    asset_href: str
    """The ``href`` value of the :class:`~pystac.Asset` being extended."""

    properties: Dict[str, Any]
    """The :class:`~pystac.Asset` fields, including extension properties."""

    additional_read_properties: Optional[Iterable[Dict[str, Any]]] = None
    """If present, this will be a list containing 1 dictionary representing the
    properties of the owning :class:`~pystac.Item`."""
    
```

pystac/extensions/raster.py
```diff
@@ -0,0 +1,715 @@
...
+    def __init__(self, asset: pystac.Asset):
+        self.asset_href = asset.href
+        self.properties = asset.properties
+        if asset.owner and isinstance(asset.owner, pystac.Item):
+            self.additional_read_properties = [asset.owner.properties]
+
+    def __repr__(self) -> str:
+        return "<AssetRasterExtension Item id={}>".format(self.asset_href)
```
duckontheweb, line 699: This was an error in the original label extension code...

```suggestion
        return "<AssetRasterExtension Asset href={}>".format(self.asset_href)
```

pystac/extensions/raster.py
```diff
@@ -0,0 +1,715 @@
...
+
+
+class SummariesRasterExtension(SummariesExtension):
```
duckontheweb, line 702:
```suggestion
class SummariesRasterExtension(SummariesExtension):
    """A concrete implementation of :class:`~SummariesExtension` that extends
    the ``summaries`` field of a :class:`~pystac.Collection` to include properties
    defined in the :stac-ext:`Raster Extension <raster>`.
    """
```

duckontheweb (changes requested):
Thanks for this @emmanuelmathot. I added some suggested changes to bring the docstrings in line with some changes from recent PRs.

We should also add the classes from `pystac.extensions.raster` into the API docs by adding them to `docs/api.rst`.

emmanuelmathot:
Thank you @duckontheweb for the review!

pystac/extensions/raster.py
```diff
@@ -0,0 +1,737 @@
...
+
+    def __init__(self, properties: Dict[str, Any]) -> None:
+        self.properties = properties
+
+    def apply(
+        self,
+        count: int,
+        min: float,
+        max: float,
+        buckets: List[int],
```
duckontheweb, line 228: My interpretation of the [Histogram Object](https://github.com/stac-extensions/raster#histogram-object) in the extension docs is that all of these fields are optional since none are marked as **REQUIRED**. If that's correct, we should probably change these to optional types (e.g. `count: Optional[int]`), update the types of the associated properties below, and update the types of the `create` method arguments as well.

duckontheweb: Sorry, meant to just leave this as a comment instead of requesting changes. If they are in fact required we can leave it as-is.

emmanuelmathot: An histogram object is atomic and all fields are required for its interpretation. That's why I removed all the optional types. I will update the extension documentation to clarify this.

emmanuelmathot: https://github.com/stac-extensions/raster/pull/15

duckontheweb: 👍

duckontheweb (approved):

Issue #327, opened 2021-05-03
Implement collection summary functionality


lossyrob:
In #309, the ability to extract collection summary information for extensions was implemented for the eo extension (see `SummariesEOExtension` in `extensions/eo.py`). This should be implemented for each other extension, and tested across all the extensions, as well as pydocs written for that functionality.

Top level sphinx documentation should also be written for the summaries extensions as a pattern.

Summary information should only be used where it makes sense; e.g. if something will be a large range of discrete values that doesn't makes sense to put into a range, then we shouldn't make a summary entry for it.

Thought should be put into how the CommonMetadata works here - should we have a similar extension functionality for getting and setting common metadata summary information?

Extensions this issue potentially applies to:
- [x] datacube (no summary fields according to [stac-utils/stac-fields](https://github.com/stac-utils/stac-fields))
- [x] eo (testing) (#355)
- [x] file (Removed in #442, no summary fields according to [stac-utils/stac-fields](https://github.com/stac-utils/stac-fields))
- [x] item_assets (Collection summaries do not apply to this extension)
- [x] label (#484)
- [x] pointcloud (#558)
- [x] projection (#372)
- [x] sar (#556)
- [x] sat (#509)
- [x] scientific (#372)
- [x] timestamps (#513)
- [x] version (no summary fields according to [stac-utils/stac-fields](https://github.com/stac-utils/stac-fields))
- [x] view (#372)

duckontheweb:
@m-mohr My understanding is that Collection summaries do not apply to the Item Assets Extension since it only has Collection scope. Is that correct?

m-mohr:
Yes, indeed. (Sorry for the misclick.)

duckontheweb:
@lossyrob I think I'm going to go ahead and start the release without the Datacube Extension summaries. I can add those in a patch release early next week.

PR #372, opened 2021-05-28
Update docs, tests, summaries for stable extensions

duckontheweb:
**Related Issue(s):** #

- #327 
- #326 

**Description:**

* Updates API docs for `view`, `projection`, and `scientific` extensions
* Makes `doi` an optional argument for `ScientificExtension` and `Publication` to match spec
* Implements summaries for `view`, `projection`, and `scientific` extensions
* Simplifies summary getter/setter docstrings

`stac-fields` indicates that [JSON Schema format should be used for `sci:citation` summaries](https://github.com/stac-utils/stac-fields/blob/d0acc12ce2e6b50ab82ef788640124bb7434a868/fields.json#L510-L512), but using a simple list seemed more appropriate, so that's how I implemented it for now. @m-mohr @matthewhanson Maybe you have more insight around the rationale for using the JSON Schema summaries for that field?

**PR Checklist:**


m-mohr:
Yeah, maybe "v" is better as "s" on citation. 

We had the case where we had a regexp for the year number in the citation summary, which would be a use case. But I agree that in general, a list of values is probably the better option.

pystac/extensions/scientific.py
```diff
@@ -176,23 +205,33 @@ def remove_publication(self, publication: Optional[Publication] = None) -> None:
...
     def get_schema_uri(cls) -> str:
         return SCHEMA_URI
 
     @staticmethod
     def ext(obj: T) -> "ScientificExtension[T]":
+        """Extends the given STAC Object with properties from the :stac-ext:`Scientific
+        Extension <scientific>`.
+
+        This extension can be applied to instances of :class:`~pystac.Item` or
+        :class:`~pystac.Collection`.
```
lossyrob, line 229: I was curious about if there was a way to autodoc the type bounds of T, and found [this extension](https://sphinx-toolbox.readthedocs.io/en/latest/extensions/more_autodoc/typevars.html#directive-autotypevar). I tried it on this branch and it didn't look that great. Probably not worth it, but considering we removed type information from the docstrings maybe there's a way to get this from the TypeVar? If not, I think having type bounds in the docs for generic types is reasonable, so that a docs user can know what to expect.

pystac/extensions/view.py
```diff
@@ -3,44 +3,40 @@
...
-        Using ViewItemExt to directly wrap an item will add the 'view' extension ID to
-        the item's stac_extensions.
+    """An abstract class that can be used to extend the properties of an
+    :class:`~pystac.Item` with properties from the :stac-ext:`View Geometry
+    Extension <view>`. This class is generic over the type of STAC Object to be
+    extended (e.g. :class:`~pystac.Item`, :class:`~pystac.Asset`).
+
+    This class will generally not be used directly. Instead, use the concrete
+    implementation associated with the STAC Object you want to extend (e.g.
+    :class:`~ItemViewExtension` to extend an :class:`~pystac.Item`).
```
lossyrob, line 39: Instead of this text for these types of abstract classes, perhaps we should show an example of using the `ext` method; something like

```suggestion
    To create a concrete instance of ViewExtension, use the :meth:`ViewExtension.ext`
    method, e.g.

    >>> item: pystac.Item = ...
    >>> view_ext = ViewExtension.ext(item)
```

lossyrob (approved):
Looking good, couple of suggestions but overall +1

pystac/extensions/view.py
duckontheweb: Yeah, I like this better. I'll update all of the relevant docstrings.

duckontheweb: Fixed in a071be035dc3e26cf040c9a534dd0225ee7b5db2

pystac/extensions/scientific.py
duckontheweb: Yeah, I experimented a bit with [sphinx-autodoc-typehints](https://pypi.org/project/sphinx-autodoc-typehints/) as well, but that one has some pretty strict limitations on dealing with [circular imports](https://github.com/agronholm/sphinx-autodoc-typehints#dealing-with-circular-imports) that would mean we need to change the way we import a lot of our type variables. 

If we come up with a better solution using autodoc we can definitely change these down the road.

Issue #381, opened 2021-06-01
Drop pyright type checking


duckontheweb:
Starting with c073712, we started using [pyright](https://github.com/microsoft/pyright) instead of [mypy](https://github.com/python/mypy) for type checking in our tests/CI. There were some good reasons for this (outlined in the comments of that commit), but since then we have re-adopted mypy (via #337). 

This issue exists to discuss whether we should drop the use of pyright for type checking in our tests and CI. 

I'm personally +1 for dropping pyright for the following reasons:

- Using pyright adds a Node.js/npm dependency to a project that is otherwise purely Python
- mypy seems to be the stricter of the two type checkers, so it's not obvious to me that pyright is adding anything

schwehr:
I'm +1 on dropping pyright just because of the Node/npm dependency.  

I created #382 for the related (but definitely distinct) issue of pytype versus mypy.  

Not to mention that life will be better when the minimum python version is 3.10 (sigh...) when the typing system just understand how to handle two types that refer to each other.

l0b0:
I'm generally in favour as well, but I should probably mention that the `pre-commit` PR includes a [commit](https://github.com/stac-utils/pystac/pull/401/commits/291c898b8f2d5a5b50bb5a7f1dcc92d508942c57) which adds package.json, package-lock.json and instructions for installing the Node modules.

PR #394, opened 2021-06-02
Improve mypy strictness

l0b0:
**Related Issue(s):** #332

**Description:** Applies stricter mypy configuration to the code base

**PR Checklist:**

### DIFF
--- a/pystac/__init__.py
+++ b/pystac/__init__.py
@@ -4,36 +4,36 @@ PySTAC is a library for working with SpatioTemporal Asset Catalogs (STACs)
 
 # flake8: noqa
 from pystac.errors import (
-    STACError,  # type:ignore
-    STACTypeError,  # type:ignore
-    ExtensionAlreadyExistsError,  # type:ignore
-    ExtensionTypeError,  # type:ignore
-    RequiredPropertyMissing,  # type:ignore
-    STACValidationError,  #  type:ignore
+    STACError,
+    STACTypeError,
+    ExtensionAlreadyExistsError,
+    ExtensionTypeError,
+    RequiredPropertyMissing,
+    STACValidationError,
 )
 
 from typing import Any, Dict, Optional
 from pystac.version import (
     __version__,
-    get_stac_version,  # type:ignore
-    set_stac_version,  # type:ignore
+    get_stac_version,
+    set_stac_version,
 )
-from pystac.media_type import MediaType  # type:ignore
-from pystac.rel_type import RelType  # type: ignore
-from pystac.stac_io import StacIO  # type:ignore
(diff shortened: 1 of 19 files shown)


pystac/__init__.py
```diff
@@ -4,36 +4,36 @@
...
-    ExtensionAlreadyExistsError,  # type:ignore
-    ExtensionTypeError,  # type:ignore
-    RequiredPropertyMissing,  # type:ignore
-    STACValidationError,  #  type:ignore
+    STACError,
+    STACTypeError,
+    ExtensionAlreadyExistsError,
+    ExtensionTypeError,
+    RequiredPropertyMissing,
+    STACValidationError,
```
lossyrob, line 12: Noting here that the type ignores were to avoid showing errors for unused imports in VS Code with Pylance, with the `reportUnusedClass ` flag set. This was set in my configuration but not a project-level configuration, where the default Pyright behavior is to ignore this.

The better solution is to ignore `reportUnusedClass` in the VS Code Pylance settings (which is the default) and to ensure `flake8` is set as the linter for the project; that way unused imports that are not in __init__.py (which are ignored via the `# flake8: noqa` directive) will be caught in VS Code.

lossyrob (approved):

pystac/__init__.py
l0b0: Are we going to support *three* different type hint linters (mypy, pylance and pyright)? That sounds like a very difficult task, especially if one of them doesn't run in CI.

duckontheweb: @lossyrob My understanding of your comment is that those ignores were specific to your VS Code setup and that we are okay to drop them as @l0b0 has done here. It seems like this is mostly a PSA to other VS Code users on how to handle this. Is that right?

Might also be good to get your input on #381. My feeling is that we should settle on 1 type checker, but would be good to get other input.

lossyrob: Pylance is backed by Pyright, so they are the same thing.

I had given a thumbs up on #381. I'd like to avoid a situation where Pyright (and therefore VS Code) thinks there's an error that passes CI, as it's great to be able to detect errors during editing, I find that to be the most accelerating change in my workflow in years - reminds me of when I had static compilation feedback working with other languages. However, a single type checker makes sense, and as mypy is more strict I'm hoping we A. don't get into the situation where pyright breaks, and B. some useful type-level code is blocked by mypy. Seems like that's not happening yet; also I don't think this project needs to fit to any of our personal preferences or workflows, so I think going with mypy only is a good choice.

You are correct in that my comment here was just a note in way of explination of why the `type:ignore`s were there in the first place, and a documentation of working out how to better set up VS Code for this project.

duckontheweb (approved):

Issue #390, opened 2021-06-02
RFC: pre-commit support


l0b0:
[pre-commit](https://pre-commit.com/) is a great way to make it easier to commit only verified code, and would replace most of the contents of the `scripts` directory with configuration.

With reference to a [template project](https://github.com/linz/template-python-hello-world/):

- [.pre-commit-config.yaml](https://github.com/linz/template-python-hello-world/blob/3abc3c3ab4dd651cd71068f88cf67ba29ce4f81e/.pre-commit-config.yaml) configures each of the hooks. Picking apart the first one:
   - `entry: black` is the executable which is run
   - [`language: system`](https://pre-commit.com/#system) means it expects the executable to be on the `PATH`, for example by activating a virtualenv where Black is installed.
   - `stages: [commit]` means the command can be installed as a pre-commit hook (the default), see below. Another common stage is `commit-msg`, which can be used to verify the commit message.
   - `types: [python]` runs the executable for each *version controlled* file of this type, so there's no need to fiddle with recursive globs and unusual extensions.
- A single command, `pre-commit run --all-files`, runs all checks on all files.
- Another single command, `pre-commit install --hook-type=pre-commit --overwrite` installs Git hooks corresponding to each of the configured `commit` stage entries. This automatically runs all of the checks (even if an early one fails) when committing, printing the output of the command for reference (such as `flake8` errors) and returning a non-zero exit code if any of the tests failed. For code formatters like Black it's also clever enough to detect file changes, and treats those as errors (since they should be staged). It even takes care of stashing changed files which are not staged while running the checks, to avoid committing a broken set of lines.

What do you think? It's certainly made the daily work much easier.

gadomski:
As a development tool, it looks pretty handy. One question, which I wasn't able to answer through a quick docs skim, is "can I selectively disable/enable `pre-commit` during development?" I often use `git commit` like `command-s`, stashing changes into a work-in-progress branch before interactively rebasing them into one or more better-constructed commits later. These intermediate commits might not be "correct" per our project configuration but I don't really care, I'm working on functionality and I'll touch up the paint job later. Is `pre-commit` compatible for my workflow? Obviously doesn't need to be, but others might have similar workflows 🤷🏽.

vincentsarago:
> can I selectively disable/enable pre-commit during development?

I guess you can just do `pre-commit uninstall`

gadomski:
[...]

Issue #366, opened 2021-05-27
RFC: pre-commit hooks for QA tasks


l0b0:
Would you be interested in some simple configuration to automatically run flake8, yapf, mypy etc. when committing files? I've already set up [pre-commit](https://pre-commit.com/) in another few projects, and it's super helpful:

- No need to manually run standard QA commands before checking in.
- Checks only the files you're committing at that point, so it's fast.
- Can tie straight into the binaries installed from requirements-dev.txt.
- Can also be run in the CI pipeline, making for a simpler CI script which just runs `pre-commit run --all-files` to run all the checks.
- Runs all the checks even if some of them fail and *then* reports the exit code, so you'd get more information than the current build script.

vincentsarago:
is this the same issue as https://github.com/stac-utils/pystac/issues/390 @l0b0 ?

l0b0:
Err, yes. :blush: Didn't realise I'd already created this, but somehow the other issue got more attention.

duckontheweb:
Superseded by #390

Issue #362, opened 2021-05-25
Failing tests on Windows


duckontheweb:
The Windows CI builds have a number of failing tests related to path manipulations (see [this build](https://github.com/stac-utils/pystac/actions/runs/875975786) for an example of those failures). We should fix those failures so we can claim Windows support.

duckontheweb:
Closed by #361

PR #401, opened 2021-06-03
Introduce pre-commit

l0b0:
**Related Issues:** #390,  #366

**Description:** Use [pre-commit](https://pre-commit.com/) to enable quick checks of code before committing.

One open question: Should I get rid of the `pre-commit` call in `scripts/test`, so that it's easier to run tests without a duplicate linting run?

**PR Checklist:**


gadomski:
> One open question: Should I get rid of the pre-commit call in scripts/test, so that it's easier to run tests without a duplicate linting run?

👍🏽

duckontheweb:
> > One open question: Should I get rid of the pre-commit call in scripts/test, so that it's easier to run tests without a duplicate linting run?
> 
> 👍🏽

👍 . I ended up splitting the lint and testing scripts in #361 as well.

l0b0:
Cool, I'll split them.

l0b0:
@duckontheweb these look like the same test failures as mentioned in #362 - do we need to merge #361 first? I'll try re-running this in the meantime.

duckontheweb:
> @duckontheweb these look like the same test failures as mentioned in #362 - do we need to merge #361 first? I'm not allowed to re-run the jobs, so I can't tell if they're intermittent.

Yeah, these look like the same ones, so it might be helpful to merge that one first. If you're comfortable with the current state of that PR, feel free to approve it. I'll give other folks a little more time to comment on it and then I can merge tomorrow if nothing comes up.

l0b0:
> If you're comfortable with the current state of that PR, feel free to approve it. I'll give other folks a little more time to comment on it and then I can merge tomorrow if nothing comes up.

Sounds good! I had a closer look at the PR, and added a few comments.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+repos:
+  - repo: local
+    hooks:
+      - id: black
+        name: black
+        entry: black
+        language: system
+        stages: [commit]
+        types: [python]
```
duckontheweb, line 12: I'm sort of new to these configs, but I think we want this to run just the check (using the `--check` flag) so that it doesn't actually reformat the files as part of the CI.

```suggestion
        types: [python]
        args: [--check]
```

.github/pull_request_template.md
```diff
@@ -6,7 +6,7 @@
 
 **PR Checklist:**
 
-- [ ] Code is formatted (run `scripts/format`)
+- [ ] Code is formatted (run `pre-commit run --all-files`)
```
duckontheweb, line 9: If my other comment from `.pre-commit-config.yaml` is correct, then this would no longer reformat the code. Maybe we should just provide a direct `black` command here that developers can use.

README.md
```diff
@@ -77,33 +77,25 @@ To run linters, code formatters, and test suites all together, use `test`:
 
 ### Code quality checks
 
-PySTAC uses [flake8](http://flake8.pycqa.org/en/latest/) and [yapf](https://github.com/google/yapf) for code formatting and style checks.
-
-To run the flake8 style checks:
+PySTAC uses [flake8](http://flake8.pycqa.org/en/latest/) and [Black](https://github.com/psf/black) for code formatting and style checks. To run both:
```
duckontheweb, line 80: Not sure how others feel about this, but it seems like this section is sort of redundant with the [Contributing docs](https://github.com/stac-utils/pystac/blob/8946c9b192e20f53b75196915bb078c91954ae76/docs/contributing.rst) and maybe should be removed.

I'm fine if we want to make that change separately since it was not the goal of this PR, but if that's the case, we may want to mention `mypy` and `doc8` here as well so it's at least consistent with those Contributing docs for now.

duckontheweb (changes requested):
Thanks for contributing this! On top of helping maintain the quality of the code, it seems like it might provide a nicer interface for running some of the linting and tests.

I think it would be good to add some basic instructions to the Contributing docs on how to [install the git hook scripts](https://pre-commit.com/#3-install-the-git-hook-scripts), how to run the all of the hooks, and how to run a single hook (e.g. `pre-commit run mypy`). I know it's in the `pre-commit` docs, but it would be nice to move towards having really clear Contributing docs that are self-contained.

README.md
matthewhanson: +1 on removing and deferring to Contributing

.pre-commit-config.yaml
l0b0: I'd prefer to leave this as is, for two reasons:

1. it means people can run a single command locally to do all the formatting, and
2. it's the last command in the CI job, so only the other pre-commit jobs depend on the output.

duckontheweb: > 2\. it achieves the same thing in terms of linting (`pre-commit run` returns a non-zero exit code when anything is changed by any target),

Thanks, that's the piece I was missing. In that case this looks good to me.

README.md
duckontheweb: Opened #440 to track consolidating these docs. @l0b0 Feel free to tackle that as part of this PR if you want, but I'm also fine with just updating the README to be accurate for now and updating those docs in a separate PR.

.github/pull_request_template.md
duckontheweb: No longer relevant given the decision in [this comment](https://github.com/stac-utils/pystac/pull/401#discussion_r650903004).

README.md
l0b0: Cheers; I've removed the section in this PR.

duckontheweb (approved):
LGTM, thanks!

duckontheweb:
> I think it would be good to add some basic instructions to the Contributing docs on how to [install the git hook scripts](https://pre-commit.com/#3-install-the-git-hook-scripts), how to run the all of the hooks, and how to run a single hook

@l0b0 I was a little premature in my approval. It looks like you have the second request here covered, but could you add some brief instructions on installing the Git hooks and running a single command to the Contributing docs?

duckontheweb (approved):

PR #406, opened 2021-06-03
Fix get_bands in AssetEOExtension to return None if no eo:bands property for the asset

### LINKED ISSUE #404: [BUG] EO extension get_bands on a non-EO asset returns Item bands
## Expected Behavior

When using `EOExtension.ext(asset).bands` on an asset that has no bands, it must return `None`.

## Current Behavior

When using `EOExtension.ext(asset).bands` on an asset that has no bands, it may return all Item bands if declared at Item level

## Identified issue

`get_bands` [method](https://github.com/stac-utils/pystac/blob/main/pystac/extensions/eo.py#L329) in EOExtension uses [`PropertiesExtension._get_property`](https://github.com/stac-utils/pystac/blob/3f2f95fc8dd4534bb64f03b6e388b22d092414bd/pystac/extensions/base.py#L60) that extends the retrieval to the asset's parent Item.

## Possible Solution

change the _get_property method with a simpler get in asset properties only.

## Steps to Reproduce

```python
item = pystac.Item.from_file("https://gist.githubusercontent.com/emmanuelmathot/00e8738216f4a359a42b25ea2ca56271/raw/309ee298080978c692fa729a2ba589f95bfacb1c/S2A_MSIL1C_20200615T060641_N0209_R134_T45WWS_20200615T080720.json")
for key, asset in item.get_assets().items():
    eo_bands = EOExtension.ext(asset).bands
    if (key == "mtd"):
        assert eo_bands is None
```

## Context (Environment)

* python 3.8
* pystac 1.0.0-beta.3

As defined in https://github.com/stac-extensions/eo/:
> The eo:bands array may optionally be used in the Item Properties to summarize the available bands in the assets. This should be a 'union' of all the possible bands represented in assets. It should be considered merely informative - clients should rely on the eo:bands of each asset.

emmanuelmathot:
**Related Issue(s):** #404

**Description:**

- fix `get_bands` in `AssetEOExtension` to return None if no `eo:bands` property for the asset
- test to reproduce #404 case

**PR Checklist:**

- [X] Code is formatted (run `scripts/format`)
- [X] Changes are added to the [CHANGELOG](https://github.com/stac-utils/pystac/blob/develop/CHANGELOG.md). See [the docs](https://pystac.readthedocs.io/en/latest/contributing.html#changelog) for information about adding to the changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Fixed
 
-- Fixed returned None by `EOExtension.get_bands` for asset without EO bands
+- Fixed returned None by `EOExtension.get_bands` for asset without EO bands ([#406](https://github.com/stac-utils/pystac/pull/406))
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Fixed
 
+- Fixed returned None by `EOExtension.get_bands` for asset without EO bands
+
 ### Removed
 
 ## [1.0.0-beta.3]
--- a/tests/extensions/test_eo.py
+++ b/tests/extensions/test_eo.py
@@ -42,6 +42,7 @@ class EOTest(unittest.TestCase):
         "data-files/eo/sample-bands-in-item-properties.json"
     )
     EO_COLLECTION_URI = TestCases.get_path("data-files/eo/eo-collection.json")
+    S2_ITEM_URI = TestCases.get_path("data-files/eo/eo-sentinel2-item.json")
 
     def setUp(self) -> None:
         self.maxDiff = None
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ### Fixed
 
+- Fixed returned None by `EOExtension.get_bands` for asset without EO bands
```
duckontheweb, line 11: Let's add the link to this PR.

duckontheweb (approved):
Just 1 comment related to the CHANGELOG, but LGTM otherwise. Thanks!

tests/extensions/test_eo.py
```diff
@@ -83,6 +84,13 @@ def test_bands(self) -> None:
         self.assertEqual(len(EOExtension.ext(item).bands or []), 3)
         item.validate()
 
+    def test_asset_bands_s2(self) -> None:
+        item = pystac.Item.from_file(self.S2_ITEM_URI)
+        for key, asset in item.get_assets().items():
+            eo_bands = EOExtension.ext(asset).bands
+            if key == "mtd":
+                assert eo_bands is None
+
```
duckontheweb, line 93: Sorry I didn't catch this earlier. It seems like it would be simpler and faster to get the `"mtd"` asset directly rather than looping over all of the assets. Since we're not making any assertions on the bands for the other assets it doesn't seem necessary to construct the extension and get the `Bands` objects for those.

```suggestion
        item = pystac.Item.from_file(self.S2_ITEM_URI)
        mtd_asset = item.get_assets()["mtd"]
        self.assertIsNone(EOExtension.ext(asset).bands)
```

emmanuelmathot: yes. Initially, the test was intended to detect some "residual" objects from other assets

Issue #389, opened 2021-06-01
eo-collection.json seems incorrect - no extensions


schwehr:
https://github.com/stac-utils/pystac/blob/ac5237f8c129ed0cd7ff4e196f42da6669b1d0b2/tests/data-files/eo/eo-collection.json

```json
{
  "type": "Collection",
  "id": "landsat-8-l1",
  "stac_version": "1.0.0-rc.4",

  "stac_extensions": [],  "title": "Landsat 8 L1",

  "summaries": { 
    "eo:bands": [
      {
        "name": "B1",
        "common_name": "coastal",
        "center_wavelength": 0.44,
        "full_width_half_max": 0.02
      },

  ],
  "eo:cloud_cover": {
      "minimum": 0.0,
      "maximum": 80.0
    }
  },
```

Shouldn't this have an extension entry with the EO extension?

m-mohr:
Yes, it should.

PR #408, opened 2021-06-04
Fix example collections with extension summaries

duckontheweb:
**Related Issue(s):** #

* #389 

**Description:**

* Adds extension URIs to examples collections with summaries of extension fields

**PR Checklist:**


lossyrob (changes requested):
One other that I think needs this treatment:
tests/data-files/catalogs/planet-example-v1.0.0-beta.2/collection.json (view, eo)


Noticing as well that there's a set of collections in `tests/data-files/commons` that are version 1.0 but contain properties with extensions; I believe they were being used to test the commons extensions when that was still a thing (merging collection properties into items). Perhaps we should drop them or set the version to one that had the commons extension and avoid migrating it for future versions.

duckontheweb:
> Noticing as well that there's a set of collections in `tests/data-files/commons` that are version 1.0 but contain properties with extensions; I believe they were being used to test the commons extensions when that was still a thing (merging collection properties into items). Perhaps we should drop them or set the version to one that had the commons extension and avoid migrating it for future versions.

Since the commons extension was [removed in `v1.0.0-beta.1`](https://github.com/radiantearth/stac-spec/blob/master/CHANGELOG.md#v100-beta1---2020-05-29), we should probably keep testing against it and just avoid migrating it in the future. It looks like putting it in a version-specific directory would do the trick since [this line](https://github.com/stac-utils/pystac/blob/46589997634bbb04892f60f727c5266755104481/tests/data-files/change_stac_version.py#L61-L62) would skip it during the migration.

duckontheweb:
> > Noticing as well that there's a set of collections in `tests/data-files/commons` that are version 1.0 but contain properties with extensions; I believe they were being used to test the commons extensions when that was still a thing (merging collection properties into items). Perhaps we should drop them or set the version to one that had the commons extension and avoid migrating it for future versions.
> 
> Since the commons extension was [removed in `v1.0.0-beta.1`](https://github.com/radiantearth/stac-spec/blob/master/CHANGELOG.md#v100-beta1---2020-05-29), we should probably keep testing against it and just avoid migrating it in the future. It looks like putting it in a version-specific directory would do the trick since [this line](https://github.com/stac-utils/pystac/blob/46589997634bbb04892f60f727c5266755104481/tests/data-files/change_stac_version.py#L61-L62) would skip it during the migration.

Actually, I can't find anywhere were those `data-files/commons` files are used specifically. We already have examples in `data-files/examples/0.9.0/extensions/commons/` that we test against for serialization and validation, so I think we're probably safe to just remove the `data-files/commons` files entirely.

duckontheweb:
@lossyrob I updated the files you mentioned and also found a few more that needed fixing.

tests/data-files/catalogs/test-case-6/12cb17a8-ae08-469c-a2be-4d0619240014/9420e0fd-3731-40e8-93ff-5b6ad0067048/collection.json
```diff
@@ -76,7 +76,9 @@
       "label:assets": []
     }
   ],
-  "stac_extensions": [],
+  "stac_extensions": [
+    "https://stac-extensions.github.io/label/v1.0.0/schema.json"
+  ],
```
lossyrob, line 81: This is curious - I'm not sure why there are `label:assets` in those properties. I don't think this collection should actually extend label, and I don't think `label:asssets` should be included in the links. Perhaps there's a bug in the migration script?

duckontheweb: It looks like that test case was added in response to #88, which was way back in PySTAC 0.3.3. It had the `label:assets` properties when it was created, so they must have been relevant at the time, but they definitely don't fit now.

I'll remove the Label extension references in that example.

duckontheweb: Fixed in f020601e502eec3feb309633c200df9c9f1bc8e0

lossyrob (approved):

duckontheweb:
@schwehr Let me know if you have any feedback on this. Otherwise, I'll plan on merging by the end of the day.

schwehr:
It looks fine as far as my limited understanding goes.   I didn't look at the unchanged rest of the files in detail.  Thanks for doing the cleanup.

PR #417, opened 2021-06-07
Drop pyright for type checking

### LINKED ISSUE #381: Drop pyright type checking

duckontheweb:
**Related Issue(s):** #

* Closes #381

**Description:**

Drops the use of `pyright` for type checking. The `test` script (and therefore the CI, as well), would now use only `mypy` for type checking.

*Note:* This does not update the [Code quality checks](https://github.com/stac-utils/pystac#code-quality-checks) section of the main README because that section is already out of date and would be updated by #401.

**PR Checklist:**


l0b0:
@duckontheweb Do you want to merge this before #401? Then I can remove the pyright stuff from it.

lossyrob (approved):

lossyrob:
@l0b0 just merged

Issue #370, opened 2021-05-27
Raise exception on attempt to extend object not implementing that extension


duckontheweb:
Calling, for instance, `pystac.extensions.eo.EOExtension.ext` on an `Item` that does not have the EO schema URI in its `stac_extensions` list will succeed and return an `ItemEOExtension` instance. You can then read and write fields like `eo:cloud_cover` even though the `Item` doesn't actually implement that extension.

It seems like the desired behavior here would be to raise an exception when using `ext` on an object that does not implement that extension for 2 reasons:

1. When using PySTAC to read existing STAC Items, this could mask a problem with an invalid object
2. This will interfere with the edge case where a user has custom fields with a prefix that overlaps with one of the extensions (maybe because the custom fields predate the extension). In this case, the fields should not be treated as if they are defined by the extension.

Originally raised as part of [this PR comment](https://github.com/stac-utils/stactools/pull/113#discussion_r638974407) in stac-utils/stactools

gadomski:
I haven't seen many cases where an `Asset` exists before the `Item` does. My instinct is to raise an exception when trying to extend an un-parented `Asset`. If folks really need to work around that exception, they can modify the properties directly.

lossyrob:
My preference would be for an exception not to be thrown if the Asset doesn't have an owner, and perhaps give a warning in the documentation. This way users can build up assets before attaching them to items legitimately if that is better for their workflow. I'd rather be trusted by the library to do the right thing than to be blocked by an exception for a legitimate usage.

gadomski:
@lossyrob how should the attachment work in this case? Would you want the `Item` to "inherit" extensions from the `Asset`? Or is it the dev's responsibility to turn on the extension in the `Item`?

lossyrob:
The latter case - if you use an `*Extension` object to access properties on the Asset that doesn't have an Item, then it's on the user to make sure the Item it is associated with has the proper extensions enabled.

duckontheweb:
The other option I considered was to make the `Asset` class "extension-aware" in some way. Then we could do a check when the asset is added to an Item and add extension URIs as necessary.

[...]

PR #419, opened 2021-06-08
Make ExtensionManagementMixin.add_to idempotent

### LINKED ISSUE #369: ExtensionManagementMixin.add_to should be idempotent
Calling the `add_to` method on a concrete sub-class of `pystac.extensions.base.ExtensionManagementMixin` multiple times will add that extension's schema URI to the `stac_extensions` list multiple times.

We should ensure that schema URIs are only added to an object's `stac_extensions` list if they are not already there.

Alternatively, we could make `stac_extensions` a `set`, since there shouldn't ever be a case where schema URIs are duplicated. This would involve changing the way values are added to that field throughout the code base, and may have implications on how objects are serialized to JSON, but would probably improve performance over checking if an element is in the list.

duckontheweb:
**Related Issue(s):** #

* Closes #369 

**Description:**

`pystac.extensions.base.ExtensionManagementMixin.add_to` will only add the schema URI for the inheriting extension class if it is not already in the object's `stac_extensions` list. 

I looked into making `stac_extensions` a `set` instead of a `list`, but the extra complexity to ensure that the set was properly serialized and deserialized did not seem worth the effort. If that seems like the better approach, I can explore it in more depth, though.

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@
 - `identify_stac_object_type` returns `None` and `identify_stac_object` raises `STACTypeError` for non-STAC objects
   ([#402](https://github.com/stac-utils/pystac/pull/402))
 - `ExtensionManagementMixin.add_to` is now idempotent (only adds schema URI to
-  `stac_extensions` once per `Item` regardless of the number of calls)
+  `stac_extensions` once per `Item` regardless of the number of calls) ([#419](https://github.com/stac-utils/pystac/pull/419))
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@
 - `EOExtension.get_bands` returns `None` for asset without EO bands ([#406](https://github.com/stac-utils/pystac/pull/406))
 - `identify_stac_object_type` returns `None` and `identify_stac_object` raises `STACTypeError` for non-STAC objects
   ([#402](https://github.com/stac-utils/pystac/pull/402))
+- `ExtensionManagementMixin.add_to` is now idempotent (only adds schema URI to
+  `stac_extensions` once per `Item` regardless of the number of calls)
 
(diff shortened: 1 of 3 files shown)


gadomski (approved):
👍🏽 to keepin' it simple for now, and if we identify this as a performance bottleneck later we know how to proceed (seems unlikely).

lossyrob:
We also talked about checking that the object to see if it has the extension on `FooExtension.ext(...)`, and throwing if it didn't exist, and also adding a `add_if_missing` flag to that method to allow a one-liner to add the extension and get the extension object. Is that still a plan? If so, should that be tackled in another PR or added to this?

duckontheweb:
> We also talked about checking that the object to see if it has the extension on `FooExtension.ext(...)`, and throwing if it didn't exist, and also adding a `add_if_missing` flag to that method to allow a one-liner to add the extension and get the extension object. Is that still a plan? If so, should that be tackled in another PR or added to this?

I started to tackle that as part of this PR, but there's a wrinkle with how to handle that when extending an Asset with no owner, since it won't have any extension info associated with it. I figured I would work through that in comments on #370 and merge this in the meantime, but I'm also fine with keeping this open until that's resolved.

duckontheweb:
> I figured I would work through that in comments on #370 and merge this in the meantime, but I'm also fine with keeping this open until that's resolved.

I will actually just implement that as part of this PR and we can get feedback on the approach that way. Probably easier than trying to discuss it in an issue.

duckontheweb:
@lossyrob Sorry to keep waffling on this, but I think it would be best to merge this and tackle #370 in a separate PR. There are some challenges around how and where to implement that exception and how to handle `Asset` instances without an owner that might be best to work out separately, and it probably doesn't make sense to hold up this change in the meantime.

lossyrob (approved):

PR #420, opened 2021-06-08
Test against Python 3.9

duckontheweb:
**Description:**

Expands CI test matrix to include Python 3.9 and 3.10. Python 3.10 is marked as "experimental", so the CI workflow will still succeed even if that job fails.

**PR Checklist:**

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -18,6 +18,7 @@ jobs:
           - "3.6"
           - "3.7"
           - "3.8"
+          - "3.9"
         os:
           - ubuntu-latest
           - windows-latest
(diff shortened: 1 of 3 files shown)


duckontheweb:
This is currently failing because of the bump to `orjson==3.5.3`. That version does not yet have a wheel for Python 3.10 and building from source requires the nightly version of Rust, which is not configured for this environment. The `orjson` maintainers seem to be good about creating these wheels (they exists for `3.5.2` on Python 3.10), so we should just wait for those to be available.

l0b0 (approved):

l0b0:
Looks good! How about introducing 3.9 and 3.10 support in separate PRs?

duckontheweb:
> How about introducing 3.9 and 3.10 support in separate PRs?

Good call. I updated this PR to focus on 3.9 and will open a separate PR for testing 3.10.

lossyrob (approved):

Issue #380, opened 2021-05-31
Equality of pystac objections?


TomAugspurger:
In writing tests for #379, I stumbled over Asset objects not defining `__eq__`.

```python
In [1]: import pystac

In [2]: pystac.Asset(href='a') == pystac.Asset(href='a')
Out[2]: False
```

Looking briefly, I don't see it for any of the core items (a couple extensions define `__eq__`). The behavior of Python is to fall back to object identity, (`a is b`).

Should pystac objects define `__eq__` that's roughly

```python
def __eq__(self, other):
    return type(self) == type(other) and self.to_dict() == other.to_dict()
```

duckontheweb:
I'm in favor of defining equality for PySTAC objects. 

As I started working on an `ItemCollection` class for #371 I realized it would be nice to be able to define `__contains__` on that class to check if an `ItemCollection` contains an `Item`. This is currently possible because I am cloning each `Item` during `ItemCollection` initialization so that we can reset the root link without mutating the original object. Since the default equality measure is `item_a is item_b`, as @TomAugspurger mentioned, this [always comes back `False`](https://github.com/duckontheweb/pystac/blob/575fbe918164288aba39bf992749054ee68fd258/tests/test_item_collection.py#L35-L40).

I can tackle this as part of the work on #371. It sounds like we should raise a `NotImplementedError` for values of `other` that cannot be compared, so I'm thinking this should be the implementation:

```python
def __eq__(self, other: object) -> bool:
    if type(self) != type(other):
        raise NotImplementedError(f"Type of other must be {type(self)}"
    return self.to_dict() == other.to_dict()

def __ne__(self, other: object) -> bool:
    if type(self) != type(other):
        raise NotImplementedError(f"Type of other must be {type(self)}"
    return self.to_dict() != other.to_dict()
```

That article mentions that Python 3 automatically makes objects unhashable if you define `__eq__` and not `__hash__`, but I'll double-check that.

TomAugspurger:
> It sounds like we should raise a NotImplementedError for values of other that cannot be compared, so I'm thinking this should be the implementation:

I haven't read closely, but I want to confirm: do you want `raise NotImplementedError` or `return NotImplemented`. The (subtle) difference is that `return NotImplemented` allows the `other` object to determine if it's equal or not. I think if both return NotImplemented then Python raises a TypeError.

[...]

Issue #371, opened 2021-05-28
compatibility with pystac-client


scottyhq:
I expect released pystac-client and pystac versions to work well together, and am currently encountering a `KeyError: 'id'` when opening saved search results from pystac-client with pystac. 

Related issue  #256, #302 and https://github.com/stac-utils/pystac-client/issues/1 

```python
from pystac_client import Client # 0.1.1 
import pystac # 0.5.6

aoi = {'type': 'Polygon',
 'coordinates': [[[-108.29, 38.88],
   [-107.97, 38.88],
   [-107.97, 39.10],
   [-108.29, 39.10],
   [-108.29, 38.88]]]}

catalog = Client.open("https://cmr.earthdata.nasa.gov/stac/LPCLOUD")
results = catalog.search(collections=['HLSS30.v1.5'], #S2
                         intersects=aoi,
                         datetime='2021-01-01/2021-01-03',
                         )
print(f"{results.matched()} items found")

items = results.items_as_collection()
items.save('items.json')

pystac.read_file('items.json')
```

```pytb
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-95-12b3d23b2c32> in <module>
     20 items.save('items.json')
     21 
---> 22 pystac.read_file('items.json')

/srv/conda/envs/notebook/lib/python3.8/site-packages/pystac/__init__.py in read_file(href)
     71         by the JSON read from the file located at HREF.
     72     """
---> 73     return STACObject.from_file(href)
     74 
     75 

/srv/conda/envs/notebook/lib/python3.8/site-packages/pystac/stac_object.py in from_file(cls, href)
    492 
    493         if cls == STACObject:
--> 494             o = STAC_IO.stac_object_from_dict(d, href=href)
    495         else:
    496             o = cls.from_dict(d, href=href)

/srv/conda/envs/notebook/lib/python3.8/site-packages/pystac/serialization/__init__.py in stac_object_from_dict(d, href, root)
     35 
     36     if info.object_type == STACObjectType.CATALOG:
---> 37         return Catalog.from_dict(d, href=href, root=root)
     38 
     39     if info.object_type == STACObjectType.COLLECTION:

/srv/conda/envs/notebook/lib/python3.8/site-packages/pystac/catalog.py in from_dict(cls, d, href, root)
    788         d = deepcopy(d)
    789 
--> 790         id = d.pop('id')
    791         description = d.pop('description')
    792         title = d.pop('title', None)

[...]

PR #402, opened 2021-06-03
Raise exception in identify_stac_object for non-STAC objects

duckontheweb:
**Related Issue(s):** #

* #340 

**Description:**

* Fixes the logic in `serialization.identify.identify_stac_object_type` for all STAC object v0.8.* and above
    As per #340 and [this comment](https://github.com/stac-utils/pystac/issues/343#issuecomment-847892370), this was previously falling back to identifying JSON-like dicts as Catalogs if they did not meet the other criteria.
* `serialization.identify.identify_stac_object_type` now returns `None` if the dict is not a STAC object, and `serialization.identify.identify_stac_object` raises a `STACTypeError` in this case
* Removes all examples for STAC versions lower than 0.8.

* **UPDATE 2021-06-04:** Removes some logic in the EO extension migration and in `serialization.migrate` that was specific to pre-0.8 STAC versions, since this was causing a significant drop in test coverage.

**PR Checklist:**

- [x] Code is formatted (run `scripts/format`)
- [x] Tests pass (run `scripts/test`)
- [x] This PR maintains or improves overall codebase code coverage.
- [ ] Changes are added to the [CHANGELOG](https://github.com/stac-utils/pystac/blob/develop/CHANGELOG.md). See [the docs](https://pystac.readthedocs.io/en/latest/contributing.html#changelog) for information about adding to the changelog.

PR #430, opened 2021-06-10
Add basic ItemCollection implementation

duckontheweb:
**Related Issue(s):** #

* #371 

**Description:**

Adds `ItemCollection` class to work with GeoJSON Feature Collections containing only STAC Items. This class can be serialized/deserialized from JSON, and also implements [`Collection`](https://docs.python.org/3/library/typing.html#typing.Collection) to make it more convenient to loop over items, get an item by index, and check if an Item is included. Note that until we implement more specific STAC object equality (as per #380), `assert item in item_collection` will always fail unless `clone_items=False` is used when instantiating the `ItemCollection`.

This PR also updates the I/O functions in `pystac.__init__` to work with the new `ItemCollection` class in response to #371. Since `ItemCollection`s are not `STACObject`s, this means that the read functions in that file (`read_file`, `read_dict`) are now not guaranteed to return a `STACObject`. This means users will need to either handle both cases in their code, or `cast` the result of those functions to the appropriate type. **I'm interested in feedback on whether this is worth the change, or if we should keep these functions unchanged and make people use `ItemCollection.from_file` in order to read ItemCollections.**

**PR Checklist:**

cc: @scottyhq

**UPDATE 2021-06-10:**

Implemented `__contains__` to make `ItemCollection` a `typed.Collection` since we now have the option not to clone Items on instantiation and we are no longer manipulating the Item links.

### DIFF
--- a/pystac/__init__.py
+++ b/pystac/__init__.py
@@ -12,7 +12,7 @@ from pystac.errors import (
     STACValidationError,
 )
 
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, Optional
 from pystac.version import (
     __version__,
     get_stac_version,
@@ -72,7 +72,7 @@ EXTENSION_HOOKS = pystac.extensions.hooks.RegisteredExtensionHooks(
 )
 
 
-def read_file(href: str) -> Union[STACObject, ItemCollection]:
+def read_file(href: str) -> STACObject:
     """Reads a STAC object from a file.
 
     This method will return either a Catalog, a Collection, or an Item based on what the
@@ -86,15 +86,18 @@ def read_file(href: str) -> Union[STACObject, ItemCollection]:
     Returns:
         The specific STACObject implementation class that is represented
         by the JSON read from the file located at HREF.
+
+    Raises:
+        STACTypeError : If the file at ``href`` does not represent a valid
+            :class:`~pystac.STACObject`. Note that an :class:`~pystac.ItemCollection` is not
+            a :class:`~pystac.STACObject` and must be read using
+            :meth:`ItemCollection.from_file <pystac.ItemCollection.from_file>`
     """
-    try:
-        return STACObject.from_file(href)
-    except STACTypeError:
-        return ItemCollection.from_file(href)
+    return STACObject.from_file(href)
 
 
 def write_file(
-    obj: Union[STACObject, ItemCollection],
+    obj: STACObject,
     include_self_link: bool = True,
     dest_href: Optional[str] = None,
 ) -> None:
@@ -113,19 +116,11 @@ def write_file(
     Args:
         obj : The STACObject to save.
         include_self_link : If ``True``, include the ``"self"`` link with this object.
-            Otherwise, leave out the self link. Ignored for :class:~ItemCollection`
-            instances.
+            Otherwise, leave out the self link.
         dest_href : Optional HREF to save the file to. If ``None``, the object will be
-            saved to the object's ``"self"`` href (for :class:`~STACObject` sub-classes)
-            or a :exc:`~STACError` will be raised (for :class:`~ItemCollection`
-            instances).
+            saved to the object's ``"self"`` href.
     """
-    if isinstance(obj, ItemCollection):
-        if dest_href is None:
-            raise STACError("Must provide dest_href when saving and ItemCollection.")
-        obj.save_object(dest_href=dest_href)
-    else:
-        obj.save_object(include_self_link=include_self_link, dest_href=dest_href)
+    obj.save_object(include_self_link=include_self_link, dest_href=dest_href)
 
 
 def read_dict(
@@ -133,31 +128,32 @@ def read_dict(
     href: Optional[str] = None,
     root: Optional[Catalog] = None,
     stac_io: Optional[StacIO] = None,
-) -> Union[STACObject, ItemCollection]:
+) -> STACObject:
     """Reads a :class:`~STACObject` or :class:`~ItemCollection` from a JSON-like dict
     representing a serialized STAC object.
 
     This method will return either a :class:`~Catalog`, :class:`~Collection`,
-    :class`~Item`, or :class:`~ItemCollection` based on the contents of the dict.
+    or :class`~Item` based on the contents of the dict.
 
     This is a convenience method for either
-    :meth:`stac_io.stac_object_from_dict <stac_io.stac_object_from_dict>` or
-    :meth:`ItemCollection.from_dict <ItemCollection.from_dict>`.
+    :meth:`stac_io.stac_object_from_dict <stac_io.stac_object_from_dict>`.
 
     Args:
         d : The dict to parse.
         href : Optional href that is the file location of the object being
-            parsed. Ignored if the dict represents an :class:`~ItemCollection`.
+            parsed.
         root : Optional root of the catalog for this object.
             If provided, the root's resolved object cache can be used to search for
-            previously resolved instances of the STAC object.  Ignored if the dict
-            represents an :class:`~ItemCollection`.
+            previously resolved instances of the STAC object.
         stac_io: Optional :class:`~StacIO` instance to use for reading. If ``None``,
             the default instance will be used.
+
+    Raises:
+        STACTypeError : If the ``d`` dictionary does not represent a valid
+            :class:`~pystac.STACObject`. Note that an :class:`~pystac.ItemCollection` is not
+            a :class:`~pystac.STACObject` and must be read using
+            :meth:`ItemCollection.from_dict <pystac.ItemCollection.from_dict>`
     """
     if stac_io is None:
         stac_io = StacIO.default()
-    try:
-        return stac_io.stac_object_from_dict(d, href, root)
-    except STACTypeError:
-        return ItemCollection.from_dict(d)
+    return stac_io.stac_object_from_dict(d, href, root)
--- a/pystac/__init__.py
+++ b/pystac/__init__.py
@@ -12,7 +12,7 @@ from pystac.errors import (
     STACValidationError,
 )
 
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, Union
 from pystac.version import (
     __version__,
     get_stac_version,
@@ -72,7 +72,7 @@ EXTENSION_HOOKS = pystac.extensions.hooks.RegisteredExtensionHooks(
 )
 
 
-def read_file(href: str) -> STACObject:
+def read_file(href: str) -> Union[STACObject, ItemCollection]:
     """Reads a STAC object from a file.
 
     This method will return either a Catalog, a Collection, or an Item based on what the
@@ -87,11 +87,16 @@ def read_file(href: str) -> STACObject:
         The specific STACObject implementation class that is represented
         by the JSON read from the file located at HREF.
     """
-    return STACObject.from_file(href)
+    try:
+        return STACObject.from_file(href)
+    except STACTypeError:
+        return ItemCollection.from_file(href)
 
 
 def write_file(
-    obj: STACObject, include_self_link: bool = True, dest_href: Optional[str] = None
+    obj: Union[STACObject, ItemCollection],
+    include_self_link: bool = True,
+    dest_href: Optional[str] = None,
 ) -> None:
     """Writes a STACObject to a file.
 
@@ -107,12 +112,20 @@ def write_file(
 
     Args:
         obj : The STACObject to save.
-        include_self_link : If this is true, include the 'self' link with this object.
-            Otherwise, leave out the self link.
-        dest_href : Optional HREF to save the file to. If None, the object will be saved
-            to the object's self href.
+        include_self_link : If ``True``, include the ``"self"`` link with this object.
+            Otherwise, leave out the self link. Ignored for :class:~ItemCollection`
+            instances.
+        dest_href : Optional HREF to save the file to. If ``None``, the object will be
+            saved to the object's ``"self"`` href (for :class:`~STACObject` sub-classes)
+            or a :exc:`~STACError` will be raised (for :class:`~ItemCollection`
+            instances).
     """
-    obj.save_object(include_self_link=include_self_link, dest_href=dest_href)
+    if isinstance(obj, ItemCollection):
+        if dest_href is None:
+            raise STACError("Must provide dest_href when saving and ItemCollection.")
+        obj.save_object(dest_href=dest_href)
+    else:
+        obj.save_object(include_self_link=include_self_link, dest_href=dest_href)
 
 
 def read_dict(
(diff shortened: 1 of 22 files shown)


matthewhanson:
I think I lean toward `STACObject.read_file` only ever returning a STACObject.  Seems wrong that a class factory function would return something other than that class.

I think it's ok to require users having to actively open an ItemCollection with ItemCollection.from_file. As long as they know that's how you open an ItemCollection.

I know it doesn't meet the request from @scottyhq , but I think that's just because before it was ambiguous on if an ItemCollection was a STACObject, because there was the ItemCollection class in stac-api-spec, and it had `stac_version`, etc. 

Now that it's clear an ItemCollection is just a GeoJSON FeatureCollection with features that are STAC Items, it's reasonable that it you would open it differently.

matthewhanson:
Wait, I think I'm wrong here....I don't think a STACObject function should return a non STACObject.

But the pystac.read_file is just a convenience function, and now that I look at the code I think this is exactly the right way to do it.  STACObject still only returns a STACObject, but the `pystac.read_file` function can return any class defined within the PySTAC library

pystac/item_collection.py
```diff
@@ -0,0 +1,125 @@
...
+    ``ItemCollection``."""
+
+    extra_fields: Dict[str, Any]
+    """Dictionary containing additional top-level fields for the GeoJSON
+    FeatureCollection."""
+
+    def __init__(
+        self, items: List[pystac.Item], extra_fields: Optional[Dict[str, Any]] = None
+    ):
+        self.items = [item.clone() for item in items]
```
matthewhanson, line 48: I think this is the proper way to do this so that a developer doesn't create an ItemCollection without knowing it could modify the original Items.

However, in most cases it's unnecessary since the most common use case is to read an ItemCollection from somewhere else using the `from_dict` method where the Items were just created.

@tomaugspurger wrote up a related issue, and I fixed some of it in pystac-client by having an option to just fetch pages as JSON rather than always ItemCollection, but when an ItemCollection is created it's going to be hit by the deepcopy after creation issue here.

matthewhanson: https://github.com/stac-utils/pystac-client/issues/49

pystac/item_collection.py
```diff
@@ -0,0 +1,125 @@
+from copy import deepcopy
+from pystac.errors import STACTypeError
+from typing import Any, Dict, Iterator, List, Optional, Sized, Iterable
+
+import pystac
+from pystac.utils import make_absolute_href, is_absolute_href
+from pystac.serialization.identify import identify_stac_object_type
+
+
+class ItemCollection(Sized, Iterable[pystac.Item]):
```
matthewhanson, line 10: In the latest iteration in pystac-client I had links as a member of ItemCollection:
https://github.com/stac-utils/pystac-client/blob/mah/updates/pystac_client/item_collection.py#L24

Here it would just be included in extra_fields, but only just as JSON, whereas with a Links array of Link types we can use the PySTAC Link object which is sorta nice.  But that implies that an ItemCollection is it's own defined Object, but we don't have it defined anywhere as having Links outside the API search response.

pystac-client doesn't need them to be Link types because it uses the raw JSON pagination links when paging so while this isn't required it does seem useful.

The most common way of using this is through pystac-client or just for opening up saved search results, and being able to understand the included links would be a nice feature.

pystac/item_collection.py
```diff
@@ -0,0 +1,125 @@
...
+    extra_fields: Dict[str, Any]
+    """Dictionary containing additional top-level fields for the GeoJSON
+    FeatureCollection."""
+
+    def __init__(
+        self, items: List[pystac.Item], extra_fields: Optional[Dict[str, Any]] = None
+    ):
+        self.items = [item.clone() for item in items]
+        for item in self.items:
+            item.clear_links("root")
```
matthewhanson, line 50: I've been thinking about this. On one hand keeping root and other hierarchical links, is nice because it essentially provides information on where the Items originally came from.  You can see the original root endpoint it came from.

On the other hand, these new Items no longer have a root themselves, nor parents, so what do their inclusion mean?  If you split out this ItemCollection into separate Item files, is that when  you would strip out the hierarchical links?   Or when you edit the file?   If the ItemCollection is simply an unaltered copy from an API then keeping the links probably makes sense.  Not sure the best approach here.

- `root` would be a way of identifying the original catalog the Item came from
- `parent` doesn't seem useful in an ItemCollection at all. I think `root` is better.  The `parent` is either going to be a collection (if from an API), or a sub-catalog if this ItemCollection was created manually from a bunch of Items.
- `collection` links should be kept, as the ItemCollection doesn't contain collections itself, a way is needed to fetch the original collection this Item belonged to
- `

pystac/item_collection.py
duckontheweb: Maybe it would make sense to have a `clone_items` parameter that defaults to `True` that would determine whether or not we clone the Items here. If `clone_items=False` then we would also want to make sure that we don't alter any links automatically, since that could cause problems in the associated Collections/Catalogs.

pystac/item_collection.py
duckontheweb: Since the `ItemCollection` is such a generic vessel for Items, it seems like the best course here might be to not take any action on the Items at all (other than possible cloning, per the comment above). If/when we introduce other convenience methods for converting to/from `Collection` instance, we could handle the links more specifically. It would also be more transparent to users if we just put Items into the ItemCollection unaltered and let them handle any cleanup they want to do.

pystac/item_collection.py
matthewhanson: This makes sense to me, let's keep the basic functionality. We can extend it as needed.

pystac/item_collection.py
matthewhanson: Agree, since we don't really know what should be done here we should just leave it as is. An ItemCollection is just a collection of Items as they are.

pystac/item_collection.py
matthewhanson: Another idea I just had is to reconsider the whole initialization of an ItemCollection.

The __init__ class could take in JSON instead of expecting them to already be Items, and from_dict would pass in the features dictionary to __init__ as JSON and the Item() constructor occurs here in the ItemCollection constructor.

This way there wouldn't be a double hit and the user wouldn't have to do anything different. This would be the primary way of using the class.   

But then constructing an ItemCollection is harder...there could be a separate factory function but that would have to deserialize the Items then call  __init__.

How about we change Items to be `Union[List[Dict], List[Item]] and handle both cases?  from_dict won't do the Item() construction.

duckontheweb: I want to make sure I'm understanding the potential use-cases that we want to support here. Here's my summary of the use-cases for an ItemCollection, let me know if I'm missing anything:

1. Deserialize a `pystac.ItemCollection` directly from either a JSON FeatureCollection file or a JSON FeatureCollection-like `dict`

    As per the suggestion in stac-utils/pystac-client#49, we may want to add an option to `pystac.Item.from_dict` to skip the `deepcopy` and then also proxy that option to the `pystac.ItemCollection.from_dict` method to improve performance.

2. Create an `ItemCollection` from an iterable of `pystac.Item` instances.

    In this case, we want to provide an option to either clone the items (to protect against mutating the originals) or not clone them (for improved performance). I also realized as I was typing this that we should relax the typing on the `items` parameter from `List[pystac.Item]` to `Iterable[pystac.Item]` since the current code will handle both equally well.

> How about we change Items to be `Union[List[Dict], List[Item]] and handle both cases? from_dict won't do the Item() construction.

I think the only other use-case that this would support would be:

3. Creating an `ItemCollection` from a list of STAC Item-like dictionaries.

Do you think there's a need to be able to do this independent of the first use-case where we are just deserializing a full ItemCollection? If so, then it would definitely be worth changing the `__init__` signature, but if not then maybe we just keep it as-is for simplicity.

pystac/item_collection.py
duckontheweb: Removed the link manipulation in 9b1fed25c879511e91ead274df5d5244520f34b8

pystac/item_collection.py
duckontheweb: Added a `clone_items` parameter to `ItemCollection.__init__` in 9b1fed25c879511e91ead274df5d5244520f34b8

duckontheweb: > I also realized as I was typing this that we should relax the typing on the `items` parameter from `List[pystac.Item]` to `Iterable[pystac.Item]` since the current code will handle both equally well.

Allowed `Iterable[pystac.Item]` for `items` argument instead of more strict `List[pystac.Item]` in 31beb9de97dd7370dab528f7d05b7d53f7003c69

matthewhanson: The first use case is certainly the most common - reading from a file or an http endpoint, and in these cases there's no need to deepcopy or clone anything. I think it will be far more rare to create an Item from a set of pystac.Item objects.

Theres another use case for ItemCollections, which is adding one or more together (e.g., pages of search results), maybe that's just an append to the features.

Looking at your recent changes with the clone_items argument, I think that should get passed on from .from_dict with kwargs, and "from_file" should automatically pass long `clone_items=False` to `.to_dict()`

duckontheweb: > I think it will be far more rare to create an Item from a set of pystac.Item objects.

Yeah, I agree we probably wouldn't use this directly very often, but I'm thinking we would use this when converting to/from `Catalog` and `Collection` instances. In that case we would probably want to be able to pass in an Iterable of `Item` instances and either clone them or not depending on the specific use-case. Something like:

```python
def manipulate_items(item: Item) -> Item:
    # Manipulate links and anything else we need here

item_collection = ItemCollection(items=map(manipulate_items, collection.get_items()))
```
Similarly, for adding `ItemCollection`s together we would already have `Item` instances, so we would probably want to just pass those Instances in directly without going to/from a dictionary.

Maybe the best course here is to change `__init__` to accept an Iterable of either Items or dicts, as you suggested, to cover our bases.

> Looking at your recent changes with the clone_items argument, I think that should get passed to **init** from `from_dict` (through kwargs), and `from_file` should automatically pass long `clone_items=False` to `.to_dict()`

Good catch, I'll update this.

duckontheweb: Can be initialized with an Iterable of either `Item` or `dict` instances as of 702a26f4a9a510d604dd223b490bb27a859ae898

duckontheweb: Ability to add `ItemCollection` instances in 1ead0d573c6f7356e947583aebec5c9c22a0987d.

This will combine unique `Items` into a new `ItemCollection`, so an `Item` that appears in 2 `ItemCollections` (determined by object equality) will only appear once in their sum (see tests or docstring for examples).

duckontheweb: > > Looking at your recent changes with the clone_items argument, I think that should get passed to **init** from `from_dict` (through kwargs), and `from_file` should automatically pass long `clone_items=False` to `.to_dict()`
> 
> Good catch, I'll update this.

I realized that there's no reason to clone Items in `from_dict` since they are being created within that method and therefore have no other references, so I just added `clone_items=False` to that method in 698c54bf94120f4896da8fb679b53a9f6c5dc76d

duckontheweb: > 1. As per the suggestion in [stac-utils/pystac-client#49](https://github.com/stac-utils/pystac-client/issues/49), we may want to add an option to `pystac.Item.from_dict` to skip the `deepcopy` and then also proxy that option to the `pystac.ItemCollection.from_dict` method to improve performance.

If it's okay with you, I may track this as a separate issue since I feel like the scope of this PR has grown enough already.

duckontheweb: @matthewhanson I think we have all of the use-cases that we discussed here in the current iteration. Let me know if you agree and I'll resolve this conversation.

duckontheweb: Based on our conversation today I changed `ItemCollection.__init__` to default to `clone_items=False`, which closes out the issues here. I'll resolve this conversation and re-request a review.

duckontheweb:
@lossyrob This should be ready for your review, too, if you want to take a look.

matthewhanson (approved):

pystac/__init__.py
```diff
@@ -71,7 +72,7 @@
 )
 
 
-def read_file(href: str) -> STACObject:
+def read_file(href: str) -> Union[STACObject, ItemCollection]:
```
lossyrob, line 75: I'm not sure it makes sense to change these top level package read/write methods to include ItemCollection. It breaks user code that may rely on the STACObject type, forcing the type differentiation to happen by the caller. Also the last two parameters of `write_file` make it feel a bit shoe-horned. On the other hand, I can see people getting confused about  why `read_file` wouldn't work on an ItemCollection if they didn't know it wasn't a core stac object.

I think my preference would be to keep these methods working with core STAC types, and force users to treat ItemCollections as a separate concept, which makes more sense with the spec as it currently stands.

lossyrob (commented):
lgtm besides the top level read and write methods, which I commented on. I'm -0.5 for the change to those methods, so if there's someone feels strongly it should be that way I'm fine with things getting merged in as is.

pystac/__init__.py
duckontheweb: Yeah, I agree with all of the points you make here, and I think my preference would be to not change the top-level functions as well. If that ends up being the decision, I will add more clear documentation to the `ItemCollection` class indicating that it is not a `STACObject` and cannot be read using those top-level methods.

@scottyhq I'm curious how much of a priority it is for you to be able to read Item Collections using `pystac.read_file` vs. `ItemCollection.from_file`.

scottyhq: Thanks for all the work on this! Yes, I'm coming at this primarily from someone new to STAC, who is unfamiliar with spec details. The key expectation I have is that after searching a STAC API for data, and saving 'results.json' to work with later, there is an straightforward way to open that file and navigate it. For Python it seems like `pystac_client` takes care of the searching and `pystac` should care of the I/O and navigation of the results. I don't think people should have to understand the concepts of ItemCollections versus Collections, Core vs Not, for this fundamental workflow. 

So if `pystac.read_file` can't handle ItemCollections, and a separate `ItemCollection.from_file()` is the way forward (or pystac_client.read_file()`?), I think that just needs to be clearly documented. 

Also useful (and I think a non-breaking change) would then be for `pystac.read_file` to have error handing that can recognize the json is an ItemCollection and suggest the correct method to open it, rather than the current `KeyError: 'id'` ?

duckontheweb: Thanks @scottyhq, that all makes sense to me. I'll change those top-level functions back to their original signatures and make sure we have clear docs on how to work with ItemCollections. I'm pretty sure the issue with a `KeyError` being raised in `pystac.read_file` is fixed by #402, but I'll add a test to be sure.

@matthewhanson Looking back at the code example in the original issue it seems like the name of the [`ItemSearch.items_as_collection`](https://github.com/stac-utils/pystac-client/blob/9e66bb5b46951988f5be23a09579e512e0fd425f/pystac_client/item_search.py#L402) method might also be misleading. Maybe we should rename that to `items_as_item_collection` so that users don't think they are saving STAC Collection?

duckontheweb: Removed `ItemCollection` handling from top-level functions and added test that `pystac.read_file` raises a `STACTypeError` instead of the `KeyError` in 404bb99390086becce038b35d18a97a8e4d6160b

matthewhanson: @duckontheweb `ItemSearch.items_as_collection` was misleading. I've renamed it already, there's now `get_pages` (get the raw JSON of the pages), `get_item_collection` (gets pages as item collections), `get_items` (iterator through all pages and items), and `get_all_items` (gets all items from all pages and returns as single ItemCollection).   This matches the `get_` syntax used in PySTAC

duckontheweb:
@lossyrob It also just occurred to me that including `ITEMCOLLECTION` in the [`STACObjectTypes` enum](https://github.com/stac-utils/pystac/blob/e2239a73eff9dbf5b41660f19370c1a6c376e387/pystac/stac_object.py#L14-L21) might be misleading as well. It looks like removing that would impact the migration and validation scripts (since it's part of the schema URI mapping), but I think we could work around that. What do you think about removing that from the `Enum`?

lossyrob:
@duckontheweb I agree that's not great, and should have been removed when ItemCollection was dropped. I'm +1 for dropping it from that Enum and using another separate constant in the migration/validation places where needed.

duckontheweb:
> @duckontheweb I agree that's not great, and should have been removed when ItemCollection was dropped. I'm +1 for dropping it from that Enum and using another separate constant in the migration/validation places where needed.

I removed this in 47017df9a8a10874e19ee6363dbebb7282c62333. I ended up just removing any ItemCollection references in the `DefaultSchemaUriMap` and in the migration code rather than try to patch it. My thinking is that we're not claiming to support ItemCollections as STACObjects, and since they won't appear while crawling a catalog this shouldn't negatively affect anything outside of users who are working with ItemCollections directly.

duckontheweb:
@lossyrob @matthewhanson Ready for another review whenever you have time. Thanks!

lossyrob (approved):

matthewhanson:
@duckontheweb Looks good to me to merge!

Issue #331, opened 2021-05-04
Bring code coverage back up to 93-94%


lossyrob:
The changes in #309 will cause the code coverage to dip to 88-89%. This issue is to ensure we bring back coverage to the 93-94% range for the PySTAC 1.0 release.

duckontheweb:
#324 and #325 will probably go a long way towards this given the low coverage of both of those modules. We may want to wait to tackle this issue until those are complete and we have a better sense of the gap.

duckontheweb:
We should also update the coverage badge link the main readme to point to the `main` branch instead of `develop` as part of this issue.

l0b0:
I'd be interested in helping with this. I recently brought the coverage in another project from 97% to 100%, and it was surprising how many issues that uncovered (PR [1](https://github.com/linz/geostore/pull/729), [2](https://github.com/linz/geostore/pull/732), [3](https://github.com/linz/geostore/pull/733), [4](https://github.com/linz/geostore/pull/734)).

duckontheweb:
We would definitely appreciate the help on this! 

One thing we may want to consider is whether we want to include testing of methods like `__repr__` and `__str__` in the coverage stats. I tend to feel like don't add much so I would be fine with excluding them, but I don't really feel strongly about it.

l0b0:
> One thing we may want to consider is whether we want to include testing of methods like `__repr__` and `__str__` in the coverage stats. I tend to feel like don't add much so I would be fine with excluding them, but I don't really feel strongly about it.

I feel a bit like something's wrong if none of the tests end up exercising these - usually that means they aren't needed and should be removed. But you can make a call on that once we get closer to 100%.

In a separate project we've ended up with [the following excludes](https://github.com/linz/geostore/blob/ab5c7ce34dc77bb01fa3e31b562e131c1815998f/pyproject.toml#L9-L13):

- `if TYPE_CHECKING:` since no tests run during type checking
- `if __name__ == "__main__":` since it can be assumed that this is run on the command line, and that the contents of this branch is tiny (typically just `main()`)
- `pragma: no cover` to exclude a few pieces of [test code](https://github.com/linz/geostore/search?q=pragma%3A+no+cover) which don't always run

duckontheweb:
> I feel a bit like something's wrong if none of the tests end up exercising these - usually that means they aren't needed and should be removed. But you can make a call on that once we get closer to 100%.

[...]

PR #442, opened 2021-06-15
Update docs and tests for Version, Label, and File Info extensions

duckontheweb:
## Related Issue(s):

* #331 
* #326

### Version Extension

* Updates docstrings to be consistent with other extensions

### File Info Extension

* Updates implementation to be consistent with v2.0.0 of the extension spec
* Makes `FileExtension` a concrete class on Assets since the spec only defines properties on Assets
* Updates docstrings to be consistent with other extensions
* Adds tests to bring coverage up to 95%

### Label Extension

* Updates docstrings to be consistent with other extensions
* Adds tests to improve coverage to 95%
* Brings code style in line with other extensions
    Added constants for all property names, changed the bodies of many of the property methods to rely on type checking for input validation, and used `get_required` and `get_opt` more consistently.

### Other

* Bumps minimum allowed coverage to 90% since this PR brings it over that threshold.

## PR Checklist:

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,5 +1,5 @@
 [report]
-fail_under = 89
+fail_under = 90
 
 [run]
 source = pystac
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@
 - Package author to `stac-utils`, email to `stac@radiant.earth`, url to this repo ([#409](https://github.com/stac-utils/pystac/pull/409))
 - `StacIO.read_json` passes arbitrary positional and keyword arguments to
   `StacIO.read_text` ([#433](https://github.com/stac-utils/pystac/pull/433))
+- `FileExtension` updated to work with File Info Extension v2.0.0 ([#442](https://github.com/stac-utils/pystac/pull/442))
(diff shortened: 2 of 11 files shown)


lossyrob (approved):
LGTM.

Noticing that the sphinx autodoc output is getting quite messy with the fully qualified types in the signatures. Looked into how to use short names a bit but came up empty.

duckontheweb:
Yeah, I was thinking that, too. I tried setting the [`add_module_names`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-add_module_names) option to `False`, but it looks like that does not have any effect on the `autodoc` extension (and it will end up shortening some other references, which may not be desirable).

Once Sphinx 4.1 hits, it looks like we might be able to take advantage of the [`autodoc-process-bases`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#event-autodoc-process-bases) event to do some post-processing of the outputs. Their milestone for that release was a few days ago, so it may be out soon. I'll open an issue to track this in case someone else has a better solution.

duckontheweb:
> Noticing that the sphinx autodoc output is getting quite messy with the fully qualified types in the signatures. Looked into how to use short names a bit but came up empty.

Tracked in #446

PR #445, opened 2021-06-15
Use environment variable for tmp dir location

### LINKED ISSUE #418: Using cwd for tempfile.TemporaryDirectory makes assumptions that are not always valid
I test in an environment where the source this is read only and the test data is in a [CAS](https://en.wikipedia.org/wiki/Content-addressable_storage).  Making assumptions about the current working directory is trouble.  At least make this setting of the current working directory be dependent on being on windows or in the github CI.  Forcing a user's test environment means I'm forever having the chase and patch out the changes.  Happy to work with anyone for a solution that works for all.

This in tests/utils/__init__.py:

```python
# Use suggestion from https://github.com/python/mypy/issues/5264#issuecomment-399407428
# to solve type errors.
def get_temp_dir() -> "TemporaryDirectory_Type[str]":
    """In the GitHub Actions Windows runner the default TMPDIR directory is on a
    different drive (C:\\) than the code and test data files (D:\\). This was causing
    failures in os.path.relpath on Windows, so we put the temp directories in the
    current working directory instead. There os a "tmp*" line in the .gitignore file
    that ignores these directories."""
    return TemporaryDirectory(dir=os.getcwd())
```

Causes a failure trying to write in read-only storage.

```
suffix = '', prefix = 'tmp'
dir = '/build/work/f4eecd22e8cbe59830e035d545bf935d2afc/google3/runfiles/google3/third_party/py/pystac'

    def mkdtemp(suffix=None, prefix=None, dir=None):
        """User-callable function to create and return a unique temporary
        directory.  The return value is the pathname of the directory.
    
        Arguments are as for mkstemp, except that the 'text' argument is
        not accepted.
    
        The directory is readable, writable, and searchable only by the
        creating user.
    
        Caller is responsible for deleting the directory when done with it.
        """
    
        prefix, suffix, dir, output_type = _sanitize_params(prefix, suffix, dir)
    
        names = _get_candidate_names()
        if output_type is bytes:
            names = map(_os.fsencode, names)
    
        for seq in range(TMP_MAX):
            name = next(names)
            file = _os.path.join(dir, prefix + name + suffix)
            try:
>               _os.mkdir(file, 0o700)
E               PermissionError: [Errno 13] Permission denied: '/build/work/f4eecd22e8cbe59830e035d545bf935d2afc/google3/runfiles/google3/third_party/py/pystac/tmpf_aoq0zi'
```

duckontheweb:
**Related Issue(s):** #

* Closes #418

**Description:**

Uses `TMPDIR` environment variable to change temporary directory location in Windows builds instead of hard-coding the CWD.

**PR Checklist:**

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -54,15 +54,17 @@ jobs:
           path: ~\AppData\Local\pip\Cache
           # Cache based on OS, Python version, and dependency hash
           key: pip-${{ runner.os }}-python${{ matrix.python-version }}-${{ hashFiles('requirements-test.txt') }}
-      
+
       - name: Install dependencies
         run: |
           pip install -r requirements-test.txt
           pip install -e ".[validation]"
-
+      
       - name: Execute test suite
         run: ./scripts/test
         shell: bash
+        env:
+          TMPDIR: "${{ matrix.os == 'windows-latest' && 'D:\\a\\_temp' || '' }}"
       
       - name: Upload All coverage to Codecov
         uses: codecov/codecov-action@v1
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-tmp*
 *.pyc
 *.egg-info
 *.eggs
--- a/tests/data-files/get_examples.py
+++ b/tests/data-files/get_examples.py
@@ -6,7 +6,7 @@ import os
 import argparse
 import json
 from subprocess import call
-from tempfile import TemporaryDirectory
+import tempfile
 from typing import Any, Dict, List, Optional
 from urllib.error import HTTPError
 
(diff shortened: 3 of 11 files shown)


lossyrob (approved):
Nice solve.

schwehr (approved):
LGTM.  

Not critical, but I do prefer the follow module-only import style: 

- There is no diff when you use more things from a module
- Some static analyzers do best with only module level imports
- `find . -name \*.py | xargs grep 'module[.]'` makes it easy to find all lines make use of the module.
- It's easier to follow at scale for me to follow what's being used (e.g. my repo has order 1e9 LOC).  I just have to keep the module in mind

```python

import tempfile
# ...

    with tempfile.TemporaryDirectory() as tmp_dir:
        # ...

duckontheweb:
> Not critical, but I do prefer the follow module-only import style:

That makes sense, I updated this in 2bd05b663bd2c9b669b5a1e6dc91c6ce5bf4141e

Issue #453, opened 2021-06-17
Allow users to bypass deepcopy in from_dict methods


lossyrob:
As described in https://github.com/stac-utils/pystac-client/issues/49, a lot of time is spent in the deepcopy for a large Item dict when deserializing to a pystac.Item. The deepcopy is done to avoid mutating the parameter dict, which I think is good practice, but also wholly unnecessary when you've just created the dict from e.g. reading from a file or deserializing from a HTTP response and are not planning on using the dict any further. In these cases, the deepcopy is only a performance hit with no benefit.

The goal of this issue is to introduce a parameter to `to_dict` methods that allow users to avoid copying the dict. This should be used in PySTAC codepaths such as `from_file` methods where the incoming dict is a temporary data structure.

lossyrob:
Closed by #454

PR #454, opened 2021-06-17
Avoid calling deepcopy in from_dict methods when unnecessary

lossyrob:
**Related Issue(s):** #453 

**Description:**
This PR adds a parameter to the `STACObject.from_dict` method and implementations named `preserve_dict`. If this is True, the dict that is passed in as a parameter will not be mutated during the operation - which is achieved by making a deep copy of dict parameter in the implementing methods. If preserve_dict is False, the incoming dict is mutated and a deepcopy is avoided. The default of this parameter is True, so that users can avoid a situation where their parameters are mutated unsuspectedly. In situations where the dict is a transient encoding of the STAC object, users should use preserve_dict=False to avoid incurring the performance cost of the deepcopy.

`preserve_dict` is set to False for methods that read a STAC object from a file, as there's no need to avoid mutating the dict in those cases.

This should help address some of the performance issues in described in https://github.com/stac-utils/pystac-client/issues/49, as pystac-client will be able to use preserve_dict=False in the [ItemCollection.from_dict method](https://github.com/stac-utils/pystac-client/blob/9e66bb5b46951988f5be23a09579e512e0fd425f/pystac_client/item_collection.py#L59)

**PR Checklist:**

### DIFF
--- a/pystac/stac_object.py
+++ b/pystac/stac_object.py
@@ -495,6 +495,7 @@ class STACObject(ABC):
         href: Optional[str] = None,
         root: Optional["Catalog_Type"] = None,
         migrate: bool = False,
+        preserve_dict: bool = True,
     ) -> "STACObject":
         """Parses this STACObject from the passed in dictionary.
 
@@ -507,6 +508,11 @@ class STACObject(ABC):
                 previously resolved instances of the STAC object.
             migrate: Use True if this dict represents JSON from an older STAC object,
                 so that migrations are run against it.
+            preserve_dict: If False, the dict parameter ``d`` may be modified
+                during this method call. Otherwise the dict is not mutated.
+                Defaults to True, which results results in a deepcopy of the
+                parameter. Set to False when possible to avoid the performance
+                hit of a deepcopy. Defaults to True.
 
         Returns:
             STACObject: The STACObject parsed from this dict.
--- a/pystac/stac_object.py
+++ b/pystac/stac_object.py
@@ -473,7 +473,7 @@ class STACObject(ABC):
             href = make_absolute_href(href)
 
         d = stac_io.read_json(href)
-        o = cls.from_dict(d, href=href, migrate=True)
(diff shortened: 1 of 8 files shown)


pystac/stac_object.py
```diff
@@ -507,6 +508,11 @@ def from_dict(
                 previously resolved instances of the STAC object.
             migrate: Use True if this dict represents JSON from an older STAC object,
                 so that migrations are run against it.
+            preserve_dict: If False, the dict parameter ``d`` may be modified
+                during this method call. Otherwise the dict is not mutated.
+                Defaults to True, which results results in a deepcopy of the
+                parameter. Set to False when possible to avoid the performance
+                hit of a deepcopy. Defaults to True.
```
duckontheweb, line 515:
```suggestion
                hit of a deepcopy.
```

Already state earlier

tests/test_catalog.py
```diff
@@ -85,6 +86,19 @@ def test_create_and_read(self) -> None:
 
             self.assertEqual(len(list(items)), 8)
 
+    def test_from_dict_preserves_dict(self) -> None:
+        catalog_dict = TestCases.test_case_1().to_dict()
+        param_dict = deepcopy(catalog_dict)
+
+        # test that the parameter is preserved
+        _ = Catalog.from_dict(param_dict)
+        self.assertEqual(param_dict, catalog_dict)
```
duckontheweb, line 95: May want to use `assertIs` here (and `assertIsNot` below) in case we update equality per #380. Same for a few other assertions below.

duckontheweb (approved):
👍 , thanks for getting this in. Just 1 minor docstring change and a possible change to some assertions that you can take or leave.

tests/test_catalog.py
lossyrob: @duckontheweb these are dicts, not PySTAC objects, so equals makes sense here right?

duckontheweb: Shoot, totally missed that. Yes, carry on!

PR #452, opened 2021-06-17
Updates to CI and README

duckontheweb:
**Description:**

* Removes Gitter badge since that channel is mostly unused
* Adds `orjson` extra install instructions
* Minor clarifications around versions
* Only run release build for this repo (I was seeing failed builds in my fork when I updated tags)
* Move coverage reporting into separate job
    Since we added the minimum coverage requirement, the `test` job fails before uploading coverage to Codecov.io. When I was working on removing support for pre-0.8 STAC versions I found myself wishing I could use that UI to figure out where the coverage had dropped. This change means the CI workflow will still fail overall, but we will still get a coverage report we can look at if we drop below the minimum threshhold.

**PR Checklist:**

- [x] Code is formatted (run `pre-commit run --all-files`)
- [x] Tests pass (run `scripts/test`)
- [x] This PR maintains or improves overall codebase code coverage.
- [ ] Changes are added to the [CHANGELOG](https://github.com/stac-utils/pystac/blob/develop/CHANGELOG.md). See [the docs](https://pystac.readthedocs.io/en/latest/contributing.html#changelog) for information about adding to the changelog.

PR #466, opened 2021-06-19
Improve coverage checks

l0b0:
**Related Issue(s):** #331

**Description:** Add tests, exclude some lines from coverage count, include tests in coverage count (highlighting unreached test code), and do some minor code changes to make it more testable.

The first PR already contains a bunch of orthogonal improvements. Let me know if you want to split it up into simpler pieces.

**PR Checklist:**

### DIFF
--- a/pystac/extensions/sar.py
+++ b/pystac/extensions/sar.py
@@ -318,7 +318,7 @@ class SarExtension(
             return cast(SarExtension[T], AssetSarExtension(obj))
         else:
             raise pystac.ExtensionTypeError(
-                f"SAR extension does not apply to type {type(obj)}"
+                f"SAR extension does not apply to type '{type(obj).__name__}'"
             )
 
 
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,10 +1,11 @@
 [report]
-fail_under = 95
+fail_under = 93
 exclude_lines =
     if TYPE_CHECKING:
 
 
 [run]
+branch = true
 source =
     pystac
     tests
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,5 +1,8 @@
 [report]
 fail_under = 92
+exclude_lines =
+    if TYPE_CHECKING:
+
 
 [run]
 source = pystac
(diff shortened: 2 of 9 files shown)


duckontheweb (commented):
This looks good to me aside from using `TestCases.get_path` in some places.

I'm not sure what is going on with the Codecov.io reports. When I get the coverage report locally I get 93.17% coverage, but the report that's getting posted here shows 88.52% for this PR. I'll look into that separately and see if I can figure out the cause.

duckontheweb:
> I'm not sure what is going on with the Codecov.io reports. When I get the coverage report locally I get 93.17% coverage, but the report that's getting posted here shows 88.52% for this PR. I'll look into that separately and see if I can figure out the cause.

@l0b0 In #452 I created a separate `coverage` CI job for uploading to Codecov.io so that we could get coverage reports even when coverage drops to help us identify gaps (this happened specifically in 71f07d93ce9cf219d09841d3dd3a55cdc690d066). I did this by directly using the `coverage run...` command in the new job, but I'm realizing now that I probably should have updated `scripts/test` to allow us to override `--fail-under` instead so that we wouldn't have to update that command in 2 places. 

I think the quick fix is to update the `coverage` CI job with the same changes you made to `scripts/test`. If you want to also update `scripts/test` so that we can use it in both cases that's great too 😄 , but I can do that separately.

pystac/extensions/sar.py
```diff
@@ -318,7 +318,7 @@ def ext(cls, obj: T, add_if_missing: bool = False) -> "SarExtension[T]":
             return cast(SarExtension[T], AssetSarExtension(obj))
         else:
             raise pystac.ExtensionTypeError(
-                f"SAR extension does not apply to type {type(obj)}"
+                f"SAR extension does not apply to type '{type(obj).__name__}'"
```
duckontheweb, line 321: It looks like this change could also be made for a number of other extensions as well. Is there a reason to only change this for the `sar` and `view` extensions, or should we update the `ext` method for all extensions in this PR?

l0b0: I can do that, but I'd need to be sure that you're OK with changing all of them so it's not wasted effort. What do you think?

l0b0:
Update: I'm splitting this into several smaller PRs for easier review/merging.

Issue #367, opened 2021-05-27
Use descriptors for Extensions?


TomAugspurger:
I'm writing up an extension, and my wrists hurt from all the typing :) I wonder if we could use [descriptors](https://docs.python.org/3/howto/descriptor.html) to cut down on the boilerplate. For a typical property that comes from the `properties` dict, it'd be something like

```python
class MyExtension:
    description = Property()
``` 

where `Property` is a descriptor like

```python
# untested
class Property:
    def __set_name__(self, owner, name):
        self.name= name

    def __get__(self, obj, objtype=None):
        return getattr(obj, "properties")[self.name)
    def __set__(self, obj, value):
        getattr(obj, "properties")[self.name] = value
```

I haven't really looked to see if this works for many of the extensions. Some downsides:

1. Unclear how this would interact with static typing. But properties are supported by mypy, and descriptors are prosperities, so it might be doable.
2. Makes `.` lookup a bit slower, though on par with `@property` so maybe it's acceptable.

duckontheweb:
> 2\. Makes `.` lookup a bit slower, though on par with `@property` so maybe it's acceptable.

It sounds like Python uses descriptors under the hood for `@property`, so we would only see a performance hit when we replace true attributes with this pattern, which isn't very often.

> I haven't really looked to see if this works for many of the extensions.

I haven't tested this, but it seems like it would be a good pattern in the extensions. Those properties are doing a dynamic lookup on properties of the underlying item anyway, and that seems like a good use-case for descriptors.

I'm definitely +1 on this suggestion in theory, but I think it would be helpful to see an example of how it plays out in one of the simpler classes like `pystac.Asset` or `pystac.Link` to determine whether the change is worth it. In addition to the question around typing we would also want to make sure that we can get docstrings to come through properly for the API docs.

TomAugspurger:
There's a slight inconsistency with how datacube handles setting `None` for a value, and I wanted to check that it's deliberate. For required fields, setting the value to `None` just sets the value in properties. For optional fields, setting the value to `None` deletes the field from properties


```
>>> ext = DatacubeExtension(...)
>>> dim = ext.dimensions["x"]
>>> dim.dim_type = None
>>> "type" in dim.properties
True
>>> dim.description = None
>>> "description" in dim.properties
False
```

If that's deliberate, then I think that the following will work.

```python
class Property:
    def __init__(self, key: str, obj: Union[str, Any]=None, required=False):
        self.key = key
        self.obj = obj
[...]

PR #483, opened 2021-06-28
REF: Use descriptors for properties

TomAugspurger:
**Related Issue(s):** #

#367

**Description:**

A proof of concept for using a descriptor to reduce the boilerplate for defining getters and setters for the attributes on an extension. This cuts down on the lines of code, and should help keep the behavior of getters and setters consistent, since they're using the same implementation.

If we're happy with this, we can expand it to other places in the library. There will be some tradeoff between the complexity of the `__get__` and `__set__` methods and how widely it can be used. Right now it's tailored to the `Datacube` extension.

**PR Checklist:**

### DIFF
--- a/pystac/extensions/datacube.py
+++ b/pystac/extensions/datacube.py
@@ -2,15 +2,14 @@
 
 https://github.com/stac-extensions/datacube
 """
-
 from abc import ABC
+import sys
 from typing import (
     Any,
     Dict,
     Generic,
     List,
     Optional,
-    Protocol,
     Set,
     TypeVar,
     Union,
@@ -24,6 +23,11 @@ from pystac.extensions.base import (
 )
 from pystac.extensions.hooks import ExtensionHooks
 from pystac.utils import get_required, map_opt
+
+if sys.version_info >= (3, 8):
+    from typing import Protocol
+else:
+    from typing_extensions import Protocol
 
 T = TypeVar("T", pystac.Collection, pystac.Item, pystac.Asset)
 
--- a/pystac/extensions/datacube.py
+++ b/pystac/extensions/datacube.py
@@ -4,7 +4,18 @@ https://github.com/stac-extensions/datacube
 """
 
 from abc import ABC
-from typing import Any, Dict, Generic, List, Optional, Set, TypeVar, Union, cast
+from typing import (
+    Any,
+    Dict,
+    Generic,
+    List,
+    Optional,
+    Protocol,
+    Set,
+    TypeVar,
+    Union,
+    cast,
+)
 
 import pystac
 from pystac.extensions.base import (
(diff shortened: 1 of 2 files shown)


pystac/extensions/datacube.py
```diff
@@ -30,33 +41,65 @@
...
+        clear_if_none: Optional[bool] = None,
+    ):
+        self.key = key
+        self.obj = obj
+        self.required = required
+        if clear_if_none is None:
+            clear_if_none = not required
+        self.clear_if_none = clear_if_none
+
+    def __get__(self, instance: HasProperties[U], owner: Any = None) -> Optional[U]:
```
TomAugspurger, line 79: LMK if the type signatures are confusing (I think they are).

But the usage below will hopefully clear things up. `U` is the type of the actual object being returned, like `str` for description, or `List[float]` for extent.

I see now that we can (maybe) tighten this up a bit. We know that if the property is `required` then return type is just `U` rather than `Optional[U]`. I'll see if mypy can handle that.

TomAugspurger:
Fixed that CI failure (apparently I forgot to re-run the test after tweaking it).

duckontheweb (commented):
The Ci failures were for the same reason as in [#264](https://github.com/stac-utils/pystac/pull/264#issuecomment-845958462) (`Protocol` not available until Python 3.8). We should be able to solve this using the same method as in e175aa637fce1e961086a07ed89ffac5267c7827.

TomAugspurger:
Thanks. Fixed in 476b20fedbbd812c58993975cb16287b58421072 hopefully.

duckontheweb (commented):
Thanks for putting this together @TomAugspurger, it's really helpful to see it in practice.

My initial impression is that this seems to add a lot of complexity to how these properties are handled and I'm not sure that the reduction in boilerplate code is worth it, especially we have to implement separate descriptor classes for each extension. If there is a way to make this abstract enough to work across all extensions, however, then I think the trade-off might be worth it. 

The other concern is the typing. Having the attributes typed as something like `_Property[str]` instead of `str` is not ideal, since these types are used in the auto-generated API docs. I did a little experimenting to see if there is a way to have the type annotations reflect the return type of `_Property.__get__`, but came up empty.

TomAugspurger:
> If there is a way to make this abstract enough to work across all extensions, however, then I think the trade-off might be worth it.

Yeah, I think that's doable. Glancing at `eo` for example, it seems to follow the same pattern. I haven't looked outside of the extensions to see if getters and setters are used elsewhere.

> since these types are used in the auto-generated API docs.

Yeah, the docs don't look great.

image

At the very least, the docstring can be set, so that `help(pystac.extensions.datacube.Dimension.description)` looks reasonable, rather than showing the docstring of this `_Property` class.

I'm less sure about the type annotations showing up there. As you say, the `__get__` is what users probably care about, but the type of the `.description` object technically is this `_Property[U]` thing.

(FYI, I consider this pretty low priority. I'll hold on any more experimentation / expanding this to other extensions until we have a consensus on whether it's worth proceeding, which can wait till after 1.0)

lossyrob:
I think the complexity of dealing with docstrings and internal code outweighs the usefulness of removing boilerplate. I'm in favor of closing - who else has thoughts?

duckontheweb:
I'm in favor of closing as well. I made another attempt at getting this to work while working on using dictionaries to back all of the STAC Items and the typing is really the big hindrance. I agree I don't think the trade-off is worth it.

TomAugspurger:
SGTM.

Just a note on the docstring, that's not too hard to fix. We could pass the docstring to the descriptor

```python
description: _Property[Optional[str]] = _Property(DIM_DESC_PROP, help="A detailed description...")
```

and have the class set `__doc__` on the result.

My main concern is around consistency in the behavior of these getters and setters across extensions (and core pystac objects). This could partially be resolved by refactoring properties like

```python
    @extent.setter
     def extent(self, v: Optional[List[Optional[float]]]) -> None:
         if v is None:
             self.properties.pop(DIM_EXTENT_PROP, None)
         else:
             self.properties[DIM_EXTENT_PROP] = v
```

to call a `self._set_optional(DIM_EXTENT_PROP, v)` (name TBD) that's shared across extension classes.

TomAugspurger:
https://adamj.eu/tech/2021/10/18/python-type-hints-how-to-type-a-descriptor/ might be helpful for the typing. It's been a while since I read through it.

PR #486, opened 2021-06-29
Run test script with/without coverage

duckontheweb:
**Related Issue(s):** #

* #466 (specifically [this comment](https://github.com/stac-utils/pystac/pull/466#issuecomment-868982522))

**Description:**

Adds ability to run `scripts/test` with or without coverage (default outside of CI is still to run with coverage). 

Also moves generation of `coverage.xml` out of `scripts/test`. The XML file is mostly used when uploading coverage to Codecov.io. Generating this file would cause all `test` jobs in the matrix to fail if the coverage dropped, which is misleading. Instead, this PR updates the `coverage` CI job to first upload the coverage report to Codecov.io so we can use it for debugging, then generates the report again using the configured `fail-under` value, which will cause `coverage` job to fail if coverage drops. This should make the source of the workflow failure a bit more obvious to developers.

**PR Checklist:**

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -97,11 +97,14 @@ jobs:
           pip install -e ".[validation]"
       
       - name: Execute test suite
-        # --fail-under=0 ensures we publish the coverage regardless of whether it meets
-        # the minimum so we can use Codecov to evaluate gaps
-        run: |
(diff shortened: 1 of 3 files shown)


duckontheweb:
cc: @l0b0

lossyrob (approved):

PR #498, opened 2021-06-29
feat: Print human readable type name in error messages

l0b0:
Based on <https://stackoverflow.com/a/13477954/96588>.

Also add tests to verify the exception being thrown.

**Related Issue(s):** #331

**Description:**

**PR Checklist:**

### DIFF
--- a/pystac/extensions/datacube.py
+++ b/pystac/extensions/datacube.py
@@ -356,7 +356,7 @@ class DatacubeExtension(
             return cast(DatacubeExtension[T], AssetDatacubeExtension(obj))
         else:
             raise pystac.ExtensionTypeError(
-                f"Datacube extension does not apply to type {type(obj)}"
+                f"Datacube extension does not apply to type '{type(obj).__name__}'"
             )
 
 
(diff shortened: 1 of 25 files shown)


pystac/extensions/sar.py
```diff
@@ -318,7 +318,7 @@ def ext(cls, obj: T, add_if_missing: bool = False) -> "SarExtension[T]":
             return cast(SarExtension[T], AssetSarExtension(obj))
         else:
             raise pystac.ExtensionTypeError(
-                f"SAR extension does not apply to type {type(obj)}"
+                f"SAR extension does not apply to type '{type(obj).__name__}'"
```
duckontheweb, line 321: Thanks for splitting this out into a separate PR. I like this change, and I think for consistency we should apply it to the `ext` methods of the other extensions as well.

l0b0: Done! I've also fixed the names of the extensions in some of them, which all had "File" in the error message.

duckontheweb: Thanks, good catch!

duckontheweb (approved):
👍

PR #500, opened 2021-06-29
RFC: Drop support for Python 3.6

### LINKED ISSUE #474: Deprecating support for older Python 3 versions
We currently test against and allow installation on Python >=3.6. If we want to drop support for any Python 3 versions, we should do it prior to the 1.0.0 release. For instance, if we think we may want to take advantage of `dataclasses` we could restrict support to Python >=3.7, or if we wanted to take advantage of `functools.cached_property` we could restrict support to Python >=3.8 so that we don't need to use backports. 

This issue is not meant to advocate for dropping Python 3.6, but rather to open up a discussion prior to the stable release so that we make an informed decision. 

Personally I am +0 on dropping support for Python 3.6 or 3.7. I think there is a good chance that users are still using those versions, and we do not have a burning need to use features introduced in later versions.

duckontheweb:
**Related Issue(s):**

* Closes #474

**Description:**

Drops support for Python 3.6 (both in terms of `python_version` check on install and in CI tests), and adds a note to the README that we follow the NEP-29 guidelines.

**PR Checklist:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,16 +6,17 @@
 [![Gitter](https://badges.gitter.im/SpatioTemporal-Asset-Catalog/python.svg)](https://gitter.im/SpatioTemporal-Asset-Catalog/python?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge)
 [![License](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
 
-PySTAC is a library for working with [SpatialTemporal Asset Catalog](https://stacspec.org) in Python 3.
+PySTAC is a library for working with [SpatioTemporal Asset Catalog](https://stacspec.org) in Python 3.
 
 ## Installation
 
-PySTAC requires Python>=3.7. This project follows the recommendations of
+PySTAC requires Python >= 3.7. This project follows the recommendations of
 [NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html) in deprecating support
 for Python versions. This means that users can expect support for Python 3.7 to be
-removed in next major or minor release after Dec 26, 2021.
+removed from the `main` branch after Dec 26, 2021 and therefore from the next release
+after that date.
 
-*Support for Python>=3.10 should be considered experimental
+*Support for Python >= 3.10 should be considered experimental
 until further notice.*
 
 PySTAC has a single required dependency (`python-dateutil`).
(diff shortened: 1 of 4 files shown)


duckontheweb:
cc: @matthewhanson @gadomski @lossyrob @TomAugspurger

lossyrob:
I believe @schwehr is still stuck on 3.6. I'd rather not remove 3.6 if that will be really disruptive to his usage of PySTAC as I believe he is a big user and has been a great contributor to the project. If that's not the case, I'm +1.

schwehr:
Am in the process of going py 3.6 to 3.7.  Go for it.  +1 for dropping 3.6.  Thanks for asking me!

README.md
```diff
@@ -10,8 +10,13 @@ PySTAC is a library for working with [SpatialTemporal Asset Catalog](https://sta
 
 ## Installation
 
-PySTAC requires Python>=3.6. Support for Python>=3.10 should be considered experimental
-until further notice.
+PySTAC requires Python>=3.7. This project follows the recommendations of
+[NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html) in deprecating support
+for Python versions. This means that users can expect support for Python 3.7 to be
+removed in next major or minor release after Dec 26, 2021.
```
schwehr, line 16: Maybe better to say a planned date to remove from the main branch.

README.md
```diff
@@ -10,8 +10,13 @@ PySTAC is a library for working with [SpatialTemporal Asset Catalog](https://sta
 
 ## Installation
 
-PySTAC requires Python>=3.6. Support for Python>=3.10 should be considered experimental
-until further notice.
+PySTAC requires Python>=3.7. This project follows the recommendations of
```
schwehr, line 13: Might be more readable with spaces: `Python >= 3.7`

duckontheweb: Yeah, good call.

README.md
duckontheweb: Yeah, that makes sense. Our current branching strategy doesn't have a way of doing patch releases without including features, so it would really be included in the next release of any kind.

duckontheweb: Fixed in c6bf9510e82aebcf800ef73769856dbde34f12ba

README.md
duckontheweb: Fixed in c6bf9510e82aebcf800ef73769856dbde34f12ba

README.md
schwehr: And I directly use main:

 I do a [copybara](https://github.com/google/copybara) sync automatically every week.   That's basically git pull + patches + remove anything non-functional except the license file.

schwehr (approved):

lossyrob (approved):

matthewhanson (approved):

PR #503, opened 2021-06-30
Increase test coverage

l0b0:
**Related Issue(s):** #331

**Description:**

**PR Checklist:**

### DIFF
--- a/tests/test_version.py
+++ b/tests/test_version.py
@@ -1,5 +1,6 @@
 import os
 import unittest
+from unittest.mock import patch
 
 import pystac
 from tests.utils import TestCases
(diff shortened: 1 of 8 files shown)


tests/test_version.py
```diff
@@ -1,28 +1,20 @@
...
-        pystac.set_stac_version(None)
+        pystac.version.STACVersion._override_version = None
 
     def test_override_stac_version_with_environ(self) -> None:
-
         override_version = "1.0.0-gamma.2"
-        os.environ["PYSTAC_STAC_VERSION_OVERRIDE"] = override_version
-        cat = TestCases.test_case_1()
-        d = cat.to_dict()
+        with patch.dict(os.environ, {"PYSTAC_STAC_VERSION_OVERRIDE": override_version}):
```
duckontheweb, line 15: I've generally preferred using this as a function decorator instead of a context manager in the past because it ensures that the variable gets set for the entire duration of that test and simplifies the function code.

The context manager works just fine in this case, but I just wanted to point that out as an alternative in case we want to establish it as a pattern.

tests/test_writing.py
```diff
@@ -35,18 +35,6 @@ def validate_file(self, path: str, object_type: str) -> List[Any]:
...
-                if not is_absolute_href(asset.href):
-                    is_valid = not is_absolute_href(asset.href)
-                    if not is_valid:
-                        # If the item href and asset href don't share
-                        # the same root, the asset href must be absolute
-                        rel_href = make_relative_href(asset.href, item_href)
-                        self.assertEqual(asset.href, rel_href)
-                    else:
-                        self.assertTrue(is_valid)
-
```
duckontheweb, line 49: Looking back through the commit history it seems like this test did have a purpose originally, but that it's been degraded over time. @lossyrob Do you know what the original intent was?

duckontheweb (approved):
A couple of comments that are more like questions, but I'm happy merging this as-is.

tests/test_version.py
l0b0: That's a good way of thinking about it. I'll just explain my reasoning since it's quite possible we have some common ground here. Basically, I try to minimize the size of basically any scope, including `with`, `@decorators`, `try…except` and the like. Each scope is another thing to keep in mind when reading each statement, so I feel like this minimizes the amount of context the reader has to keep in mind when reading.

duckontheweb: Yeah, that all makes good sense to me! Let's keep this as-is.

tests/test_writing.py
lossyrob: Yep, it used to be that when you saved a catalog as SELF_CONTAINED and/or RELATIVE_PUBLISHED it would also modify the asset hrefs to be absolute/relative (I forget which). Looks like some of the catalog type information got deleted here, probably as part of the change to remove LinkType. So this looks like code that was refactored to pass with refactors, but ended up meaningless.

Issue #479, opened 2021-06-26
Support for 'processing' extension in pystac.Provider


scottyhq:
Copying some discussion from gitter from @matthewhanson and @gadomski who recommended using the processing extension to add software lineage information to a STAC 

As noted https://github.com/stac-extensions/processing#item-properties-and-collection-provider-fields these extension fields are supposed to go under "providers"

But currently `pystac.Provider` only accepts "name", "roles", and "url". Aside from formally implementing support for the processing extension are there any workarounds? For example if I just want to add arbitrary fields to create the example in the extension repo: 
https://github.com/stac-extensions/processing/blob/b0dc14b3a2d0de77b0b789eb6a5ce2de6dad7771/examples/collection.json#L11-L25
```
  "providers": [
    {
      "name": "European Union/ESA/Copernicus",
      "roles": [
        "producer",
        "licensor"
      ],
      "url": "https://sentinel.esa.int/web/sentinel/user-guides/sentinel-2-msi",
      "processing:lineage": "Generation of Level-1C User Product",
      "processing:level": "L1C",
      "processing:facility": "Copernicus S2 Processing and Archiving Facility",
      "processing:software": {
        "IPF-S2L1C": "02.06"
      }
    },
```

gadomski:
More generally, if extensions can add attributes to a `Provider` (or other components of STAC objects), then every component object needs to have an `extra_fields` or `properties` dictionary. `Link` supports extra attributes, but (e.g.) `Extent` and `Provider` don't.

As an aside, is there a rule for when an extension attribute should be specified at the top-level or inside of a `properties` dictionary? Does it depend on the object being extended?

duckontheweb:
> More generally, if extensions can add attributes to a `Provider` (or other components of STAC objects), then every component object needs to have an `extra_fields` or `properties` dictionary. `Link` supports extra attributes, but (e.g.) `Extent` and `Provider` don't.

Agreed.



> is there a rule for when an extension attribute should be specified at the top-level or inside of a `properties` dictionary? Does it depend on the object being extended?

In terms of the JSON object itself, the [General Conventions section of the Extension docs](https://github.com/radiantearth/stac-spec/tree/master/extensions#general-conventions) provides some guidance on this. I think the idea is that since an Item is a GeoJSON Feature (and therefore can have a `properties` object) extra fields should go there, but for all other STAC objects they should go at the top level.

[...]

Issue #321, opened 2021-05-03
Create type annotations for attributes 


lossyrob:
a la https://www.python.org/dev/peps/pep-0526/#class-and-instance-variable-annotations

See discussion in comment thread here: https://github.com/stac-utils/pystac/pull/309#discussion_r624389895

PR #510, opened 2021-07-02
`properties` -> `extra_fields` for Asset and Link

duckontheweb:
**Related Issue(s):**

* #479 (see [this comment](https://github.com/stac-utils/pystac/issues/479#issuecomment-869968926))
* #321

**Description:**

Adds type annotations for `Link` and `Asset` attributes and uses these in API documentation. Changes `properties` to `extra_fields` for `Link` and `Asset` classes to use more consistent naming across PySTAC classes. 

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@
 
 ### Changed
 
+- Renamed `Asset.properties` -> `Asset.extra_fields` and `Link.properties` ->
+  `Link.extra_fields` for consistency with other STAC objects
+  ([#510](https://github.com/stac-utils/pystac/pull/510))
+
 ### Fixed
 
 - Bug in `pystac.serialization.identify_stac_object_type` where invalid objects with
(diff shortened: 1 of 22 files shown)


gadomski (approved):
LGTM. The only thing that was initially surprising was the `properties` attribute on all extensions, but it made sense after reading for a bit (they aren't extra fields, they really are the extension's data).

duckontheweb:
> LGTM. The only thing that was initially surprising was the `properties` attribute on all extensions, but it made sense after reading for a bit (they aren't extra fields, they really are the extension's data).

Yeah, I thought that was too much to take on in this PR, but I agree that is confusing terminology as well. It also interferes with the naming convention in the Label Extension, since that defines a `"label: properties"` attribute. I can open a separate issue to discuss whether we want to change that.

lossyrob (approved):

Issue #505, opened 2021-06-30
Can we remove `__str__` methods on enums?


l0b0:
A bunch of enums in the code base include this code:

```
def __str__(self) -> str:
    return str(self.value)
```

I'm not sure if these are useful, since we can get the same with `Foo.BAR.value`.

duckontheweb:
At one point I thought I had tried this and it caused a number of test failures, but that doesn't seem to be the case anymore (the only test failures are the ones that call `str(SomeEnum.SOME_MEMBER)`). I assumed originally that these were needed for JSON serialization, but that may not be true.

l0b0:
Looking at the use cases for these methods, it appears there are a bunch of tests like `self.assertEqual(str(VersionRelType.LATEST), "latest-version")`. After changing these to `self.assertEqual(VersionRelType.LATEST.value, "latest-version")` we probably can get rid of them, since they are just testing a static value and not any functionality.

PR #514, opened 2021-07-06
chore: Remove unused `__str__` methods

l0b0:
They were only accessed by tests.

**Related Issue(s):** #505

**Description:**

**PR Checklist:**

### DIFF
--- a/tests/extensions/test_file.py
+++ b/tests/extensions/test_file.py
@@ -9,8 +9,8 @@ from pystac.extensions.file import FileExtension, ByteOrder, MappingObject
 
 class ByteOrderTest(unittest.TestCase):
     def test_to_str(self) -> None:
-        self.assertEqual(str(ByteOrder.LITTLE_ENDIAN), "little-endian")
-        self.assertEqual(str(ByteOrder.BIG_ENDIAN), "big-endian")
+        self.assertEqual(ByteOrder.LITTLE_ENDIAN.value, "little-endian")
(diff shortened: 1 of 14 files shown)


tests/extensions/test_file.py
```diff
@@ -9,8 +9,8 @@
 
 class ByteOrderTest(unittest.TestCase):
     def test_to_str(self) -> None:
-        self.assertEqual(str(ByteOrder.LITTLE_ENDIAN), "little-endian")
-        self.assertEqual(str(ByteOrder.BIG_ENDIAN), "big-endian")
+        self.assertEqual(ByteOrder.LITTLE_ENDIAN.value, "little-endian")
+        self.assertEqual(ByteOrder.BIG_ENDIAN.value, "big-endian")
```
duckontheweb, line 13: I think the only purpose for these tests was to provide test coverage for the `__str__` methods. Does coverage drop if we exclude these? It doesn't seem that useful to me to just test an Enum's value.

l0b0: According to the report above the coverage actually went *up* after this change :)

duckontheweb: Sorry, should have been more clear in the comment. I realize the current changes have increased coverage. I'm wondering if we can remove these tests entirely. It's probably not crucial for this PR, but something to consider since they don't seem to add much value.

duckontheweb (approved):

PR #519, opened 2021-07-06
chore: Remove unused locals

l0b0:
**Related Issue(s):** #331

**Description:**

**PR Checklist:**

### DIFF
--- a/pystac/catalog.py
+++ b/pystac/catalog.py
@@ -626,7 +626,6 @@ class Catalog(STACObject):
         template: str,
         defaults: Optional[Dict[str, Any]] = None,
         parent_ids: Optional[List[str]] = None,
-        **kwargs: Any,
     ) -> List["Catalog"]:
         """Walks through the catalog and generates subcatalogs
         for items based on the template string.
--- a/pystac/extensions/base.py
+++ b/pystac/extensions/base.py
@@ -57,7 +57,7 @@ class PropertiesExtension(ABC):
     ``additional_read_properties`` will take precedence.
     """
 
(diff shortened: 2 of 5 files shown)


pystac/stac_io.py
```diff
@@ -101,7 +101,7 @@ def json_loads(self, txt: str, *args: Any, **kwargs: Any) -> Dict[str, Any]:
             result = json.loads(txt)
         return result
 
-    def json_dumps(self, json_dict: Dict[str, Any], *args: Any, **kwargs: Any) -> str:
```
lossyrob, line 104: I believe these are intentional to allow for overriding classes to user arbitrary arguments? cc @matthewhanson

duckontheweb: That's right. We have these here intentionally in case someone wants to overwrite `json_dumps` with an implementation that needs to pass other arguments to the JSON parser. The same goes for `save_json`, since it calls `json_dumps` using those additional arguments.

matthewhanson: Yes, these are overridden in Client.

The **kwargs in `generate_subcatalogs` is not used in a subclass that I know of.

l0b0: Why not just pass the arbitrary arguments by default? Then the user won't have to override the implementation if they're using the built-in `json` module or `orjson`.

l0b0: I've moved the relevant commit to https://github.com/stac-utils/pystac/pull/542.

duckontheweb (approved):
Looks good, thanks for all the clean-up work you are doing!

PR #529, opened 2021-07-07
chore: Remove unused constructor

l0b0:
**Related Issue(s):** #

**Description:**

**PR Checklist:**

### DIFF
--- a/pystac/extensions/projection.py
+++ b/pystac/extensions/projection.py
@@ -46,9 +46,6 @@ class ProjectionExtension(
        >>> proj_ext = ProjectionExtension.ext(item)
     """
 
-    def __init__(self, item: pystac.Item) -> None:
-        self.item = item
-
     def apply(
         self,
         epsg: Optional[int],


pystac/extensions/projection.py
```diff
@@ -286,7 +286,8 @@ class ItemProjectionExtension(ProjectionExtension[pystac.Item]):
     """The :class:`~pystac.Item` properties, including extension properties."""
 
     def __init__(self, item: pystac.Item):
-        self.item = item
+        super().__init__(item)
+
```
duckontheweb, line 290: Actually, I think we should remove the `ProjectionExtension.__init__` method entirely so that it is consistent with the other extension implementations that work on Assets and Items. `ItemProjectionExtension` is the only sub-class that accepts an `item` argument to its init (`AssetProjectionExtension` has a different signature), so it seems like this may have been a mistake.

cc: @lossyrob

lossyrob: +1

l0b0: Cool, can do!

matthewhanson (approved):

duckontheweb (approved):

Issue #481, opened 2021-06-26
Move CommonMetadata to use Extensions API


lossyrob:
CommonMetadata can be set on Assets or Items, and can have Collection summaries. This makes it most similar to an extension, even though it's not technically an extension.

CommonMetadata currently uses the old extension API when dealing with Assets. We should modify the CommonMetadata API so that it fits with the new extension API.

duckontheweb:
Looking at the [Common Metadata Basics](https://github.com/radiantearth/stac-spec/blob/master/item-spec/common-metadata.md#basics) spec and it's a little unclear to me whether the `title` and `description` fields are meant to be applied to Assets as well since the spec refers only to Items (e.g. "Descriptive fields to give a basic overview of a STAC Item."). The [current `CommonMetadata` implementation](https://github.com/stac-utils/pystac/blob/a4fe4cd5f2dcde2c997c6138b2c75a963914b9d7/pystac/item.py#L40-L65) for those fields treats them differently than the remaining Common Metadata fields in that the can only be set on an `Item` and there is not optional `asset` argument to the setter.

Should we allow `title` and `description` to be set on `Asset` as well as `Item` instances?

lossyrob:
Title and description are already on an Asset, so we coudl allow setting without a conflict. It's a little weird that you'd be able to set the title of an asset via both

```
asset.title = "foo"
CommonMetadata(asset).title = "foo"
```

but :shrug:

duckontheweb:
> Title and description are already on an Asset, so we coudl allow setting without a conflict. It's a little weird that you'd be able to set the title of an asset via both
> 
> ```
> asset.title = "foo"
> CommonMetadata(asset).title = "foo"
> ```
> 
> but 🤷

Yeah, we don't include `datetime` in the `CommonMetadata` properties even though it is part of the Common Metadata spec, maybe for the same reason.

My vote would be to include all fields from the [Common Metadata spec](https://github.com/radiantearth/stac-spec/blob/master/item-spec/common-metadata.md) as `CommonMetadata` properties even if they are already defined as instance attributes in the object itself.

duckontheweb:
Closed by #563

Issue #315, opened 2021-04-28
Implement RequiredPropertyMissing exception where appropriate


lossyrob:
#309 implements an exception that raises when a required property is missing, using the `repr` of an object in the message to give users as much detail as possible to identify where the error is. This issue will apply after that PR is merged, and the goal of the issue is to implement throwing this exception (normally via the `get_required` utility method) where appropriate and ensuring that object `repr`s give details for any object that is throwing this exception.

duckontheweb:
@lossyrob My sense is that this would represent a breaking change since many of these cases currently return `None`, but would raise and exception after applying this change. If that's the case it might be good to get this in under an RC1 release, but if it's not breaking we could probably release without it.

PR #532, opened 2021-07-07
Release v1.0.0-rc.3

duckontheweb:
**Description:**

Update version and CHANGELOG for v1.0.0-rc.3 release.

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@
 
 ### Added
 
+### Removed
+
+### Changed
+
+### Deprecated
+
+## [v1.0.0-rc.3]
+
+### Added
+
 - (Experimental) support for Python 3.10 ([#473](https://github.com/stac-utils/pystac/pull/473))
 -  `LabelTask` enum in `pystac.extensions.label` with recommended values for
   `"label:tasks"` field ([#484](https://github.com/stac-utils/pystac/pull/484))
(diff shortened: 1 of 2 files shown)


l0b0:
Several of us at Toitū Te Whenua Land Information New Zealand are very keen for the 1.0 release. Are there any things in particular which you would like feedback on?

lossyrob (approved):

duckontheweb:
> Several of us at Toitū Te Whenua Land Information New Zealand are very keen for the 1.0 release. Are there any things in particular which you would like feedback on?

At this point we are really trying to make sure we get any potentially breaking changes in prior to a 1.0 release. As far as I can tell, the only remaining issues that fit that description are #481 (since it will change the API for working with `CommonMetadata`) and #315 (which will result in exceptions in cases that now return `None`). #380 also has the potential to break user code by changing the definition of equality, but I'm inclined to address that post-1.0 since I think it may take some time to flesh out and I'm not sure how many downstream applications are actually relying on a certain definition of equality.

I'm currently working on #481 and plan on tackling #315 after that unless someone else gets to it first and then I would anticipate putting out a 1.0 release after that.

@lossyrob Do you have any other thoughts on this?

lossyrob:
I think we'd be ready for a full release after the changes you mentioned, and it will be great to fully unblock a lot of downstream work for STAC 1.0

There is also the work of updating the quickstart, tutorials, and any out-of-date documentation that's left to be done. Since we can render the v1.0 release docs off of the 1.0 release branch and not a tag, those doc updates could happen after the tag and pypi release are made.

PR #538, opened 2021-07-07
refactor: Avoid implicit reexport in tests

l0b0:
**Related Issue(s):** #332 

**Description:** First step towards `strict = True` mypy setting.

**PR Checklist:**

### DIFF
--- a/tests/data-files/change_stac_version.py
+++ b/tests/data-files/change_stac_version.py
@@ -7,9 +7,10 @@ import re
 import argparse
 import json
 
-import pystac
+from pystac import read_dict
+from pystac.version import get_stac_version
 
-TARGET_VERSION = pystac.get_stac_version()
+TARGET_VERSION = get_stac_version()
 
 
 def migrate(path: str) -> None:
@@ -28,7 +29,7 @@ def migrate(path: str) -> None:
                     path, cur_ver, TARGET_VERSION
                 )
             )
-            obj = pystac.read_dict(stac_json, href=path)
+            obj = read_dict(stac_json, href=path)
             migrated = obj.to_dict(include_self_link=False)
             with open(path, "w") as f:
                 json.dump(migrated, f, indent=2)
--- a/tests/data-files/get_examples.py
+++ b/tests/data-files/get_examples.py
@@ -10,8 +10,10 @@ import tempfile
 from typing import Any, Dict, List, Optional
 from urllib.error import HTTPError
 
-import pystac
-from pystac.serialization import identify_stac_object
+from pystac.version import get_stac_version
+from pystac.stac_object import STACObjectType
+from pystac.stac_io import StacIO
+from pystac.serialization.identify import identify_stac_object
 
 
 def remove_bad_collection(js: Dict[str, Any]) -> Dict[str, Any]:
@@ -23,7 +25,7 @@ def remove_bad_collection(js: Dict[str, Any]) -> Dict[str, Any]:
             if rel is not None and rel == "collection":
                 href: str = link["href"]
                 try:
-                    json.loads(pystac.StacIO.default().read_text(href))
+                    json.loads(StacIO.default().read_text(href))
                     filtered_links.append(link)
                 except (HTTPError, FileNotFoundError, json.decoder.JSONDecodeError):
                     print("===REMOVING UNREADABLE COLLECTION AT {}".format(href))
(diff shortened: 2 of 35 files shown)


duckontheweb (approved):
This looks good to me, but I know there has been some discussion about how to do these imports in other PRs, so I want to make sure I understand the benefits.

It seems like this changes what is implicitly re-exported rather than removing any implicit re-exports (which would be impossible). For instance, we can no longer do this:

```python
from tests.extensions.test_custom import pystac
```

...but we can now do this:

```python
from tests.extensions.test_custom import Link, Asset, RangeSummary, ...
```

Is the ultimate goal here to help eliminate circular imports or is there another aim? Also, if there are some documented best practices on this that I've missed feel free to just link to those and I can read up on it myself; this is mostly for my own understanding.

Thanks!

duckontheweb:
I guess maybe my real question is why `mypy` cares about this in `strict ` mode...

l0b0:
> This looks good to me, but I know there has been some discussion about how to do these imports in other PRs, so I want to make sure I understand the benefits.
> 
> It seems like this changes what is implicitly re-exported rather than removing any implicit re-exports (which would be impossible). For instance, we can no longer do this:
> 
> ```python
> from tests.extensions.test_custom import pystac
> ```
> 
> ...but we can now do this:
> 
> ```python
> from tests.extensions.test_custom import Link, Asset, RangeSummary, ...
> ```

That's right; basically, importing from the place where something is defined rather than indirectly via `__init__.py` all over the place.
 
> Is the ultimate goal here to help eliminate circular imports or is there another aim? Also, if there are some documented best practices on this that I've missed feel free to just link to those and I can read up on it myself; this is mostly for my own understanding.

When I started the work on this I simply assumed that mypy was doing the right thing, since this type of import is what I'm used to from elsewhere. This assumption was strengthened when I discovered that using direct imports caused a bunch of circular imports to break things. It was further strengthened when I tried running `isort` on the "main" branch and that *also* caused circular imports. Not being able to reorder imports seems like we're just on the verge of breaking things at all times, and any new import could break things in unexpected ways.

On the other hand, the [documentation](https://mypy.readthedocs.io/en/latest/command_line.html#cmdoption-mypy-no-implicit-reexport) mentions how including values in `__all__` makes them explicitly reexported. I doubt that using `__all__` will allow us to use `isort` (that is, I don't expect it to *prevent* circular imports), but I can't be sure without trying. I'll try adding `__all__` to some `__init__.py` file(s) and see if that allows us to use `isort` and `mypy --strict` without circular imports.

duckontheweb:
> emptying out the `__init__.py` files and splitting things up to avoid any remaining circular imports is the way to go.

Yeah, I agree that this is definitely the way to go for both the tests and the code base in general. My only hesitation is that it's nice for users to be able to do something like `from pystac import Item` rather than `from pystac.item import Item` so they don't need to remember which module everything is in. I'm guessing that will come up as part of #543, though, so I don't see any reason that we can't merge this PR.

l0b0:
> > emptying out the `__init__.py` files and splitting things up to avoid any remaining circular imports is the way to go.
> 
> Yeah, I agree that this is definitely the way to go for both the tests and the code base in general. My only hesitation is that it's nice for users to be able to do something like `from pystac import Item` rather than `from pystac.item import Item` so they don't need to remember which module everything is in. I'm guessing that will come up as part of #543, though, so I don't see any reason that we can't merge this PR.

I forgot to mention, I think there is a workaround to get the best of both worlds: avoid reexports inside of pystac itself, but import everything we expect the users to use into the root `__init__.py`, similar to how it looks now. That way the root `__init__.py` is only used by pystac users, not within pystac itself, but users still have the same usability as now. I can't think of a reason why this won't work off the top of my head; how about you?

duckontheweb:
> a

This would definitely be the ideal situation; it would be nice not to change the way users have to import from `pystac`, if possible. To be honest, I'm not really sure whether it's possible or not (I'm constantly surprised by the Python import system 🙂 ), but my instinct is that resolving the circular imports that come up as part of this PR is going to mean that we're able to do it.

l0b0:
I've tried to pull apart some of the circular imports, but without success so far. `pystac.catalog`, `pystac.collection` and `pystac.item` all seem to use each other, and that's when excluding the `if TYPE_CHECKING` imports (which in some cases should be removed since the relevant module class is already in scope). It looks like we'll have to pull some core things apart to remove these circular imports.

PR #542, opened 2021-07-08
chore: Remove unused parameters recursively

l0b0:
Since `orjson.dumps` and `json.dumps` don't support the same options, we
probably don't want to forward arbitrary arguments to them in any case.

**Related Issue(s):** #331

**Description:**

**PR Checklist:**

### DIFF
--- a/pystac/stac_io.py
+++ b/pystac/stac_io.py
@@ -98,7 +98,7 @@ class StacIO(ABC):
         if orjson is not None:
             result = orjson.loads(txt)
         else:
-            result = json.loads(txt)
+            result = json.loads(txt, *args, **kwargs)
         return result
 
     def json_dumps(self, json_dict: Dict[str, Any], *args: Any, **kwargs: Any) -> str:
@@ -115,9 +115,11 @@ class StacIO(ABC):
             json_dict : The dictionary to serialize
         """
         if orjson is not None:
-            return orjson.dumps(json_dict, option=orjson.OPT_INDENT_2).decode("utf-8")
+            return orjson.dumps(json_dict, option=orjson.OPT_INDENT_2, **kwargs).decode(
+                "utf-8"
+            )
         else:
-            return json.dumps(json_dict, indent=2)
+            return json.dumps(json_dict, *args, indent=2, **kwargs)
 
     def stac_object_from_dict(
         self,


duckontheweb (commented):
Thanks for moving this into a separate issue, I think it warrants its own discussion.

The reason we had these in the first place is to enable developers to customize `json_dumps` and then be able to pass arguments through `save_json` without also having to overwrite that method. I created a branch off of this one with [a failing test](https://github.com/duckontheweb/pystac/blob/6bf6e3434cf73369863e22ab182a4cbea33d51d6/tests/test_stac_io.py#L106-L132) to illustrate an example usage. That failing test is based on using `ujson`, but another example might be a case where someone wants to customize `json_dumps` to use the `indent` argument in certain cases. 

I think an argument could be made that it is not that much extra effort to also overwrite `save_json` in this case and that maybe it's actually preferable to explicitly overwrite any method signatures that need customizing rather than rely on automatically passing arbitrary arguments. We're passing these arbitrary arguments from `StacIO.save_json` to `StacIO.json_dumps`, but not to `StacIO.write_text`, which is also called by that method. If someone adds arguments in a custom `StacIO.write_text` implementation they would also need to overwrite `StacIO.save_json` to include those arguments. Why is this required for `write_text` but not `json_dumps`?

It would be good to get comments from maintainers of other downstream libraries to see if/how they use this feature.

@matthewhanson @gadomski Are there examples of where this kind of customization is used?

@lossyrob Are there other examples of usages that would require these arbitrary `args/kwargs`?

gadomski:
> Are there examples of where this kind of customization is used?

Not in core stactools. In the main method override, `args` and `kwargs` are not used: https://github.com/stac-utils/stactools/blob/e90826847e5ebaff2884e2e2f881b29bc3491efd/src/stactools/core/io/__init__.py#L23-L30

l0b0:
> The reason we had these in the first place is to enable developers to customize `json_dumps` and then be able to pass arguments through `save_json` without also having to overwrite that method. I created a branch off of this one with [a failing test](https://github.com/duckontheweb/pystac/blob/6bf6e3434cf73369863e22ab182a4cbea33d51d6/tests/test_stac_io.py#L106-L132) to illustrate an example usage. That failing test is based on using `ujson`, but another example might be a case where someone wants to customize `json_dumps` to use the `indent` argument in certain cases.

Why not just always forward `*args` and `**kwargs` to the JSON library? That way they are not unused, and as a bonus, someone who needs any of these extra arguments don't need to write their own copy of the method just to pass them through. Like this (without the docstring):

```python
    def json_dumps(self, json_dict: Dict[str, Any], *args: Any, **kwargs: Any) -> str:
        if orjson is not None:
            return orjson.dumps(json_dict, *args, option=orjson.OPT_INDENT_2, **kwargs).decode("utf-8")
        else:
            return json.dumps(json_dict, *args, indent=2, **kwargs)
```

duckontheweb:
> > The reason we had these in the first place is to enable developers to customize `json_dumps` and then be able to pass arguments through `save_json` without also having to overwrite that method. I created a branch off of this one with [a failing test](https://github.com/duckontheweb/pystac/blob/6bf6e3434cf73369863e22ab182a4cbea33d51d6/tests/test_stac_io.py#L106-L132) to illustrate an example usage. That failing test is based on using `ujson`, but another example might be a case where someone wants to customize `json_dumps` to use the `indent` argument in certain cases.
> 
> Why not just always forward `*args` and `**kwargs` to the JSON library? That way they are not unused, and as a bonus, someone who needs any of these extra arguments don't need to write their own copy of the method just to pass them through. Like this (without the docstring):
> 
> ```python
>     def json_dumps(self, json_dict: Dict[str, Any], *args: Any, **kwargs: Any) -> str:
>         if orjson is not None:
>             return orjson.dumps(json_dict, *args, option=orjson.OPT_INDENT_2, **kwargs).decode("utf-8")
>         else:
>             return json.dumps(json_dict, *args, indent=2, **kwargs)
> ```

Yes, that totally makes sense! Not sure why we didn't do that in the first place...

l0b0:
@duckontheweb I've changed the code to take the optional arguments where applicable. For example, `orjson.loads` only ever takes a single argument, so I haven't changed that call.

duckontheweb (approved):
Looks great, thanks for helping us iron this out.

PR #543, opened 2021-07-08
refactor: Avoid implicit re-exports in main module

l0b0:
**Related Issue(s):** #332

**Description:** Work in progress until I resolve circular imports

**PR Checklist:**

### DIFF
--- a/pystac/collection.py
+++ b/pystac/collection.py
@@ -18,10 +18,11 @@ from typing import (
 import dateutil.parser
 from dateutil import tz
 
-import pystac
-from pystac import STACObjectType, CatalogType
+from pystac.rel_type import RelType
+from pystac.stac_io import StacIO
+from pystac.stac_object import STACObjectType
 from pystac.asset import Asset
-from pystac.catalog import Catalog
+from pystac.catalog import Catalog, CatalogType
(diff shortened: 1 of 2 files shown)


duckontheweb:
@l0b0 One thing to consider here is that clearing out the top-level `__init__.py` file would represent a breaking change since it would change the way all of our classes, etc. need to be imported. If that is the ultimate plan, we should probably get this in before a 1.0.0 release.

cc: @lossyrob

lossyrob:
As a I mentioned in the other PR, I'm -1 on clearing out the top level of types.

I'm also -1 of making code quality refactors that modify the API at this point, as I think it will slow down a release, which is of higher value IMO.

duckontheweb:
> As a I mentioned in the other PR, I'm -1 on clearing out the top level of types.
> 
> I'm also -1 of making code quality refactors that modify the API at this point, as I think it will slow down a release, which is of higher value IMO.

I am in agreement, and based on [this comment](https://github.com/stac-utils/pystac/pull/538#issuecomment-880969428) I believe @l0b0  is as well. If that is the case, and we are not going to make major changes to how users import classes and function from the top-level of PySTAC, then I don't think we need to rush to get this into the 1.0 release.

Issue #525, opened 2021-07-07
EO extension can't be added to Collections


lossyrob:
Currently the typing of the EOExtension includes `ExtensionManagementMixin[pystac.Item]`, but should be  `ExtensionManagementMixin[pystac.Collection, pystac.Item]`

Collections should list a schema URI in it's `stac_extensions` field if that Collection has summaries or assets that implement an extension (see [these docs](https://github.com/radiantearth/stac-spec/tree/master/extensions#extension-ids-in-stac_extensions))

Solving this issue should include checking for similar issues in the other extensions.

jisantuc:
I'll grab this -- could someone with more repo powers than I have assign me?

duckontheweb:
> I'll grab this -- could someone with more repo powers than I have assign me?

Done!

jisantuc:
Ok so here's where I've gotten to so far, and I need a little help with Python type gymnastics:

1. `ExtensionManagementMixin` expects a single type parameter (try to pass it two -- mypy isn't happy about it)
2. extending plural monomorhpizations of `ExtensionManagementMixin` isn't allowed -- you can't for example extend both `ExtensionManagementMixin[pystac.Item]` _and_ `ExtensionManagementMixin[pystac.Collection]`
3. To deal with 2, I thought I'd just extend `ExtensionManagmentMixin[T]` for the generic type `T`, defined in `eo.py`, but `ExtensionManagementMixin` requires a subtype of `STACObject`, and `Asset` is not a subtype of `STACObject`. I think there are reasons for that -- they don't have ids or extensions or links -- but it's a hurdle for extending `ExtensionManagementMixin[T]`

So I'm a bit flummoxed. I could be flummoxed because I'm not super familiar with Python's type system though -- is there a straightforward way out of this bind?

The current state of my branch has the generic version -- https://github.com/jisantuc/pystac/blob/bugfix/js/add-collection-to-ext-mixin-where-appropriate/pystac/extensions/eo.py#L279. You can run `scripts/test` from there to see what `mypy` is mad about with the bounds (ignore the non-`eo.py` errors -- those are also my fault and need the same treatment that ends up resolving this riddle, whatever that ends up being)

duckontheweb:
@jisantuc After looking at this a bit more, I think we might be able to solve this using `typing.Protocol` in `ExtensionsManagementMixin` instead of generics. I'm going to poke around a bit and post an example here if it seems to work.

duckontheweb:
[...]

PR #547, opened 2021-07-09
Add collection to ExtensionManagementMixin type params where appropriate

jisantuc:
**Related Issue(s):** #525 

**Description:** This PR adds `pystac.Collection` to the `ExtensionManagementMixin` type params (via a `Union`) for all extensions that provide concrete implementations of `SummariesExtension`.

**PR Checklist:**

- ~Documentation has been updated to reflect changes, if applicable~

### DIFF
--- a/pystac/extensions/eo.py
+++ b/pystac/extensions/eo.py
@@ -376,9 +376,15 @@ class EOExtension(
                 f"EO extension does not apply to type '{type(obj).__name__}'"
             )
 
-    @staticmethod
-    def summaries(obj: pystac.Collection) -> "SummariesEOExtension":
+    @classmethod
+    def summaries(
+        cls, obj: pystac.Collection, add_if_missing: bool = False
+    ) -> "SummariesEOExtension":
         """Returns the extended summaries object for the given collection."""
+        if not add_if_missing:
+            cls.validate_has_extension(obj)
+        else:
+            cls.add_to(obj)
         return SummariesEOExtension(obj)
 
 
@@ -470,10 +476,6 @@ class SummariesEOExtension(SummariesExtension):
     the ``summaries`` field of a :class:`~pystac.Collection` to include properties
     defined in the :stac-ext:`Electro-Optical Extension <eo>`.
     """
-
-    def __init__(self, collection: pystac.Collection):
-        EOExtension.add_to(collection)
-        super().__init__(collection)
 
     @property
     def bands(self) -> Optional[List[Band]]:
--- a/pystac/extensions/eo.py
+++ b/pystac/extensions/eo.py
@@ -28,7 +28,7 @@ from pystac.extensions import view
 from pystac.serialization.identify import STACJSONDescription, STACVersionID
 from pystac.utils import get_required, map_opt
 
-T = TypeVar("T", pystac.Item, pystac.Asset, pystac.Collection)
+T = TypeVar("T", pystac.Item, pystac.Asset)
 
 SCHEMA_URI: str = "https://stac-extensions.github.io/eo/v1.0.0/schema.json"
 PREFIX: str = "eo:"
--- a/pystac/extensions/eo.py
+++ b/pystac/extensions/eo.py
@@ -471,6 +471,10 @@ class SummariesEOExtension(SummariesExtension):
     defined in the :stac-ext:`Electro-Optical Extension <eo>`.
     """
 
+    def __init__(self, collection: pystac.Collection):
+        EOExtension.add_to(collection)
+        super().__init__(collection)
+
     @property
     def bands(self) -> Optional[List[Band]]:
         """Get or sets the summary of :attr:`EOExtension.bands` values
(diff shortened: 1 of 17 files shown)


pystac/extensions/eo.py
```diff
@@ -274,7 +275,9 @@ def band_description(common_name: str) -> Optional[str]:
 
 
 class EOExtension(
-    Generic[T], PropertiesExtension, ExtensionManagementMixin[pystac.Item]
+    Generic[T],
```
jisantuc, line 278: I had originally added `pystac.Collection` to the `Generic[T]` type param as well, but I think that was incorrect -- the `T` param there I think refers to the specific types of entities that this extension applies directly to, and collections aren't one of those

duckontheweb: That's correct, the `T` type should only apply to objects that can be extended by this extension (in this case `Asset` and `Item`) and the generic argument(s) to `ExtensionManagementMixin` cover any object that can contain the extension URI in its `stac_extensions` list (in this case `Item` and `Collection`).

duckontheweb (approved):
This looks great, thanks for getting it in!

Originally I was thinking we should add test cases for using `add_to` and `remove_from` on `Collection` instances to make sure it actually works at runtime, but I'm pretty confident that we are covering our bases with the typing. What do you think @jisantuc?

jisantuc:
How to test was going to be my next question -- conditions are:

1. runs without crashing
2. adds the URI appropriately

?

duckontheweb:
> How to test was going to be my next question -- conditions are:
> 
> 1. runs without crashing
> 2. adds the URI appropriately
> 
> ?

Yeah, and:
3. `remove_from` removes the URI

jisantuc:
Done in 6fb98fd and b6b2bf8 -- now with a .05% coverage improvement! 🌶

duckontheweb:
> Done in [6fb98fd](https://github.com/stac-utils/pystac/commit/6fb98fd93a6e19c14b67003f6e6b6ebe0357fbd6) and [b6b2bf8](https://github.com/stac-utils/pystac/commit/b6b2bf8b7d14101c3378f4d9474aa46ed32c4522) -- now with a .05% coverage improvement! 🌶

Yeehaw!

pystac/extensions/eo.py
```diff
@@ -296,7 +299,7 @@ def apply(
         self, bands: Optional[List[Band]] = None, cloud_cover: Optional[float] = None
     ) -> None:
         """Applies label extension properties to the extended :class:`~pystac.Item` or
-        :class:`~pystac.Collection`.
+        :class:`~pystac.Asset`.
```
duckontheweb, line 302: Good catch!

pystac/extensions/eo.py
```diff
@@ -468,6 +471,10 @@ class SummariesEOExtension(SummariesExtension):
     defined in the :stac-ext:`Electro-Optical Extension <eo>`.
     """
 
+    def __init__(self, collection: pystac.Collection):
+        EOExtension.add_to(collection)
+        super().__init__(collection)
+
```
duckontheweb, line 477: I think it might be better to put this logic into `EOExtension.summaries` and have that method behave similarly to [`EOExtension.ext`](https://github.com/stac-utils/pystac/blob/12eff70f9423d642c9909c92b4ba228bf97bef0e/pystac/extensions/eo.py#L350):
1. Add an `add_if_missing: bool = False` argument to `EOExtension.summaries`
2. If `add_if_missing` is `False`, raise an exception if someone tries to get summaries for a Collection that is missing the schema URI.
3. Add the schema URI, if necessary, when `add_is_missing` is `True`

The reasoning would be similar to that for the `ext` method (see [this discussion](https://github.com/stac-utils/stactools/pull/113#discussion_r639860502) for context). Basically, in cases where someone is using PySTAC for reading existing Collections it should let them know the extension is not implemented rather than fixing it behind the scenes. If someone really does want to fix it automatically, then `add_if_missing` gives them an easy but explicit way to do that.

@lossyrob @gadomski What are your thoughts on this?

This would obviously need to be applied to other extension with this pattern, too, if we do decide to go this route.

duckontheweb (changes requested):
Looks good overall.

I'm curious to get other input on how we handle automatically adding schema URIs to objects when summarizing (see review comment for more detail) before merging.

pystac/extensions/eo.py
gadomski: I've been liking the `add_if_missing` pattern; I can omit it when testing to get trigger an error on a missing schema, but it's still a one-liner to get an extension in (e.g.) `Item` creation. So I'm :+1: on @duckontheweb's proposal.

lossyrob: +1

jisantuc: Thanks for the links, I'm making this change this morning

jisantuc: I think I've faithfully followed this suggestion in 6b4413f -- lemme know if that's not what was intended, if it's consistent with what's requested I'll copy-🍝 my way through the other extensions

duckontheweb (approved):
Awesome, thanks for getting this in!

Issue #546, opened 2021-07-09
Slowness in `Item.to_dict()`, seemingly from links


TomAugspurger:
I'm trying to diagnose a slowdown I'm observing in `pystac.to_dict()` is slower. Right now I'm unsure if the slowdown is due to a change in pystac or in our STAC endpoint. Anyway, if I call `clear_links()` prior to `to_dict()` then things are faster (goes from about 1s to 200µs)

```python

In [1]: import pystac

In [2]: item = pystac.Item.from_file("https://planetarycomputer-staging.microsoft.com/api/stac/v1/collections/sentinel-2-l2a/items/S2A_MSIL2A_20160208T190542_R0
   ...: 13_T10TET")
   ...: %time _ = item.to_dict()
CPU times: user 25 ms, sys: 339 µs, total: 25.4 ms
Wall time: 987 ms

In [3]: item = pystac.Item.from_file("https://planetarycomputer-staging.microsoft.com/api/stac/v1/collections/sentinel-2-l2a/items/S2A_MSIL2A_20160208T190542_R0
   ...: 13_T10TET")
   ...: item.clear_links()
   ...: %time _ = item.to_dict()
CPU times: user 167 µs, sys: 18 µs, total: 185 µs
Wall time: 192 µs
```

So a couple questions:

1. Why is having those links present slowing things down? Glancing at the output of `to_dict`, I don't see why it would need to make an HTTP call or anything like that, right? Maybe through some chain calling `link.get_href()`?
2. Have there been any recent changes that would have introduced this slowdown? We've also been making changes to that STAC endpoint that could have slowed things down. Brief testing shows that RC1 - RC3 all have similar performance on this same endpoint (about 1s to do `to_dict()`
3. `to_dict()` has a parameter `include_self_link`. Could we add a parameter to exclude *all* links? For my application, I only need the assets and extension info, so making requests to fetch the link

TomAugspurger:
All the time is seemingly spent on `Item.get_root()`

```
link = item.links[0]
%lprun -f pystac.link.Link.get_href link.get_href()

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   112                                               def get_href(self) -> Optional[str]:
   121                                                   # get the self href
   122         1          8.0      8.0      0.0          if self.is_resolved():
   123                                                       href = cast(pystac.STACObject, self.target).get_self_href()
   124                                                   else:
[...]

PR #549, opened 2021-07-09
Allow root to be set during to_dict

lossyrob:
**Related Issue(s):** #546

**Description:**
There was existing unused arguments for setting the root on from_dict for Catalogs, Collections and Items. This commit implements and tests that the root is set to the parameter if present, and adds the parameter to ItemCollection, which itself does not have a root but instead passes the root onto each of the Items parsed during the from_dict call.

This will allow pystac_client to pass in the client Catalog as the root of the Items returned during a search. This will drastically speed up the `item.to_dict()` calls as the logic in to_dict checks the root to see if the HREFs should be absolute or relative, which ends up reading the root object for each Item in the search results; instead if the root is set at the time of `item.from_dict`, the reading of the root link of the item can be skipped.

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Changed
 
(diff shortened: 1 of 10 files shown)


duckontheweb (approved):
Looks great. I had always kind of wondered why that `root` parameter was hanging around...

Issue #521, opened 2021-07-06
Extensions: Throw error if `add_if_missing` is used when called on an Asset without a parent


lossyrob:
E.g. https://github.com/stac-utils/pystac/blob/main/pystac/extensions/eo.py#L366-L369

We should throw an exception if `add_if_missing` is used on an Asset without a parent. This is an explicitly set flag, so it would make sense that the user should be aware when it isn't having any effect.

PR #554, opened 2021-07-13
Raise exception when using add_if_missing on ownerless Asset

duckontheweb:
**Related Issue(s):**

* #521

**Description:**

Raises an `STACError` if `add_if_missing` is `True` for an `Asset` with no `owner`. 

This PR also moves the `add_if_missing` logic into `ExtensionManagementMixin.validate_has_extension` for DRYer code in the extension implementations.

**PR Checklist:**

### DIFF
--- a/pystac/extensions/base.py
+++ b/pystac/extensions/base.py
@@ -126,15 +126,26 @@ class ExtensionManagementMixin(Generic[S], ABC):
         )
 
     @classmethod
-    def validate_has_extension(cls, obj: Union[S, pystac.Asset]) -> None:
-        """Given a :class:`~pystac.STACObject` or :class:`pystac.Asset` instance, checks
-        if the object (or its owner in the case of an Asset) has this extension's schema
-        URI in it's :attr:`~pystac.STACObject.stac_extensions` list."""
-        extensible = obj.owner if isinstance(obj, pystac.Asset) else obj
-        if (
-            extensible is not None
-            and cls.get_schema_uri() not in extensible.stac_extensions
-        ):
+    def validate_has_extension(
+        cls, extensible: Optional[S], add_if_missing: bool
+    ) -> None:
+        """Given a :class:`~pystac.STACObject`, checks if the object has this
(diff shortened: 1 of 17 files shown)


pystac/extensions/base.py
```diff
@@ -126,15 +126,32 @@ def has_extension(cls, obj: S) -> bool:
...
+            add_if_missing : Whether to add the schema URI to the object if the URI is
+                not already present.
+
+        Raises:
+            STACError : If ``add_if_missing`` is ``True`` and ``extensible`` is None.
+        """
+        if add_if_missing:
+            if extensible is None:
+                raise pystac.STACError(
+                    "Can only add schema URIs to Assets with an owner."
```
lossyrob, line 147: We don't have a use case where something other than an Asset would be passed in by an extension developer, but I can see this being a bit confusing - it's not clear that if 'extensible' is None then it is an Asset without an owner.

"extensible" sounds like it should be a boolean as well, perhaps obj to match the other methods?

It might be worth having a separate method called `validate_owner_has_extension` that would call `validate_has_extension` in the case where there was an owner, and in the case where it doesn't to either fail if add_if_missing is true or silently do nothing. That might be more explicit about handling the logic around the ownership and still buy the one-liner when checking assets.

duckontheweb: > "extensible" sounds like it should be a boolean as well, perhaps obj to match the other methods?

Yeah, agreed. Changed this in d7de3120f657ddb55bef526cfa19c95f32035976



> It might be worth having a separate method called `validate_owner_has_extension`

Added this in 610e686214f484603a8941e511622493ff942f19. The typing is a bit tricky since `asset.owner` is `Union[Item, Collection]` and the `obj` argument to `validate_has_extension` is `S` (whatever `ExtensionManagementMixin` is generic over). In most cases these two end up lining up, but there's no guarantee, and the Item Asset extension is one example where they wouldn't. I solved this by just casting `asset.owner` to `S` and putting the onus on the specific extension implementation to ensure that an Asset owner has the right type.

lossyrob (approved):

PR #560, opened 2021-07-16
Use extension API for CommonMetadata

duckontheweb:
**Related Issue(s):**

- #481

**Description:**

- Moves `CommonMetadata` and `Provider` to their own modules
- Changes `CommonMetadata` to use similar API to `PropertyExtension`
- Updates `Asset.extra_fields` to include all fields (not just "extra" ones) and renames to `Asset.fields`
   This change was necessary to provide a mutable container on `Asset` that we could use in `CommonMetadata`

**TODO:**

- [ ] Add tests to bring coverage back up
- [ ] Check documentation for other references to `Asset.extra_fields` or `Asset.properties` (as it used to be named).

**PR Checklist:**


duckontheweb (commented):
@gadomski This changes `Asset.extra_fields` -> `Asset.fields` because it now includes all Asset fields instead of just the ones that weren't handled specifically as instance attributes. This sort of undoes of #510 in that the attribute is now called something different from what's in `Link`, `Item`, etc. This seemed like the right way to go since it represents something different, but let me know what you think. I can also add in an `extra_properties` attribute to make it consistent, but it's not as simple as subsetting the dictionary because there were places in the code where we were assigning to `extra_fields` and expecting it to stay in sync with the `Asset` attributes.

duckontheweb:
> @gadomski This changes `Asset.extra_fields` -> `Asset.fields` because it now includes all Asset fields instead of just the ones that weren't handled specifically as instance attributes. This sort of undoes of #510 in that the attribute is now called something different from what's in `Link`, `Item`, etc. This seemed like the right way to go since it represents something different, but let me know what you think. I can also add in an `extra_properties` attribute to make it consistent, but it's not as simple as subsetting the dictionary because there were places in the code where we were assigning to `extra_fields` and expecting it to stay in sync with the `Asset` attributes.

Actually, @gadomski, feel free to ignore this. Had a conversation with @lossyrob and we decided to not change the `Asset` API for now.

PR #563, opened 2021-07-16
Use extension-like API for CommonMetadata

duckontheweb:
**Related Issue(s):**

- #481

**Description:**

- Moves `CommonMetadata` and `Provider` to their own modules
- Defines equality for `Provider` objects
- Changes `CommonMetadata` to use similar API to `PropertyExtension`
- Adds `Asset.common_metadata` property

**TODO:**

- [x] Add tests to bring coverage back up

**PR Checklist:**


pystac/common_metadata.py
```diff
@@ -0,0 +1,215 @@
...
+from pystac.errors import STACError
+from typing import Any, cast, Dict, List, Optional, Type, TYPE_CHECKING, TypeVar, Union
+
+import pystac
+from pystac import utils
+
+if TYPE_CHECKING:
+    from pystac.provider import Provider
+    from pystac.asset import Asset
+    from pystac.item import Item
```
lossyrob, line 11: I think the pattern to use explicitly weird type names with TYPE_CHECKING, e.g. `from pystac.item import Item as Item_Type`, helps with making sure you don't use those types for anything else besides type signatures in situations where you can't import the type for both type checking and type usage. When I was doing the typing I was hitting runtime errors that weren't getting picked up by mypy/pyright because things were imported with TYPE_CHECKING True, and I was using them as run-time imported types. So I think we should be doing the aliasing, though it's not a super important change for the functionality of this PR and shouldn't be blocking (if for example there's inconsistency elsewhere about TYPE_CHECKING imports)

lossyrob (approved):
A suggestion about TYPE_CHECKING import aliasing but otherwise LGTM

pystac/common_metadata.py
duckontheweb: Yeah, that's a really good point. I just liked the way this looked, but I was running into those issues as well.

duckontheweb: Fixed in 5a72a2a04510ae1db54dd35d6e405b5379d5e781

Issue #425, opened 2021-06-10
How to set absolute href with 'http'?


nikkopante:
Hi, I am generating STAC jsons on local computer. I want to set an absolute endpoint with `URL` as href. However I cannot save it.
```
OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect: 'http:'
```

```
URL = 'http://sample-stac.s3-website-us-east-1.amazonaws.com/sample/stac'
for image in images:
	...
	item = pystac.Item(
                id=id_name,
                 geometry=footprint,
                 bbox=bbox,
                 datetime=date_time_obj,
                 properties={})
	...
	item_href = '{url}/stac/{capture_name}/{capture_name}.json'.format(url=URL, capture_name=id_name)
        item.set_self_href(item_href)
        item.resolve_links()
        item.validate()
        catalog.add_item(item)


catalog_href = '{url}/catalog.json'.format(url=URL)
catalog.set_self_href(catalog_href)
catalog.resolve_links()
catalog.validate()
catalog.save(catalog_type=pystac.CatalogType.ABSOLUTE_PUBLISHED)
```

How can I use an endpoint to set it as hrefs while using the local computer?

I have tried putting
```
STAC_IO.write_text_method = my_write_method
```
from [STAC_IO](https://pystac.readthedocs.io/en/latest/concepts.html#using-stac-io)

duckontheweb:
@cnpante12 Happy to try to help with this. A couple of questions for you...

Are you hoping to save the catalog to your local drive, or do you want to save it to the `http://sample-stac.s3-website-us-east-1.amazonaws.com/sample/stac/stac/{capture_name}/{capture_name}.json` location in the URL?

Also, what version of PySTAC are you using?

cnpante12:
Thank you @duckontheweb.
I am using the latest version of STAC. I want to save the catalog in the local drive but I want the href to be `http://sample-stac.s3-website-us-east-1.amazonaws.com/sample/stac/stac/{capture_name}/{capture_name}.json`. I am doing this because I will manually upload the static files in S3.

duckontheweb:
@cnpante12 If you are still looking for a solution to this, you should be able to accomplish this using the `dest_href` argument to `Catalog.save`. In your example above, this would involve changing the final line from...

```python
catalog.save(catalog_type=pystac.CatalogType.ABSOLUTE_PUBLISHED)
```

...to...

```python
local_path = ...  # This should be a path to a directory on your local machine
[...]

PR #565, opened 2021-07-17
Allow to save a catalog at a specific location

volaya:
This PR allows to specify a folder where a catalog should be saved, to be used instead of the pth taken for the catalog self href

This will allow implementing the functionaity requested at https://github.com/stac-utils/stactools/issues/12

It is still a work in progress. I am creating this PR to share the approach and gather some feedback about it. I will add unit tests and implement the corresponding code in StacTools later.

### DIFF
--- a/pystac/catalog.py
+++ b/pystac/catalog.py
@@ -29,7 +29,7 @@ from pystac.serialization import (
     identify_stac_object,
     migrate_to_latest,
 )
-from pystac.utils import is_absolute_href, make_absolute_href
+from pystac.utils import is_absolute_href, make_absolute_href, make_relative_href
 
 if TYPE_CHECKING:
     from pystac.item import Asset as Asset_Type, Item as Item_Type
@@ -705,16 +705,22 @@ class Catalog(STACObject):
 
         return result
 
-    def save(self, catalog_type: Optional[CatalogType] = None) -> None:
+    def save(
+        self,
+        catalog_type: Optional[CatalogType] = None,
+        dest_href: Optional[str] = None,
+    ) -> None:
         """Save this catalog and all it's children/item to files determined by the object's
-        self link HREF.
+        self link HREF or a specified path.
 
         Args:
             catalog_type : The catalog type that dictates the structure of
                 the catalog to save. Use a member of :class:`~pystac.CatalogType`.
                 If not supplied, the catalog_type of this catalog will be used.
                 If that attribute is not set, an exception will be raised.
-
+            dest_href : The location where the catalog is to be saved.
+                If not supplied, the catalog's self link HREF is used to determine
+                the location of the catalog file and children's files.
         Note:
             If the catalog type is ``CatalogType.ABSOLUTE_PUBLISHED``,
             all self links will be included, and hierarchical links be absolute URLs.
@@ -724,6 +730,7 @@ class Catalog(STACObject):
             If the catalog  type is ``CatalogType.SELF_CONTAINED``, no self links will
             be included and hierarchical links will be relative URLs.
         """
+
         root = self.get_root()
         if root is None:
             raise Exception("There is no root catalog")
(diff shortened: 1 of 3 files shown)


duckontheweb:
I think this will also help with #425

pystac/catalog.py
```diff
@@ -705,15 +705,24 @@ def generate_subcatalogs(
...
         Args:
             catalog_type : The catalog type that dictates the structure of
                 the catalog to save. Use a member of :class:`~pystac.CatalogType`.
                 If not supplied, the catalog_type of this catalog will be used.
                 If that attribute is not set, an exception will be raised.
+            dest_href : The location where the catalog is to be saved.
+                If not supplied, the catalog's self link HREF is used to determine
+                the location of the catalog file and children's files.
+                This is olnly used if the catalog type is
+                ``CatalogType.ABSOLUTE_PUBLISHED``.
```
duckontheweb, line 725: It might be good to add a check at the beginning of this method that raises a `STACError` if `dest_href` is not `None` and `root.catalog_type != CatalogType.ABSOLUTE_PUBLISHED`. I could see users being confused if they pass in a `dest_href` that just gets ignored silently.

pystac/catalog.py
```diff
@@ -739,9 +748,20 @@ def save(self, catalog_type: Optional[CatalogType] = None) -> None:
 
         for item_link in self.get_item_links():
             if item_link.is_resolved():
-                cast(pystac.Item, item_link.target).save_object(
-                    include_self_link=items_include_self_link
-                )
+                item = cast(pystac.Item, item_link.target)
+                item_dest_href = None
```
duckontheweb, line 752:
```suggestion
```

This ends up just being unused

pystac/catalog.py
```diff
@@ -739,9 +748,20 @@ def save(self, catalog_type: Optional[CatalogType] = None) -> None:
...
+                if dest_href is not None and root.catalog_type in [
+                    CatalogType.ABSOLUTE_PUBLISHED
+                ]:
+                    rel_href = make_relative_href(
+                        item.get_self_href(), self.get_self_href()
+                    )
```
duckontheweb, line 758: Since `get_self_href` can return `None`, we'll want to check for that case and raise an exception.

pystac/catalog.py
```diff
@@ -753,7 +773,17 @@ def save(self, catalog_type: Optional[CatalogType] = None) -> None:
             if root_link and root_link.get_absolute_href() == self.get_self_href():
                 include_self_link = True
 
-        self.save_object(include_self_link=include_self_link)
+        catalog_dest_href = None
+        if dest_href is not None and root.catalog_type in [
+            CatalogType.ABSOLUTE_PUBLISHED
+        ]:
+            rel_href = make_relative_href(self.get_self_href(), self.get_self_href())
```
duckontheweb, line 780: Same as above; we should check for the case where either (or both) of the `get_self_href` calls return `None` and raise a helpful exception here.

duckontheweb (commented):
Thanks for tackling this @volaya! This approach looks good to me, but I'm wondering if we really need to restrict `dest_href` to only apply to "absolute published" catalogs. I'm admittedly not as familiar with the nuances of the different catalog types, but it seems like if someone has to explicitly pass in the `dest_href` argument then they may have a good reason to do it for other catalog types. Maybe that would require much more complicated updating of the catalog links, though, not sure 🤷🏻

pystac/catalog.py
volaya: I replaced calls to `get_self_href()` with `self_href`, which raises an exception in case of `None` value

pystac/catalog.py
volaya: I restricted it to `CatalogType.ABSOLUTE_PUBLISHED`, because the original issue in the stactools repo mentioned the functionality for that case, but it's true that here it makes sense for other types as well (If anyone has a different opinion here or I am missing something, please comment here). I have removed that check and now the `dest_href` parameter is used regardless of the catalog type

duckontheweb (approved):
This looks good to me @volaya, thanks for contributing this! 

You mentioned in the original description that this is a WIP. Is that still true or is it ready to merge?

volaya:
@duckontheweb This should be now fine to merge. I added the matching functionality in StacTools here: https://github.com/stac-utils/stactools/pull/175

duckontheweb:
Coverage upload seems to be failing due to codecov/codecov-action#437, but everything else passes, so I'll merge this.

Issue #508, opened 2021-07-01
Add rich reprs to STAC objects


TomAugspurger:
IPython / jupyter define a protocol where objects defining methods like `_repr_html_` will have the ability to produce rich representations in a notebook. See https://ipython.readthedocs.io/en/stable/config/integrating.html.

I think this would be very useful for exploring STAC objects. https://github.com/dask/distributed/pull/4857 has some designs / code that might be worth stealing. In particular the handling of nested objects would apply to catalogs, collections, and even items.

See http://matthewrocklin.com/blog/2019/07/04/html-repr for some background.

DahnJ:
This seems like an immensely useful feature to have, especially given the nested structure STAC has.

PR https://github.com/stac-utils/pystac/pull/573 was closed since it used an extra dependency `jinja2`. The agreed-upon solution was to create a new package `pystac-jupyter`. Did that ever get made? If not, what would be needed to help it get made?

In case it didn't get made: would it perhaps be possible/convenient to only include `jinja2` as an [optional dependency](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies) and only attach/display the HTML representations if `jinja2` is installed? I haven't thought much about this, just thought I'd mention it.

cc: @moradology

TomAugspurger:
IMO, this would have to go in pystac itself to be worthwhile. https://ipython.readthedocs.io/en/stable/config/integrating.html#formatters-for-third-party-types mentions a way for third-party libraries to register formatters for types in other libraries, but that's not too user-friendly.

I think an optional dependency is a good compromise. The `_repr_html_` would then just need a

```python
try:
    import jinja2
except ImportError:
    return repr(self)
...
```

And in case it's helpful, `rich.inspect` can give a (verbose) repr:

<img width="754" alt="image" src="https://user-images.githubusercontent.com/1312546/144679292-9324d527-f11a-47bc-9900-4502291de733.png">

But I think that the nested nature of STAC objects means things like HTML dropdowns / disclosure triangles will be helpful.

duckontheweb:
Closed by #743

PR #573, opened 2021-07-20
Rich, HTML representations for Jupyter Notebook

moradology:
**Related Issue(s):** #508

**Description:**

This PR is intended to provide useful HTML representations of some core STAC objects as well as a structure appropriate to adding more such representations. These are primarily used within Jupyter notebooks, but could in principle find use elsewhere.

image

**PR Checklist:**


moradology:
In the above screenshot, `detail`/`summary` pairs don't appear to be rendering with arrows on their left (as expected). I'm not sure if this is a quirk of my particular installation, but would be interested if anyone has ideas

duckontheweb (commented):
Thanks for working on this @moradology , I like the idea of having these rich representations for contexts like Jupyter.

My biggest concern with this approach is adding `jinja2` as a dependency. PySTAC has been extremely careful about not adding non-essential dependencies (see [this comment](https://github.com/stac-utils/pystac/issues/331#issuecomment-867671808) from @lossyrob that gives some good rationale for this), and since `jinja2` is only being added to support a narrow set of features I don't feel that it's worth breaking that tradition.

I'm wondering if this kind of functionality would be better in a separate library (e.g. `pystac-jupyter`, `pystac-html`, etc.) that could provide more specific support in these environments. If we do want it in PySTAC, I think we should try to accomplish this through a combination of factory functions and f-strings instead of adding a new dependency.

moradology:
After writing this, reading your comment and a very brief discussion with @lossyrob, I'm inclined to agree that adding 1/4th more dependencies(!) is probably not wise. I'm not a huge fan of the f-string approach as it stands to add some clutter to already rather large classes and jinja templates are awfully convenient by comparison. Ironically, the f-string approach has fewer downsides within an external project focused on ergonomics inside the Jupyter context. My vote for preserving this kind of functionality - and I do think it is valuable for those who like to develop and to produce learning materials with Jupyter - would be to extend the primary `pystac` classes as necessary from inside an external project like `pystac-jupyter`, as you mentioned.

lossyrob:
I'd agree that creating a `pystac-jupyter` package is the best route, though a bit heavier of a lift. 

Noting that @mmcfarland pointed to these docs previously, which allows you to set custom formatters for types and would probably be useful here https://ipython.readthedocs.io/en/stable/config/integrating.html#formatters-for-third-party-types

duckontheweb:
@moradology Are you okay with me closing this PR since it sounds like this will be implemented as a separate library?

moradology:
Very

PR #575, opened 2021-07-22
Release v1.0.1

duckontheweb:
Patch release v1.0.1 to include change from #574.

@lossyrob I'm not sure if this is the best way to handle updating the CHANGELOG and library version. Making the change on this branch means our `main` branch is out of sync with the latest CHANGELOG and version. Applying the changes to `main` would mean that installing from that branch would include changes from #565 but be marked as `v1.0.1`, which didn't seem right. Let me know if there's a better strategy for this.

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,18 +1,10 @@
 # Changelog
 
-## [Unreleased]
-
-### Added
-
-### Removed
+## [v1.0.1]
 
 ### Changed
 
 - HREFs in `Link` objects with `rel == "self"` are converted to absolute HREFs ([#574](https://github.com/stac-utils/pystac/pull/574))
-
-### Deprecated
-
-### Fixed
 
 ## [v1.0.0]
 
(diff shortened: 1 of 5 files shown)


lossyrob:
What I've done in the past is:
- Make this PR against main, updating the version number and changelog
- Cherry pick these changes as well as #574 into the release branch

As you noted, there's a bit of inconsistency with the 1.0.1 version being on main ... but that inconsistency already exists, as it's currently labeled 1.0.0 where it has additional features to than that release.

Two ways to handle this I see are:
- Consider the version number on `main` as being the previous release, understanding that there will be additional changes to `main` that are not actually included in the release
- Change the version number on `main` to update the version number as part of the release process; e.g. we would have had a PR to update the version on `main` to be v1.1.0. 

I think the second option has issues because you don't actually know the next release number until you make it - e.g. in this situation, where the next release is actually a bugfix release.

I don't see too much issue with keeping things the way they are, in that the version number on main is the previous release + all the the unreleased development. So if we go with option 1, then the process I described should handle that fine, where the main version is 1.0.1 with the latest changelog, and both the version change and the desired bugfix PRs are cherry-picked into the release branch

duckontheweb:
> What I've done in the past is:
> 
> * Make this PR against main, updating the version number and changelog
> * Cherry pick these changes as well as [Relative self hrefs break link resolution #574](https://github.com/stac-utils/pystac/pull/574) into the release branch
> 
> As you noted, there's a bit of inconsistency with the 1.0.1 version being on main ... but that inconsistency already exists, as it's currently labeled 1.0.0 where it has additional features to than that release.

Thanks, that makes sense; I was having a hard time thinking through this clearly. I'll redo this PR on `main`.

PR #579, opened 2021-07-27
Include typing information during packaging

cuttlefish:
**Related Issue(s):**
stac-utils/stactools#155

**Description:**
This includes type information in the packages so that users of `pystac` can utilize it. This is particularly useful if the downstream project uses `mypy` or any of its competitors.

**PR Checklist:**

- ~[ ] Documentation has been updated to reflect changes, if applicable~

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,8 +20,9 @@ setup(
     author_email="stac@radiant.earth",
     url="https://github.com/stac-utils/pystac",
     packages=find_packages(),
+    package_data={p: ["py.typed"] for p in find_packages()},
     py_modules=[splitext(basename(path))[0] for path in glob("pystac/*.py")],
-    include_package_data=False,
+    include_package_data=True,
     python_requires=">=3.7",
     install_requires=[
         "python-dateutil>=2.7.0",
--- a/setup.py
+++ b/setup.py
@@ -20,9 +20,8 @@ setup(
     author_email="stac@radiant.earth",
     url="https://github.com/stac-utils/pystac",
     packages=find_packages(),
-    package_data={p: ["py.typed"] for p in find_packages()},
+    package_data={"": ["py.typed"]},
     py_modules=[splitext(basename(path))[0] for path in glob("pystac/*.py")],
-    include_package_data=True,
     python_requires=">=3.7",
     install_requires=[
         "python-dateutil>=2.7.0",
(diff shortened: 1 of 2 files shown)


cuttlefish:
@duckontheweb Python packaging is not my forte, but this seems to work based on my testing.
I'm hoping to use it in `stactools` stac-utils/stactools#182

setup.py
```diff
@@ -20,8 +20,9 @@
     author_email="stac@radiant.earth",
     url="https://github.com/stac-utils/pystac",
     packages=find_packages(),
+    package_data={p: ["py.typed"] for p in find_packages()},
```
duckontheweb, line 23:
```suggestion
    package_data={"": ["py.typed"]},
```

This will ensure that `py.typed` is included for all packages included in the distribution.

setup.py
```diff
@@ -20,8 +20,9 @@
...
     py_modules=[splitext(basename(path))[0] for path in glob("pystac/*.py")],
-    include_package_data=False,
+    include_package_data=True,
```
duckontheweb, line 25: I think we actually just want to leave this out completely and let the `package_data` argument handle this. This argument has caused a lot of confusion in general because it only works with files that are under version control and/or in a `MANIFEST.in` file. That covers our case here, but I usually prefer to just use `package_data` so that we're explicit about what we're including.

duckontheweb (commented):
Thanks for this @cuttlefish!

Based on past experience and brushing up on the [Data Files Support](https://setuptools.readthedocs.io/en/latest/userguide/datafiles.html#data-files-support) docs I think it might be best to just use `package_data` and not `include_package_data` (see specific comments below). Let me know what you think.

setup.py
cuttlefish: Ahh, when I was working out how to do it in a `setup.cfg` for `stactools` I saw something about `"" keys matching all` and had no idea what it was talking about :laughing:. Thanks!

setup.py
cuttlefish: You're right. It's working for me with this argument removed.

cuttlefish:
Thanks so much @duckontheweb! You're right about `include_package_data`. Just specifying `package_data` seems to do the trick.
I've also pulled in @gadomski's build fix, so hopefully the CI passes.

duckontheweb (approved):
👍🏻

Issue #589, opened 2021-07-29
Set _stac_io on constructed object if `root` passed into `from_dict`


matthewhanson:
The `from_dict` functions can now take in a root catalog. If provided, the resulting object should set it's `_stac_io` object to that of the root. 

As it stands now, if you get Collections from a root catalog, those collections will switch back to using the DefaultStacIO class.

PR #590, opened 2021-07-30
Apply StacIO instance from root in Catalog.set_root

duckontheweb:
**Related Issue(s):**

- #589

**Description:**

- Adds "public" `Catalog.stac_io` property that reads/writes to `Catalog._stac_io` for backwards compatibility.

    Since we are reading and writing to this attribute from outside the owning class in a number of cases it seems appropriate to make it a public property.
 
- Adds the `StacIO` instance from `root.stac_io` to the calling `Catalog` in `Catalog.set_root`.

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### Added
 
+- "Public" `Catalog.stac_io` property to match "private" `Catalog._stac_io` attribute
+  ([#590](https://github.com/stac-utils/pystac/pull/590))
+
 ### Removed
 
 ### Changed
@@ -11,6 +14,9 @@
 - Enable [strict
   mode](https://mypy.readthedocs.io/en/latest/command_line.html?highlight=strict%20mode#cmdoption-mypy-strict)
   for `mypy` ([#591](https://github.com/stac-utils/pystac/pull/591))
+- `Catalog.set_root` also sets `Catalog.stac_io` for the calling instance to be the same
+  as `root.stac_io`, if that value is not `None`
+  ([#590](https://github.com/stac-utils/pystac/pull/590))
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,6 @@
 
 ### Added
 
-- "Public" `Catalog.stac_io` property to match "private" `Catalog._stac_io` attribute
-  ([#590](https://github.com/stac-utils/pystac/pull/590))
-
 ### Removed
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Removed
 
+- Exclude `tests` from package distribution. This should make the package lighter ([#604](https://github.com/stac-utils/pystac/pull/604))
+
 ### Changed
 
 - Enable [strict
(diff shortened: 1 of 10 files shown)


lossyrob:
> Since we are reading and writing to this attribute from outside the owning class in a number of cases it seems appropriate to make it a public property.

The intent with using the underscore was to indicate that the `_stac_io` property was not part of the public API, i.e. not for user consumption but only for internal library use. This is a different mechanism than the strict public/private-per-class strategy found in typed OO languages. As a Python user, when I see an underscore, I know that I can't rely on that property as a user and should stay away from messing with it. By making this "public" (no underscore), we're committing to this as public API that we should expect users to read from, set into, and generally rely on. Is this the intended change, or do we want that property only to be used within the library code?

duckontheweb:
> The intent with using the underscore was to indicate that the `_stac_io` property was not part of the public API, i.e. not for user consumption but only for internal library use. This is a different mechanism than the strict public/private-per-class strategy found in typed OO languages. As a Python user, when I see an underscore, I know that I can't rely on that property as a user and should stay away from messing with it. By making this "public" (no underscore), we're committing to this as public API that we should expect users to read from, set into, and generally rely on. Is this the intended change, or do we want that property only to be used within the library code?

That's a good point. If someone were constructing a `Catalog` instance manually (using `__init__`), then they might want to set the `stac_io` attribute directly if they don't want to use the default for some reason. However, it's probably not worth the change until that case actually comes up. I can revert that part of the PR unless there are others who feel it should be a public API.

@gadomski @matthewhanson Are either of you setting `Catalog._stac_io` directly in `stactools` or `pystac-client`?

gadomski:
> @gadomski @matthewhanson Are either of you setting Catalog._stac_io directly in stactools or pystac-client?

**pystac-client** does, e.g. here: https://github.com/stac-utils/pystac-client/blob/main/pystac_client/client.py#L67. **stactools** does not.

duckontheweb:
> > @gadomski @matthewhanson Are either of you setting Catalog._stac_io directly in stactools or pystac-client?
> 
> **pystac-client** does, e.g. here: https://github.com/stac-utils/pystac-client/blob/main/pystac_client/client.py#L67. **stactools** does not.

Thanks @gadomski. Based on the fact that the `pystac-client` use-case is calling a method not present in any of the PySTAC `StacIO` classes I think it's probably enough of an edge case that it doesn't justify making this a public attribute.

Unless there are other objections or arguments for a public `Catalog.stac_io` attribute I'll revert that change.

duckontheweb:
Reverted `Catalog.stac_io` -> `Catalog._stac_io` and rebased to resolve merge conflicts. @lossyrob @matthewhanson This is ready for a review.

CHANGELOG.md
```diff
@@ -4,13 +4,19 @@
 
 ### Added
 
+- "Public" `Catalog.stac_io` property to match "private" `Catalog._stac_io` attribute
```
lossyrob, line 7: This entry should be removed

lossyrob (approved):
+1 after the CHANGELOG is updated to remove the public stac_io property mention

duckontheweb:
@matthewhanson I'll plan on merging this today unless I hear from you that there are issues.

Issue #480, opened 2021-06-26
Unable to control subfolders with LayoutTemplate


scottyhq:
the `generate_subcatalogs()` command is very handy for creating new catalogs with a layout based on STAC item properties! But it seems there is a bug or missing documentation. All the examples seem to map 1 property to 1 subfolder (https://pystac.readthedocs.io/en/latest/api.html?highlight=template#templating). 

In many cases it can be useful to merge properties names into a single subfolder name or do some other customization. For example I was expecting the template string to behave like a python f'string'. In fact it looks like the '/' and all other characters outside ${} are ignored 

```python
template = '${mgrs:utm_zone}${mgrs:latitude_band}${mgrs:grid_square}/${year}'
collection.generate_subcatalogs(template=template)
collection.describe()
```

Expected Output:
```
* <Collection id=sentinel1-rtc-aws>
    * <Catalog id=10UCU>
        * <Catalog id=2017>
          * <Item id=S1A_20170101_10UCU_ASC>
          * <Item id=S1A_20170125_10UCU_ASC>
          * <Item id=S1B_20170102_10UCU_ASC>
```

Actual Output:
```
* <Collection id=sentinel1-rtc-aws>
    * <Catalog id=10>
        * <Catalog id=U>
            * <Catalog id=CU>
                * <Catalog id=2017>
                  * <Item id=S1A_20170101_10UCU_ASC>
                  * <Item id=S1A_20170125_10UCU_ASC>
                  * <Item id=S1B_20170102_10UCU_ASC>
```

scottyhq:
@lossyrob @fnattino is there a workaround for this that I'm overlooking? Pinging you both b/c of your work and comments on this function in https://github.com/stac-utils/pystac/pull/241

fnattino:
Hi @scottyhq @lossyrob, apologies for the late answer, I wanted to have a look at the issue but could not find time until now. 

It looks to me that there is no workaround at the moment, but I think this could be fixed rather easily.

At the moment, the subcatalogs that are added to the structure are taken from `LayoutTemplate.get_template_values(item)`:
https://github.com/stac-utils/pystac/blob/e744897c63fd5561040396231f52e1cec398b024/pystac/catalog.py#L668
Iterating over that dictionary leads to one subdirectory being added for each property in the template, regardless of the separator between the property names. 

Using instead ` LayoutTemplate.substitute(item)`, should lead to the correct behaviour: 
```python
subcats = LayoutTemplate.substitute(item)
[...]

PR #595, opened 2021-08-04
Better control on subfolders in `generate_subcatalogs`

fnattino:
**Related Issue(s):** #480 

**Description:**

Fix `Catalog.generate_subcatalogs()` to allow for better control on the subcatalogs:
* subfolders can include multiple template values (e.g. using "${property1}_${property2}" will add a single subfolder layer);
* subfolders that are common to all items can be setup directly via the template (e.g. "2021-08-04/${property1}" will add the `2021-08-04` subcatalog for all items, without having to use the `defaults` argument)

**PR Checklist:**

### DIFF
--- a/pystac/catalog.py
+++ b/pystac/catalog.py
@@ -665,13 +665,10 @@ class Catalog(STACObject):
         for link in item_links:
             link.resolve_stac_object(root=self.get_root())
             item = cast(pystac.Item, link.target)
-            item_parts = layout_template.get_template_values(item)
+            subcat_ids = layout_template.substitute(item).split("/")
             id_iter = reversed(parent_ids)
             if all(
-                [
-                    "{}".format(id) == next(id_iter, None)
-                    for id in reversed(list(item_parts.values()))
-                ]
+                ["{}".format(id) == next(id_iter, None) for id in reversed(subcat_ids)]
             ):
                 # Skip items for which the sub-catalog structure already
                 # matches the template. The list of parent IDs can include more
@@ -679,12 +676,11 @@ class Catalog(STACObject):
                 keep_item_links.append(link)
                 continue
             curr_parent = self
-            for k, v in item_parts.items():
-                subcat_id = "{}".format(v)
+            for subcat_id in subcat_ids:
                 subcat = curr_parent.get_child(subcat_id)
                 if subcat is None:
-                    subcat_desc = "Catalog of items from {} with {} of {}".format(
-                        curr_parent.id, k, v
+                    subcat_desc = "Catalog of items from {} with id {}".format(
+                        curr_parent.id, subcat_id
                     )
                     subcat = pystac.Catalog(id=subcat_id, description=subcat_desc)
                     curr_parent.add_child(subcat)
--- a/pystac/layout.py
+++ b/pystac/layout.py
@@ -54,6 +54,10 @@ class LayoutTemplate:
     | ``collection``     | The collection ID of an Item's collection.             |
     +--------------------+--------------------------------------------------------+
 
+    The forward slash (``/``) should be used as path separator in the template
+    string regardless of the system path separator (thus both in POSIX-compliant
+    and Windows environments).
+
     Examples::
 
         # Uses the year, month and day of the item
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
 
 ### Fixed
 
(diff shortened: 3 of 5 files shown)


pystac/catalog.py
```diff
@@ -665,26 +665,22 @@ def generate_subcatalogs(
         for link in item_links:
             link.resolve_stac_object(root=self.get_root())
             item = cast(pystac.Item, link.target)
-            item_parts = layout_template.get_template_values(item)
+            subcat_ids = layout_template.substitute(item).split("/")
```
duckontheweb, line 668: Thanks for submitting this @fnattino! It looks like there is a test failure on Windows that I think is probably due to splitting on `"/"`, which wouldn't be used as a filepath delimiter on Windows:

```
FAIL: test_generate_subcatalogs_works_for_subcatalogs_with_same_ids (test_catalog.CatalogTest)
----------------------------------------------------------------------
  File "D:\a\pystac\pystac\tests\test_catalog.py", line 541, in test_generate_subcatalogs_works_for_subcatalogs_with_same_ids
    self.assertEqual(len(result), 6)
AssertionError: 4 != 6

----------------------------------------------------------------------
Ran 527 tests in 52.455s

FAILED (failures=1, skipped=3)
```

I'm not as familiar with the templating code in PySTAC, but I'll take a look later today to see if there is a good fix for this.

fnattino: Uh, thanks for the heads-up! Indeed the template used in one of the tests is  constructed via the `join_path_or_url` function:
https://github.com/stac-utils/pystac/blob/e744897c63fd5561040396231f52e1cec398b024/tests/test_catalog.py#L515
which results in `"\"` being used as separator on Windows (and the test failing in the splitting).

A question: Are users on Windows actually expected to provide the template using `"\"` as separator? Since the template is just a string used to define the catalog structure, would it make sense to adopt the convention of using one separator (say `"/"`) on all OSs? If you agree, I would be happy to fix the test and add a line to the documentation..

duckontheweb: > Since the template is just a string used to define the catalog structure, would it make sense to adopt the convention of using one separator (say `"/"`) on all OSs?

I think the safest bet might actually be to use `os.sep` instead of `"/"` so that it uses whatever path separator is set for that system.

```suggestion
            subcat_ids = layout_template.substitute(item).split(os.sep)
```

I'm a little fuzzy on whether you can use `/` as a path separator on Windows, and I know that in rare cases users can also set a different path separator for their system (but I don't know if `/` will also still work in that case). If we were to start using `/` as our file path separator we might also need to update the `LayoutTemplate.substitute` code to replace `"/"` with `os.sep` in order to get actual sub-directories (again, I'm a little fuzzy on how exactly this works on Windows), which would cause problems when templating URLs on Windows.

@gadomski @matthewhanson Do either of you have a strong opinion on how this should work?

gadomski: My instinct is to always use `/` when creating paths, templates, etc, and only convert to OS paths when needed to do filesystem operations. IMO using [PurePath](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath) would be the cleanest abstraction but would likely require some significant refactoring.

duckontheweb: > My instinct is to always use `/` when creating paths

Yeah, mine too. And actually now that I look at assertions [this this](https://github.com/stac-utils/pystac/blob/e744897c63fd5561040396231f52e1cec398b024/tests/test_layout.py#L46) it seems like we are implicitly assuming that we're using a `/` separator anyway.

@fnattino Let's stick with your suggestion of using `"/"` as the separator, updating the test and adding a note to the docs.

fnattino: Just implemented the changes, sorry for the delay with this!

pystac/layout.py
```diff
@@ -54,6 +54,10 @@ class LayoutTemplate:
     | ``collection``     | The collection ID of an Item's collection.             |
     +--------------------+--------------------------------------------------------+
 
+    The forward slash (``/``) should be used as path separator in the template
+    string regardless of the system path separator (thus both in POSIX-compliant
+    and Windows environments).
+
```
duckontheweb, line 60: 👍🏻

pystac/catalog.py
duckontheweb: No problem, thanks for this!

duckontheweb (approved):
Looks great to me, thanks again!

PR #607, opened 2021-08-17
Get link title from target, when possible

### LINKED ISSUE #358: Include 'title' in href when the target link's title can be easily discerned
There's going to be one minor change for STAC 1.0.0, that's just a small recommendation, but it'd be great to implement in pystac.

See https://github.com/radiantearth/stac-spec/pull/1133

I don't think we need to try to always fill out the title, but when it's easy for PySTAC, like since it's also writing out the title of the target entity (item, catalog, collection) then it'd be good to include it. Perhaps we also could have some nudge to encourage people to include titles...

This was originally thought about as an 'option' in stactools, see https://github.com/stac-utils/stactools/issues/22 but since it's getting in the spec as a recommendation it probably makes sense to just try to do it by default in pystac.

duckontheweb:
**Related Issue(s):**

- Closes #358 

**Description:**

Gets `Link.title` from the target object (`Link._target_object`) when `title` is `None` and it is possible to do so.

**PR Checklist:**

**TODO:**

- [ ] Add unit test for title being included in serialized `Link`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Enable [strict
   mode](https://mypy.readthedocs.io/en/latest/command_line.html?highlight=strict%20mode#cmdoption-mypy-strict)
   for `mypy` ([#591](https://github.com/stac-utils/pystac/pull/591))
+- Links will get their `title` from their target if no `title` is provided ([#607](https://github.com/stac-utils/pystac/pull/607))
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


gadomski (approved):
LGTM. Is this "auto-generated" title included in serialization, e.g. are we (possibly) implicitly adding fields to the generated JSON? Not saying that's a bad thing, just curious.

duckontheweb:
> LGTM. Is this "auto-generated" title included in serialization, e.g. are we (possibly) implicitly adding fields to the generated JSON? Not saying that's a bad thing, just curious.

Yes, I think that was actually the original motivation behind @cholmes 's issue. The idea was to automatically generate link titles in STAC catalogs that people are creating to be in line with the best practices. 

Now that you mention it though, I should add a test specifically for that...

gadomski (approved):

lossyrob (approved):

PR #617, opened 2021-08-31
RFC: Use mutable dictionaries for object properties

duckontheweb:
**Related Issue(s):**

- #546 

**Description:**

This PR is intended to be a proof-of-concept for changing the JSON-serializable classes to maintain a mutable dictionary in a `fields` property instead of setting instance attributes. The current draft only implements this for the `Asset` class, but if there is interest I will attempt to implement this for the `Link` class next since we have seen performance issues in the `to_dict` method there.

Some possible advantages to this approach are:

- Improvement in serialization/deserialization performance, especially when using `preserve_dict=False`
- Possibly easier/clearer typing
   If we are able to implement this for all JSON serializable classes, then we may not need some of the `Union` types in our `PropertiesExtension` classes since those could utilize the `fields` dictionary instead. This _may_ also ease some of our circular imports since we might be able to avoid some blocks like:
   ```python
   if isinstance(obj, pystac.Item):
       ...
   elif isinstance(obj, pystac.Asset):
       ...
   ```
   that are necessitating those imports
- Methods like `Item.set_collection` might no longer be necessary if we can handle them within the setters
 
Some possible hiccups:

- Handling attributes of the `Item` class might be tricky since some of them are at the top level and some are in the `properties`.
- Possibly worse performance when accessing instance attributes (e.g. `Asset.href`) since we are now accessing those as properties.
- May not be able to completely preserve backwards compatibility
   The current behavior of the `extra_fields` attribute on some of the classes may be hard to replicate. Currently, this attribute is a mutable dictionary that only includes the fields not covered by specific attributes (e.g. `Asset.href`), but also allows mutatation of the object (e.g. `asset.extra_fields["some_field"] = "some_value"` will add the `"some_field"` property to the Asset when serialized).

**PR Checklist:**

cc: @TomAugspurger @gjoseph92

### DIFF
--- a/pystac/asset.py
+++ b/pystac/asset.py
@@ -60,7 +60,7 @@ class Asset(base.JSONObject):
     @property
     def href(self) -> str:
         """Link to the asset object. Relative and absolute links are both allowed."""
-        return self._get_field("href", str, required=True)
+        return utils.get_required(self.fields.get("href"), self, "href")
 
     @href.setter
     def href(self, v: str) -> None:
@@ -69,7 +69,7 @@ class Asset(base.JSONObject):
     @property
     def title(self) -> Optional[str]:
         """Optional displayed title for clients and users."""
-        return self._get_field("title", str)
+        return self.fields.get("title")
 
     @title.setter
     def title(self, v: Optional[str]) -> None:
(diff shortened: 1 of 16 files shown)


duckontheweb:
I've been working on this some more here and there, but I'm having trouble implementing this in a non-breaking way. I will open up a new PR with a breaking implementation that we can consider as part of a new major version release.

PR #654, opened 2021-11-08
Fix string enum serialization

### LINKED ISSUE #626: Change of behavior in `str(pystac.RelType.root)`
In pystac 1.0, calling str on the value of an enum like `RelType.ROOT` would return the value:

```python
In [1]: import pystac

In [2]: str(pystac.RelType.ROOT)  # pystac 1.0.0
Out[2]: 'root'
```

On the latest release (1.1.0) that includes the class name.

```python
In [4]: str(pystac.RelType.ROOT)
Out[4]: 'RelType.ROOT'
```

I suspect this was changed in https://github.com/stac-utils/pystac/pull/514.

I was relying on the old behavior in https://github.com/TomAugspurger/xstac, which seems useful to me. Do we want to restore the 1.0 behavior? Or stick with the 1.1 behavior?

### LINKED ISSUE #652: link.py to_dict is not properly serializing RelType enum values
The `to_dict` function in `link.py` is not serializing `rel` properly in case of using a `RelType` enum.
(`rel` is an `Union[str, RelType]`)
https://github.com/stac-utils/pystac/blob/0ca29725fc58b1d6e6c21d0928ddafb690c0536f/pystac/link.py#L314

In my example, by default the `rel` is `pystac.RelType.ROOT` and when I call the `to_dict` function on a `Collection` object, the `link` included in the `dictionary` returned is the following:
`'links': [{'rel': <RelType.ROOT: 'root'>, 'href': '/home/paul/code/linz/topo-processor/collection.json', 'type': <MediaType.JSON: 'application/json'>}]`
I would expect that `to_dict` return the `str` enum value `"root"` in my case.

Note that causes an `Unsupported type: 'RelType'` error when calling a `jsonschema-rs` validator with this dict.

duckontheweb:
**Related Issue(s):**

- Closes #652 
- Closes #626

**Description:**

Creates a `StringEnum` class that overwrites the default `__str__` method to return the value of the enum. All instances of classes inheriting from `str` and `Enum` have been updated to inherit from this base class, including `RelType`.

**PR Checklist:**

### DIFF
--- a/tests/test_link.py
+++ b/tests/test_link.py
@@ -186,6 +186,18 @@ class LinkTest(unittest.TestCase):
         assert link.title == link_title
         assert link.to_dict().get("title") == link_title
 
+    @unittest.expectedFailure
+    def test_serialize_link(self) -> None:
+        href = "https://some-domain/path/to/item.json"
+        title = "A Test Link"
+        link = pystac.Link(pystac.RelType.SELF, href, pystac.MediaType.JSON, title)
+        link_dict = link.to_dict()
+
+        assert str(link_dict["rel"]) == "self"
+        assert str(link_dict["type"]) == "application/json"
+        assert link_dict["title"] == title
+        assert link_dict["href"] == href
+
 
 class StaticLinkTest(unittest.TestCase):
     def setUp(self) -> None:
--- a/tests/test_link.py
+++ b/tests/test_link.py
@@ -192,10 +192,10 @@ class LinkTest(unittest.TestCase):
         link = pystac.Link(pystac.RelType.SELF, href, pystac.MediaType.JSON, title)
         link_dict = link.to_dict()
 
-        assert str(link_dict["rel"]) == "self"
-        assert str(link_dict["type"]) == "application/json"
-        assert link_dict["title"] == title
-        assert link_dict["href"] == href
+        self.assertEqual(str(link_dict["rel"]), "self")
+        self.assertEqual(str(link_dict["type"]), "application/json")
+        self.assertEqual(link_dict["title"], title)
+        self.assertEqual(link_dict["href"], href)
 
 
 class StaticLinkTest(unittest.TestCase):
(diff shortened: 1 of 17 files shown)


duckontheweb:
cc: @tomaugsperber

tests/test_link.py
```diff
@@ -186,6 +186,17 @@ def test_title_as_init_argument(self) -> None:
         assert link.title == link_title
         assert link.to_dict().get("title") == link_title
 
+    def test_serialize_link(self) -> None:
+        href = "https://some-domain/path/to/item.json"
+        title = "A Test Link"
+        link = pystac.Link(pystac.RelType.SELF, href, pystac.MediaType.JSON, title)
+        link_dict = link.to_dict()
+
+        assert str(link_dict["rel"]) == "self"
```
TomAugspurger, line 195: Is this project using unittest still, or has it migrated to pytest? Might want self.assertEqual.

TomAugspurger (approved):

tests/test_link.py
duckontheweb: We're still on `unittest`. I would definitely be in favor of switching to `pytest` if someone were willing to do the legwork to convert our tests.

duckontheweb: Changed these to `self.assertEqual` in the latest commit, thanks for the suggestion!

lossyrob:
@duckontheweb  I see a number of Enums that are still inheriting from `(str, enum.Enum)`. See the raster, sar, and sat extensions. Is there a reason these shouldn't also be StringEnums?

duckontheweb:
> @duckontheweb I see a number of Enums that are still inheriting from `(str, enum.Enum)`. See the raster, sar, and sat extensions. Is there a reason these shouldn't also be StringEnums?

Ah, sorry. The find/replace I used was just looking for `str, Enum`. I'll broaden that and convert those others as well.

duckontheweb:
> > @duckontheweb I see a number of Enums that are still inheriting from `(str, enum.Enum)`. See the raster, sar, and sat extensions. Is there a reason these shouldn't also be StringEnums?
> 
> Ah, sorry. The find/replace I used was just looking for `str, Enum`. I'll broaden that and convert those others as well.

@lossyrob This is fixed in the latest commit.

lossyrob (approved):

Issue #657, opened 2021-11-09
Absolute links in a self contained STAC Catalog Items


pomadchin:
Generation of a SELF_CONTAINED catalog produces all correct relative links in catalogs and collections but not in **items**.

Code example to reproduce:

```python
import pystac
from datetime import datetime, timezone
from shapely.geometry import box, mapping

def tiny():
    # Create a catalog
    catalog = pystac.Catalog(id="example-catalog-id", description="descr")
    collection = pystac.Collection(
        "example-collection-id",
        "descr",
        pystac.Extent(
            spatial=pystac.SpatialExtent([[-180, -90, 180, 90]]),
            temporal=pystac.TemporalExtent(
                [[datetime(2020,7,3, tzinfo=timezone.utc), None]]
            ),
        ),
        license='proprietary'
    )

    item = pystac.Item(
        id="example-item-id",
        stac_extensions=[],
        geometry=mapping(box(0, 0, 90, 90)),
        bbox=[0, 0, 90, 90],
        datetime=datetime(2021,11,1),
        properties={}
    )

    collection.add_item(item)
    catalog.add_child(collection)
    catalog_path = './data/catalog/example-catalog'
    catalog.normalize_hrefs(catalog_path)
    catalog.validate_all()
    print(f'Saving catalog to {catalog_path}...')
    catalog.save(catalog_type=pystac.CatalogType.SELF_CONTAINED)
    print("Done!")

if __name__ == "__main__":
    tiny()
```

Item json in this case:

```javascript
{
  "type": "Feature",
  "stac_version": "1.0.0",
  "id": "example-item-id",
  // ...
  "links": [
    {
      "rel": "root",
      "href": "/fullpath/data/catalog/example-catalog/example-collection-id/collection.json",
      "type": "application/json"
    },
    {
      "rel": "collection",
      "href": "/fullpath/data/catalog/example-catalog/example-collection-id/collection.json",
      "type": "application/json"
    },
    {
      "rel": "parent",
      "href": "/fullpath/data/catalog/example-catalog/example-collection-id/collection.json",
      "type": "application/json"
    }
  ],
  // ...
  "collection": "example-collection-id"
}
```

Gist with the same code, requirements.txt and the output example: https://gist.github.com/pomadchin/7c4c8365e1d3098956acc974f39a4cb6

I will try to prep a PR with a unit test to cover this case as well.

duckontheweb:
Closed via #658

PR #658, opened 2021-11-09
Add an extra SELF_CONTAINED catalog test

pomadchin:
**Related Issue(s):** #657

**Description:**
Adds a unit test that tests #657 

I notcied that there is an issue only with catalogs of the full hierarchy:

* catalog
  * collection
     * item

The following one works as expected:
* catalog
  * item

**PR Checklist:**

### DIFF
--- a/tests/test_catalog.py
+++ b/tests/test_catalog.py
@@ -842,11 +842,13 @@ class CatalogTest(unittest.TestCase):
             cat = TestCases.test_case_9()
             cat.normalize_hrefs(tmp_dir)
             cat.validate_all()
-            print(f'Saving catalog to: {tmp_dir}')
+            print(f"Saving catalog to: {tmp_dir}")
             cat.save(catalog_type=CatalogType.SELF_CONTAINED)
 
             # read the item back, to make experiment clean
-            item = Item.from_file(f'{tmp_dir}/collection-issue-657/item-issue-657/item-issue-657.json')
+            item = Item.from_file(
+                f"{tmp_dir}/collection-issue-657/item-issue-657/item-issue-657.json"
+            )
             # ensure that all links in the output json are relative
             # everything of a nested level > 2 is not relative?
             for link in item.links:
--- a/tests/test_catalog.py
+++ b/tests/test_catalog.py
@@ -1119,7 +1119,7 @@ class FullCopyTest(unittest.TestCase):
             self.check_link(link, tag)
 
     def check_catalog(self, c: Catalog, tag: str) -> None:
-        self.assertEqual(len(c.get_links("root")), 1)
+        self.assertEqual(len(c.get_links("root")), 1, msg=f"{c}")
 
         for link in c.links:
             self.check_link(link, tag)
--- a/tests/test_catalog.py
+++ b/tests/test_catalog.py
@@ -837,6 +837,23 @@ class CatalogTest(unittest.TestCase):
                 c2.catalog_type = CatalogType.ABSOLUTE_PUBLISHED
                 check_all_absolute(c2)
 
+    def test_self_contained_catalog_collection_item_links(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            cat = TestCases.test_case_9()
+            cat.normalize_hrefs(tmp_dir)
+            cat.validate_all()
+            print(f'Saving catalog to: {tmp_dir}')
+            cat.save(catalog_type=CatalogType.SELF_CONTAINED)
+
+            # read the item back, to make experiment clean
+            item = Item.from_file(f'{tmp_dir}/collection-issue-657/item-issue-657/item-issue-657.json')
+            # ensure that all links in the output json are relative
+            # everything of a nested level > 2 is not relative?
+            for link in item.links:
+                self.assertFalse(is_absolute_href(link.target))
+            # use time sleep to pause there and inspect catalogs manually
+            # time.sleep(10000)
+
     def test_full_copy_and_normalize_works_with_created_stac(self) -> None:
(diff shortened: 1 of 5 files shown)


tests/test_catalog.py
```diff
@@ -837,6 +837,23 @@ def check_all_absolute(cat: Catalog) -> None:
...
+            cat.validate_all()
+            print(f'Saving catalog to: {tmp_dir}')
+            cat.save(catalog_type=CatalogType.SELF_CONTAINED)
+
+            # read the item back, to make experiment clean
+            item = Item.from_file(f'{tmp_dir}/collection-issue-657/item-issue-657/item-issue-657.json')
+            # ensure that all links in the output json are relative
+            # everything of a nested level > 2 is not relative?
+            for link in item.links:
+                self.assertFalse(is_absolute_href(link.target))
```
pomadchin, line 853: Is it even a valid test? pystac can fix links somehow... to make this test even more explicit I can load item as a json dict 🦐

duckontheweb: Yeah, `link.target` can be either a string or a `STACObject` instance, but `is_absolute_href` expects only strings, so this is raising a `mypy` error. It might be best to do something like:

```python
for link in item.links:
    href = link.get_href()
    self.assertIsNotNone(href)
    self.assertFalse(is_absolute_href(href))
```

pomadchin: done! thanks

tests/test_catalog.py
```diff
@@ -837,6 +837,27 @@ def check_all_absolute(cat: Catalog) -> None:
...
+            # read the item back, to make experiment clean
+            item = Item.from_file(
+                f"{tmp_dir}/collection-issue-657/item-issue-657/item-issue-657.json"
+            )
+            # ensure that all links in the output json are relative
+            # everything of a nested level > 2 is not relative?
+            for link in item.links:
+                href = link.get_href()
+                self.assertIsNotNone(href)
+                self.assertFalse(is_absolute_href(href))
```
pomadchin, line 857: Is there a way around the following error: 

```python
tests/test_catalog.py:858: error: Argument 1 to "is_absolute_href" has incompatible type "Optional[str]"; expected "str"  [arg-type]
Found 1 error in 1 file (checked 88 source files)
```

duckontheweb: Sorry, I led you a little astray there. I thought `mypy` would recognize `self.assertIsNotNone` as eliminating the possibility of a `None` type for that variable, but it looks like it does not. 

Using `assert href is not None` instead of `self.assertIsNotNone(href)` fixes that issue.

lossyrob:
I just updated to avoid checking the "self" link, which per the spec is always absolute. This ends up making the test pass. I'm going to try and get it to re-fail in the way @pomadchin mentioned in Gitter (where the rel="collection" link was absolute from a saved self-contained catalog.

pomadchin:
@lossyrob I can parse it via `json.loads` to make it more explicit, do you want me to try that?

lossyrob:
@pomadchin I'm poking around a bit to see if I can trigger it again, will probably use that, but let me take a swing and report back

lossyrob:
This issue is caused by the order of operations when creating the collection/catalogs, and the logic surrounding deciding whether a link HREF should be absolute or relative based on the link owner's root.

When the Item is first added to the Collection, that Collection is set as the Item's root. When that Collection is added to the Catalog, the Catalog is set as the Collection's root, but the Item's root stays set as the Collection. When the save method is called, it changes the Catalog's type to SELF_CONTAINED, which means that any link owned by a STACObject whose root is that Catalog will return a relative HREF. However, as the Item's root is still the Collection, and that Collection's type never changed from the default ABSOLUTE_PUBLISHED, the links owned by that Item returns an absolute href.

I added some logic that, when setting the root of a Catalog, crawls through the resolved links of that Catalog and recursively sets the root as the new root passed in.

This seems to have caused other errors in other tests, so I'll investigate why that is.

tests/utils/test_cases.py
```diff
@@ -242,32 +242,3 @@ def test_case_8() -> Collection:
                 "data-files/catalogs/" "planet-example-v1.0.0-beta.2/collection.json"
             )
         )
-
-    @staticmethod
-    def test_case_9() -> Catalog:
```
lossyrob: Inlined the creation of this catalog as it's specific to that issue, the "test_case_X" methods are meant to represent a set of reusable test Catalogs that are used through many tests.

lossyrob:
Latest commits fix this error - turns out that during the cloning and full_copy process, setting the children and item's root during the parent `set_root` call was causing conflicts. For `clone`, the purpose of calling `set_root(None)` is to clear only the target object's root link, and shouldn't modify the children or items. The logic in `full_copy` recurses the links of the clone itself, and so the new logic in `set_root` on Catalog was interfering with that processing. 

I'd consider this a non-ideal fix, and highlights some of the complexity with root-tracking and link logic. @matthewhanson tagging you to take a look as this is an outcome of the hierarchical link refactor, which itself was to reduce some of the complexity of the previous link work. Maybe you can see a better workaround/solution?

pomadchin:
🚀 thanks!

duckontheweb (approved):
Thanks @pomadchin and @lossyrob for getting this ironed out. I agree that the root-tracking in these hierarchical links probably needs to be reworked to reduce some of the complexity, but this seems like the best solution until we are able to do that.

duckontheweb:
@matthewhanson I'm going to merge this so we can go ahead with a new release tomorrow. If your review turns up a better way to handle this just let me know and we can try to sneak in another update before the release.

PR #663, opened 2021-11-12
Allow PySTAC to skip transforming hrefs to relative or absolute based on CatalogType

lossyrob:
**Related Issue(s):** #546

**Description:**
This PR adds a flag to the `to_dict` methods of STACObjects and ItemCollection, along with Link, to skip the logic that looks to the root catalog type to determine if hierarchical link HREFs should be relative or absolute. 

The current implementation always checks the link's owner for a root catalog, and if it exists and if that root catalog is of type RELATIVE_PUBLISHED or SELF_CONTAINED, and that link is a hierarchical link, then the link's href is transformed to be relative to the link's owner's root catalog.

The default option is to maintain this behavior for the default to_dict method, along with the Catalog, Collection and Item to_dict methods; but this flag now allows users to skip this link href transformation, saving reads.

The ItemCollection.to_dict method was modified to skip link transformation by default. If Items exist in an ItemCollection, the likely desired behavior is not to modify the link hrefs based on the root catalog type, and the result of transforming hrefs in situations where the ItemCollection does not itself have a root catalog results in a root link resolution for each Item, which can be very slow.

This addresses the performance issue raised in https://github.com/stac-utils/pystac/issues/546#issuecomment-966730630

This PR also:

- Passes the ItemCollection "clone_items" flag to the "preserve_dict" parameter of the resulting pystac.Item.from_dict(...) call in the case where the user passes in Dicts instead of Items. This makes the "clone_items" meaningful in both cases, and allows that flag to enable performance gains where the user doesn't want to re-use the Dicts
- Adds a "MockDefaultStacIO" for tests that want to use a MockStacIO instance for the default StacIO. This required changing the type of the parameter to StacIO.set_default from `Type[StacIO]` to `Callable[[], StacIO]`, which I believe is a completely non-breaking change.

**PR Checklist:**

- [ ] Documentation has been updated to reflect changes, if applicable _I updated docstrings, but let me know if you think this belongs in the main docs - I didn't find a good place for it and it seems low enough level that IMO API docs should be sufficient_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Added Table-extension ([#646](https://github.com/stac-utils/pystac/pull/646))
 - Stable support for Python 3.10 ([#656](https://github.com/stac-utils/pystac/pull/656))
 - `.python-version` files are now ignored by Git ([#647](https://github.com/stac-utils/pystac/pull/647))
+- Added a flag to allow users to skip transforming hierarchical link HREFs based on root catalog type ([#663](https://github.com/stac-utils/pystac/pull/663))
 
 ### Removed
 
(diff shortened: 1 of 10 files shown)


duckontheweb (approved):
Thanks @lossyrob, this looks good to me. I verified that the change in type of the `StacIO.set_default` parameter does not raise any `mypy` errors when used as described [in the docs](https://pystac.readthedocs.io/en/latest/concepts.html#i-o-in-pystac), so I agree it is probably not a breaking change.

PR #692, opened 2021-12-22
Update Landsat tutorial

### LINKED ISSUE #665: Documentation issues - Quickstart gives errors, and tutorial use of .ext fails
I've been trying to get familiar with PySTAC recently, and have found a number of issues with the documentation.

**Quickstart**
Running the code in the quickstart guide at https://pystac.readthedocs.io/en/1.0/quickstart.html gives a number of errors. For example, the first code block runs `from pystac import STAC_IO`, which fails as it should be `from pystac import StacIO`. A bit further down, the call to `cat.get_children()` fails with a `TypeError: list indices must be integers or slices, not str` (I couldn't work out how to fix this). I suspect there may be other issues too.

**Tutorials**
In the _Creating a Landsat STAC_ tutorial (https://pystac.readthedocs.io/en/1.0/tutorials/creating-a-landsat-stac.html#Adding-the-EO-extension), it tells you to run `item.ext.enable('eo')` but this fails, saying that the `ext` attribute doesn't exist. After asking this question on the STAC Gitter chat, I was told to follow the instructions [here](https://github.com/stac-utils/pystac/blob/main/docs/concepts.rst#adding-an-extension), which worked.

(As an aside, I'd be interested as to why the `.ext` accessor was removed, as it seems to be a nicer way of adding extensions)

duckontheweb:
**Related Issue(s):**

- Closes #665

**Description:**

Updates the Landsat tutorial to be up-to-date with the current extensions and StacIO APIs.

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Serialization error when using `pystac.extensions.file.MappingObject` ([#700](https://github.com/stac-utils/pystac/pull/700))
 - Use `PropertiesExtension._get_property` to properly set return type in `TableExtension` ([#712](https://github.com/stac-utils/pystac/pull/712))
 - `DatacubeExtension.variables` now has a setter ([#699](https://github.com/stac-utils/pystac/pull/699)])
+- Landsat STAC tutorial is now up-to-date with all package changes ([#692](https://github.com/stac-utils/pystac/pull/674))
 
 ### Deprecated
 
--- a/docs/tutorials/creating-a-landsat-stac.ipynb
+++ b/docs/tutorials/creating-a-landsat-stac.ipynb
@@ -25,7 +25,10 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "import pystac"
+    "import pystac\n",
+    "from pystac.extensions.eo import EOExtension\n",
+    "from pystac.extensions.projection import ProjectionExtension\n",
+    "from pystac.extensions.view import ViewExtension"
    ]
   },
   {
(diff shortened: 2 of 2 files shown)


duckontheweb:
cc: @robintw

gadomski (changes requested):
So it's not touched by this PR right now, but there's some typing issues, e.g. in the "Set the providers" block:

Screen Shot 2022-01-14 at 8 35 22 AM

Probably worth cleaning those up as well since we're hitting this file? Alternately, is there an equivalent Python typing way to do Rust's `Into`/`From`, since `ProviderRole` is trivially convertable to/from a string?

duckontheweb:
> So it's not touched by this PR right now, but there's some typing issues, e.g. in the "Set the providers" block:
> 
> Screen Shot 2022-01-14 at 8 35 22 AM
> 
> Probably worth cleaning those up as well since we're hitting this file? Alternately, is there an equivalent Python typing way to do Rust's `Into`/`From`, since `ProviderRole` is trivially convertable to/from a string?

@gadomski I fixed the type hinting in 7822724261edb0d04dd5ab15891f1a1e1f215b04 and checked it using `nbqa mypy` (only for this tutorial). I ended up adding type hints to the functions as well as a `TypedDict` to describe the `scene` dictionaries that are used throughout the tutorial in order to satisfy the type checker and hopeful make things more clear. Let me know if you think all of those changes are worthwhile.

As a side note, we may want to add some `nbqa` checks to our tutorials as part of our test suite to catch these issues across the tutorials.

gadomski (approved):
Thanks, LGTM. This is one of those cases where type-checking can get in the way a little bit, but I think making sure that the notebooks match the "corrent" way of using the API will save end users headaches down the road.

👍🏽 to adding `nbqa` to CI.

lossyrob (approved):

PR #700, opened 2022-01-06
Store file:values as list of dicts in Item properties

### LINKED ISSUE #605: pystac.extensions.file.MappingObject doesn't work
`pystac.extensions.file.MappingObject` is broken and/or the docs should not specify to use it. An example:

```python
from pystac.extensions.file import FileExtension, MappingObject

FILE_SIZE = 100
VALUES = {
    1: "Temperate or sub-polar needleleaf forest",
    2: "Sub-polar taiga needleleaf forest",
    3: "Tropical or sub-tropical broadleaf evergreen forest",
...
}

values: List[MappingObject] = [
    MappingObject.create(values=[i], summary=s) for i, s in VALUES.items()
]

file_ext = FileExtension.ext(data_asset, add_if_missing=True)
file_ext.apply(size=FILE_SIZE, values=values)
```
results in
```
TypeError: Object of type MappingObject is not JSON serializable
```
when saving the STAC. The following change is a viable workaround:
```python
values: List[Any] = [
    dict(values=[i], summary=s) for i, s in VALUES.items()
]
```

duckontheweb:
**Related Issue(s):**

- Closes #605 

**Description:**

Converts `List[MappingObject]` to `List[Dict[str, Any]]` in the `FileExtension.values` setter. Also adds `from_dict` and `to_dict` methods to the `MappingObject` class.

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 - Quickstart tutorial is now up-to-date with all package changes ([#674](https://github.com/stac-utils/pystac/pull/674))
 - Creating absolute URLs from absolute URLs ([#697](https://github.com/stac-utils/pystac/pull/697)])
+- Serialization error when using `pystac.extensions.file.MappingObject` ([#700](https://github.com/stac-utils/pystac/pull/700))
 
 ### Deprecated
 
--- a/pystac/extensions/file.py
+++ b/pystac/extensions/file.py
@@ -13,7 +13,7 @@ from pystac.serialization.identify import (
     STACJSONDescription,
     STACVersionID,
 )
-from pystac.utils import StringEnum, get_required
+from pystac.utils import StringEnum, get_required, map_opt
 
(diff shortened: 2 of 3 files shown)


gadomski (commented):
👍🏽 to this fix, I just ran into this myself.

A more general question: since we're essentially rolling our own (de)serialization library here, is it worth making an explicit class (or framework) for a (De)Serializable object? There can be other little classes deep in extensions (like this) that it could be nice to not have to re-implement this logic.

duckontheweb:
> A more general question: since we're essentially rolling our own (de)serialization library here, is it worth making an explicit class (or framework) for a (De)Serializable object? There can be other little classes deep in extensions (like this) that it could be nice to not have to re-implement this logic.

Yeah, I think that would be useful. I started to play around with this a bit in #617, but haven't given it much attention lately. I think it would be useful to decouple the JSON serialization/deserialization from the links logic in `STACObject` so that we could re-use it in more places as you suggest.

gadomski (approved):

PR #704, opened 2022-01-07
Use only search link supporting the json mimeType

mneagul:
**Related Issue(s):** 
N/A

**Description:**
pystac allways selects the first search link without taking into consideration the mimeType. For STAC catalogs offering multiple search links with different mimeTypes this might result in erroneous selection of the search link  (eg. in case the first search link points to an endpoint returning `text/html` or a different represenation)

**PR Checklist:**


duckontheweb:
@mneagul Thanks for submitting this. 

I think that changing `get_single_link` to only return JSON links is too restrictive since the `type` field is [not required for Links](https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#link-object) and this method could conceivably be used by users or in downstream code to get a non-JSON link. I believe the test failure we're seeing here is due to the first case (a missing `type` parameter in one of the links), but I have not verified that.

However, I think it would be useful to add an optional `media_type` argument to `get_single_link` that only returns links with the given `type` value if `media_type` is not `None` and maintains the current behavior of only filtering on `rel` if `media_type` is `None`. If the `media_type` argument defaulted to `None`, then this should be backward compatible but still allow users to filter by `type` if needed.

duckontheweb:
Regarding how this applies specifically to `search` links...

The `search` link is defined in the [STAC API spec](https://github.com/radiantearth/stac-api-spec/tree/master/item-search#link-relations) and this library has mostly limited its scope to concepts and constructs from the core STAC Spec and the STAC Extensions hosted in [STAC Extensions](https://stac-extensions.github.io/) organization. If we add the `media_type` argument to `get_single_link` then it might also be worth a PR on [`pystac-client`](https://github.com/stac-utils/pystac-client) to ensure it is using that argument to filter the links where appropriate. 

It might also be worth adding a `media_type` argument to the `get_links` method so that users can filter multiple links in the same way.

duckontheweb (commented):
@mneagul Could you run the test & lint checks by running `./scripts/test` at the top level of the repo? I'm getting a few formatting and type hint errors when I run that.

It also looks like there is a conflict in `CHANGELOG.md` so you will probably need to rebase your branch onto `stac-utils/main`.

Thanks!

duckontheweb:
@lossyrob @gadomski Any idea why the CI wouldn't be running for this (it did when the PR was originally opened)? I don't see an option to approve the workflow like I normally do for new contributors even after removing the required status checks in the branch protection settings.

lossyrob:
@duckontheweb I see "This branch has conflicts that must be resolved" so looks like it needs to merge upstream?

mneagul:
I created a new PR in pystack-client: https://github.com/stac-utils/pystac-client/pull/142 using the functionality added in this PR.

duckontheweb (commented):
@mneagul Thanks for contributing this, looks good! Lets add a couple of tests to the [`CatalogTest` class in `tests/test_catalog.py`](https://github.com/stac-utils/pystac/blob/2d175d11223738778b54dfd37451134b0c291d44/tests/test_catalog.py#L76) that use the new `media_type` argument in `get_links` and `get_single_link` and then I think this is ready to merge.

mneagul:
I added tests for both functions, including an sample catalog.

tests/test_catalog.py
```diff
@@ -1218,6 +1218,26 @@ def test_get_all_collections(self) -> None:
...
+
+        self.assertEqual(
+            catalog.get_single_link("search").href, "./search.html"  # type: ignore
+        )
+        self.assertEqual(
+            catalog.get_single_link(
+                "search", media_type="application/geo+json"
+            ).href,  # type: ignore
+            "./search.json",
+        )
```
duckontheweb, line 1232:
```suggestion
        html_link = catalog.get_single_link("search")
        assert html_link is not None
        self.assertEqual(html_link.href, "./search.html")

        geojson_link = catalog.get_single_link(
            "search", media_type="application/geo+json"
        )
        assert geojson_link is not None
        self.assertEqual(
            geojson_link.href,
            "./search.json",
        )
```

These type errors are coming up because there is a possibility that `catalog.get_single_link` could return `None`. Rather than ignoring the errors, we've typically been using `assert` statements to verify that the result is not `None.

tests/utils/test_cases.py
```diff
@@ -242,3 +242,10 @@ def test_case_8() -> Collection:
                 "data-files/catalogs/" "planet-example-v1.0.0-beta.2/collection.json"
             )
         )
+
+    @staticmethod
+    def test_case_9() -> Catalog:
+        """Example catalog exposed by GeoServer"""
+        return Catalog.from_file(
+            TestCases.get_path("data-files/catalogs/test-case-9/catalog.json")
+        )
```
duckontheweb, line 251:
```suggestion
```

I think it might be best to just use an existing test case and add a new link within the test method (see suggestion above). I'm not 100% sure because these pre-date me working on the project, but I think the test cases are mostly for tests where we want to make sure something works on a few different representative examples (like [here](https://github.com/stac-utils/pystac/blob/2d175d11223738778b54dfd37451134b0c291d44/tests/test_catalog.py#L248)), so I don't think we want to add new cases just for a single test (thoughts @lossyrob?).

tests/test_catalog.py
```diff
@@ -1218,6 +1218,26 @@ def test_get_all_collections(self) -> None:
         self.assertGreater(len(all_collections), 0)
         self.assertTrue(all(isinstance(c, pystac.Collection) for c in all_collections))
 
+    def test_get_single_links_media_type(self) -> None:
+        catalog = TestCases.test_case_9()
```
duckontheweb, line 1222:
```suggestion
        catalog = TestCases.test_case_1()

        catalog.links.append(
            pystac.Link(rel="search", target="./search.html", media_type="text/html")
        )
        catalog.links.append(
            pystac.Link(
                rel="search", target="./search.json", media_type="application/geo+json"
            )
        )
```
See comment below about manually adding the links instead of creating a new test case.

duckontheweb (changes requested):
Sorry, @mneagul, I feel like I'm putting you through the ringer with this PR...

I left some suggested changes and comments regarding creating a new test case, but the changes look good otherwise.

tests/test_catalog.py
mneagul: Thank you :) I missed that the assert would fix the type validation. I completely agree with you change suggestions..

tests/test_catalog.py
```diff
@@ -1218,6 +1218,32 @@ def test_get_all_collections(self) -> None:
...
+        html_link = catalog.get_single_link("search")
+        assert html_link is not None
+        self.assertEqual(html_link.href, "./search.html")
+        json_link = catalog.get_single_link("search", media_type="application/geo+json")
+        assert json_link is not None
+        self.assertEqual(json_link.href, "./search.json")
+
+    def test_get_links_media_type(self) -> None:
+        catalog = TestCases.test_case_9()
+        self.assertEqual(len(catalog.get_links("search")), 2)
```
duckontheweb, line 1242:
```suggestion
        catalog = TestCases.test_case_1()

        catalog.links.append(
            pystac.Link(rel="search", target="./search.html", media_type="text/html")
        )
        catalog.links.append(
            pystac.Link(
                rel="search", target="./search.json", media_type="application/geo+json"
            )
        )
```

This should fix the CI error. We probably don't need to assert then length of the links if we are adding them manually here.

mneagul:
I was unaware of the interaction between `mypy` and the `assert` behaviour :D I apologies for this :D

pystac/stac_object.py
```diff
@@ -81,21 +81,40 @@ def remove_links(self, rel: Union[str, pystac.RelType]) -> None:
 
         self.links = [link for link in self.links if link.rel != rel]
 
-    def get_single_link(self, rel: Union[str, pystac.RelType]) -> Optional[Link]:
+    def get_single_link(
+        self,
+        rel: Union[str, pystac.RelType],
+        media_type: Optional[Union[str, pystac.MediaType]] = None,
+    ) -> Optional[Link]:
         """Get single link that match the given ``rel``.
```
duckontheweb, line 89: Sorry I didn't catch this earlier, but we should update the docstring here as well.
```suggestion
        """Get single link that match the given ``rel`` and, optionally,
        ``media_type``. If ``media_type`` is ``None``, then the link is
        matched only on the ``rel`` value.
```

tests/data-files/catalogs/test-case-9/catalog.json
```diff
@@ -0,0 +1,27 @@
+{
```
duckontheweb, line 1: Let's remove this file as well since it's no longer needed.

duckontheweb:
@mneagul Once `tests/data-files/catalogs/test-case-9/catalog.json` has been removed this should be ready to approve and merge.

PR #687, opened 2021-12-19
Reformat docs and move to PyData sphinx theme

duckontheweb:
**Related Issue(s):**

- Closes #635

**Description:**

This changes the Sphinx theme for our docs from `alabaster` to the [PyData Sphinx Theme](https://pydata-sphinx-theme.readthedocs.io/en/latest/index.html) as suggested in #635. Additionally, it makes some changes to the layout and structure of the docs to (hopefully) make them a bit more user-friendly, especially for newcomers. The layout changes were based heavily on the [Bokeh documentation](https://docs.bokeh.org/en/latest/). Those docs are one of the examples for the PyData Sphinx Theme, and to my eye they provide a really clear landing page that helps direct different audiences to the right place in the docs.

I had also always found the API docs to be hard to navigate, especially as the codebase grew larger. I reworked those docs so that the API Reference has a landing page with cards organized by topics (either related to STAC or PySTAC) and a table of contents organized by module.

~I am opening as a Draft while I get the docs to build on Sphinx as an example, and would love to get some feedback on the theme change, the high-level changes to the structure, and the API Reference changes. Updating some of the tutorials that are still out of date (e.g. #602) will be covered in separate PRs.~

Preview of the new docs is available at https://pystac--687.org.readthedocs.build/en/687/.

**PR Checklist:**

- [x] Code is formatted (run `pre-commit run --all-files`)
- [x] Tests pass (run `scripts/test`)
- [x] Documentation has been updated to reflect changes, if applicable
- [x] This PR maintains or improves overall codebase code coverage.
- [x] Changes are added to the [CHANGELOG](https://github.com/stac-utils/pystac/blob/main/CHANGELOG.md). See [the docs](https://pystac.readthedocs.io/en/latest/contributing.html#changelog) for information about adding to the changelog.

PR #705, opened 2022-01-07
Add instance attribute annotations to all classes

### LINKED ISSUE #321: Create type annotations for attributes

duckontheweb:
**Related Issue(s):**

- Closes #321 

**Description:**

Adds type annotations for instance attributes for all classes. Where the class docstring had an `Attributes` section, I removed those and added a docstring on each type annotation.

**PR Checklist:**

### DIFF
--- a/pystac/cache.py
+++ b/pystac/cache.py
@@ -59,6 +59,18 @@ class ResolvedObjectCache:
             to collections.
     """
 
+    id_keys_to_objects: Dict[str, "STACObject_Type"]
+    """Existing cache of a key made up of the STACObject and it's parents IDs mapped
+    to the cached STACObject."""
+
+    hrefs_to_objects: Dict[str, "STACObject_Type"]
+    """STAC Object HREFs matched to their cached object."""
+
+    ids_to_collections: Dict[str, "Collection_Type"]
+    """Map of collection IDs to collections."""
+
+    _collection_cache: Optional["ResolvedObjectCollectionCache"]
+
     def __init__(
         self,
         id_keys_to_objects: Optional[Dict[str, "STACObject_Type"]] = None,
@@ -69,7 +81,7 @@ class ResolvedObjectCache:
         self.hrefs_to_objects = hrefs_to_objects or {}
         self.ids_to_collections = ids_to_collections or {}
 
-        self._collection_cache: Optional[ResolvedObjectCollectionCache] = None
+        self._collection_cache = None
 
     def get_or_cache(self, obj: "STACObject_Type") -> "STACObject_Type":
         """Gets the STACObject that is the cached version of the given STACObject; or, if
@@ -233,6 +245,9 @@ class CollectionCache:
     in common properties.
     """
 
+    cached_ids: Dict[str, Union["Collection_Type", Dict[str, Any]]]
+    cached_hrefs: Dict[str, Union["Collection_Type", Dict[str, Any]]]
+
     def __init__(
         self,
         cached_ids: Optional[
@@ -274,6 +289,9 @@ class CollectionCache:
 
 
 class ResolvedObjectCollectionCache(CollectionCache):
+
+    resolved_object_cache: ResolvedObjectCache
+
     def __init__(
         self,
         resolved_object_cache: ResolvedObjectCache,
--- a/pystac/extensions/file.py
+++ b/pystac/extensions/file.py
@@ -3,7 +3,7 @@
 https://github.com/stac-extensions/file
 """
 
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, Iterable, List, Optional, Union
 
 import pystac
 from pystac.extensions.base import ExtensionManagementMixin, PropertiesExtension
@@ -98,6 +98,16 @@ class FileExtension(
        >>> asset: pystac.Asset = ...
        >>> file_ext = FileExtension.ext(asset)
     """
+
+    asset_href: str
+    """The ``href`` value of the :class:`~pystac.Asset` being extended."""
+
+    properties: Dict[str, Any]
+    """The :class:`~pystac.Asset` fields, including extension properties."""
+
+    additional_read_properties: Optional[Iterable[Dict[str, Any]]] = None
+    """If present, this will be a list containing 1 dictionary representing the
+    properties of the owning :class:`~pystac.Item`."""
 
     def __init__(self, asset: pystac.Asset):
         self.asset_href = asset.href
--- a/pystac/extensions/pointcloud.py
+++ b/pystac/extensions/pointcloud.py
@@ -2,7 +2,7 @@
 
 https://github.com/stac-extensions/pointcloud
 """
-from typing import Any, Dict, Generic, List, Optional, TypeVar, cast, Union
+from typing import Any, Dict, Iterable, Generic, List, Optional, TypeVar, cast, Union
 
 import pystac
 from pystac.extensions.base import (
@@ -53,6 +53,8 @@ class Schema:
     Use :meth:`Schema.create` to create a new instance of ``Schema`` from
     properties.
     """
+
+    properties: Dict[str, Any]
 
     def __init__(self, properties: Dict[str, Any]) -> None:
         self.properties = properties
@@ -132,6 +134,8 @@ class Statistic:
 
     Use :meth:`Statistic.create` to create a new instance of
     ``Statistic`` from property values."""
+
+    properties: Dict[str, Any]
 
     def __init__(self, properties: Dict[str, Any]) -> None:
         self.properties = properties
@@ -475,6 +479,9 @@ class ItemPointcloudExtension(PointcloudExtension[pystac.Item]):
     :meth:`PointcloudExtension.ext` on an :class:`~pystac.Item` to extend it.
     """
 
+    item: pystac.Item
+    properties: Dict[str, Any]
+
     def __init__(self, item: pystac.Item):
         self.item = item
         self.properties = item.properties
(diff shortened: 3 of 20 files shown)


duckontheweb:
@l0b0 Since you've done so much work on type hints in this library it would be great to get your input as well.

pystac/extensions/file.py
```diff
@@ -99,6 +99,16 @@ class FileExtension(
        >>> file_ext = FileExtension.ext(asset)
     """
 
+    asset_href: str
+    """The ``href`` value of the :class:`~pystac.Asset` being extended."""
+
+    properties: Dict[str, Any]
+    """The :class:`~pystac.Asset` fields, including extension properties."""
+
+    additional_read_properties: Optional[Iterable[Dict[str, Any]]] = None
```
schwehr, line 108: Why is this Iterable?

pystac/layout.py
```diff
@@ -85,7 +95,7 @@ def __init__(
         self.defaults = defaults or {}
 
         # Generate list of template vars
-        template_vars: List[str] = []
+        template_vars = []
```
schwehr, line 98: Was this annotation a problem?

pystac/summaries.py
```diff
@@ -58,6 +58,12 @@ def __lt__(self: "T", other: "T") -> bool:
 
 
 class RangeSummary(Generic[T]):
+    minimum: T
+    """The minimum range value."""
```
schwehr, line 62: These doc strings seem redundant.  If the project style requires them, then definitely leave them.  Otherwise consider dropping them.  Without them, the code is more compact and to me just as readable.  But I'm fine with either way.

pystac/summaries.py
```diff
@@ -194,16 +202,21 @@ def summarize(
...
+    lists: Dict[str, List[Any]]
+    other: Dict[str, Any]
+    ranges: Dict[str, RangeSummary[Any]]
+    schemas: Dict[str, Dict[str, Any]]
+
     def __init__(
         self, summaries: Dict[str, Any], maxcount: int = DEFAULT_MAXCOUNT
     ) -> None:
         self._summaries = summaries
         self.maxcount = maxcount
```
schwehr, line 214: These should go in the class properties, yes?

schwehr (changes requested):
Thanks for working on this!

pystac/layout.py
duckontheweb: This was actually a mistake. I was trying to remove annotations that were now redundant with the instance variables annotations, but inadvertently removed this as well.

However, this annotation may not be necessary since `mypy` has no problem inferring the type.

pystac/extensions/file.py
duckontheweb: Good question, I'm actually not sure where that came from. I'll change that to `Optional[Dict[str, Any]]` for all of the extensions.

pystac/summaries.py
duckontheweb: The intention was to get these to show up in the API docs, but I now realize that we don't include `RangeSummary` in the API docs anyway. I'll remove these docstrings and add the `RangeSummary` classeto the API docs with an `:undoc-members` option so that these get included in the class docs.

pystac/summaries.py
duckontheweb: I'm not sure I follow what you are suggesting? Could you maybe provide a suggested change or a code snippet to illustrate what you are thinking?

pystac/cache.py
```diff
@@ -59,6 +59,18 @@ class ResolvedObjectCache:
             to collections.
     """
 
+    id_keys_to_objects: Dict[str, "STACObject_Type"]
+    """Existing cache of a key made up of the STACObject and it's parents IDs mapped
```
l0b0, line 63: Should we [use `typing.Annotated` here](https://stackoverflow.com/a/8820636/96588)?

pystac/extensions/pointcloud.py
```diff
@@ -492,6 +499,16 @@ class AssetPointcloudExtension(PointcloudExtension[pystac.Asset]):
     :meth:`PointcloudExtension.ext` on an :class:`~pystac.Asset` to extend it.
     """
 
+    asset_href: str
+    """The ``href`` value of the :class:`~pystac.Asset` being extended."""
+
+    properties: Dict[str, Any]
+    """The :class:`~pystac.Asset` fields, including extension properties."""
+
+    additional_read_properties: Optional[Iterable[Dict[str, Any]]] = None
```
l0b0, line 508: [Ditto](https://github.com/stac-utils/pystac/pull/705/files#r780809983).

pystac/item.py
```diff
@@ -49,33 +49,49 @@ class Item(STACObject):
...
+    :attr:`~pystac.CommonMetadata.end_datetime` in :attr:`~pystac.Item.common_metadata`
+    will supply the datetime range of the Item."""
+
+    extra_fields: Dict[str, Any]
+    """Extra fields that are part of the top-level JSON fields the Item."""
+
+    geometry: Optional[Dict[str, Any]]
+    """Defines the full footprint of the asset represented by this item, formatted
+    according to `RFC 7946, section 3.1 (GeoJSON)
+    <https://tools.ietf.org/html/rfc7946>`_."""
```
l0b0, line 80: Since this is referring to a known standard, would it be possible to lock it down more than `Any`? I'm not familiar enough with GeoJSON to tell whether that'd be a lot of work, though.

pystac/layout.py
l0b0: Getting rid of redundant annotations is good. I approve of this message. :grin:

pystac/summaries.py
l0b0: @schwehr Do you mean like `lists: Dict[str, List[Any]]` above? If so, I don't think it's necessary. Mypy knows the type of `maxcount`, so it will infer the type of `self.maxcount`. It can't do the same for empty lists and dicts, which is why they need to be declared at the class level.

schwehr: Think of the humans 😀.  Also, just because there is a type in the init, doesn't mean the type never changes in the class attribute.  The reader shouldn't assume

pystac/extensions/file.py
duckontheweb: Actually, it looks like this was introduced prior to this PR, probably as part of #309. @lossyrob Maybe you can clarify what the intent was here. The goal of `additional_read_properties` seems to be to fall back to reading a property from the owning Item in the case where an Asset does not have that property. However, it seems like this could be accomplished by typing this as `Optional[Dict[str, Any]]` and doing something like `self.additional_read_properties = item.properties` instead of `self.additional_read_properties = [item.properties]`. Are there any cases where we would want to pass in more than 1 dictionary for this?

pystac/summaries.py
duckontheweb: Okay, I see now. Yes, @schwehr my intention here was to have any instance attributes be typed at the class level. As you said, even if it's not strictly required by `mypy`, I do think it's helpful as documentation.

pystac/layout.py
duckontheweb: @schwehr Unless this is causing problems with other type checkers I'm going to continue to leave this annotation out.

pystac/extensions/file.py
lossyrob: This was to account for the situation where you'd want to have multiple levels of property fallbacks - e.g. if there was something on the Asset, Item or Collection summary, or some set of defaults. I don't believe it covers a current need, but I wanted to future proof it a bit given that we've had this situation in the past (both Item properties and Collection properties (rip) were able to define values)

pystac/item.py
lossyrob: The full type signature is a mess: something like `Dict[str, Union[List[float], List[Union[List[float], float]]]]` - I'm not sure that goes deep enough for the unions. IMO json dicts should generally be classified as `Dict[str, Any]` including geojson.

duckontheweb: Yeah, the [Rust `geojson` library](https://docs.rs/geojson/0.22.2/geojson/enum.Value.html) has some interesting recursive typing to handle GeoJSON that we might be able to borrow, but I agree that it's probably not worth it. It seems like it would only be a useful check if users are using a similar level of specificity when typing their inputs. I think it's more likely that users will be using `Dict[str, Any]` to type any GeoJSON they use here, and making this too specific might lead to typing errors for them.

pystac/summaries.py
duckontheweb: Fixed in 152d951918a3952d868ee208f03f86da60763ceb

pystac/item.py
l0b0: @lossyrob That type isn't so bad - it'll give programmers a lot more information than `Dict[str, Any]`. But if the type ends up being recursive we might as well give up, because Mypy doesn't support recursive types yet.

pystac/cache.py
duckontheweb: #687 adds the classes from the `cache` submodule to the API docs, so when this PR has merged with those changes these attributes will already appear in the docs. Since it seems like the main purpose of `typing.Annotated` is to attach documentation to variables it doesn't seem like we would gain much by using it here and it would make the type annotations more complicated. Would there be other advantages to using `typing.Annotated` that I'm missing?

l0b0: Sounds good; I expected `typing.Annotated`'s main purpose was exactly that. I've not used it myself, I'm basically just a fan of explicitly tying the code and documentation together rather than using comments :)

pystac/item.py
duckontheweb: Actually, to accommodate a basic MultiPolygon we would need `Dict[str, Union[str, List[float], List[List[float]], List[List[List[float]]], List[List[List[List[float]]]]]]`. I haven't tried it out, but I think we would need a recursive definition to accommodate GeometryCollections since GeometryCollections have a `"geometry"` field that is an array of GeoJSON Geometries. The GeoJSON spec also does not prohibit other fields in addition to `"type"`, `"coordinates"`, and `"geometry"`, which means we would probably need something more like `Dict[str, Union[str, Any, List[float], List[List[float]], List[List[List[float]]], List[List[List[List[float]]]]]]`, which seems like it is starting to be more confusing than helpful.
My feeling is that it is more helpful to link to the spec and if people want a more rigorous check they can use the validation features.

schwehr (approved):

pystac/extensions/pointcloud.py
duckontheweb: Resolved via [comment above](https://github.com/stac-utils/pystac/pull/705#discussion_r784457060)

Issue #469, opened 2021-06-22
Support pathlike objects in read_file, from_file


TomAugspurger:
https://www.python.org/dev/peps/pep-0519/ describes the file system path protocol. Objects like `pathlib.Path` implement `__fspath__()` to get their string representation of the path. It's nice for libraries to accept pathlike objects anywhere they accept a path.

We can do this by calling `os.fspath(href)` on the argument passed to `read_file` / `from_file`.

```python
In [1]: import pathlib

In [2]: import pystac

In [3]: p = pathlib.Path("tests/data-files/examples/1.0.0/catalog.json")

In [4]: pystac.read_file(p)  # this currently raises.
```

duckontheweb:
Closed via #728

PR #728, opened 2022-01-24
Accept path-like HREF arguments in I/O methods

duckontheweb:
**Related Issue(s):**

- #469 

**Description:**

- Implements the [`os.PathLike`](https://docs.python.org/3/library/os.html#os.PathLike) interface for `pystac.Link` objects. This implementation always returns a `str` representation (never `bytes`). See [PEP 519](https://www.python.org/dev/peps/pep-0519/) for details on the file system path protocol.
- Updates the package-level `pystac.read_file` & `pystac.write_file` methods to accept `os.PathLike` objects for their HREF arguments in addition to strings.
- Updates the `StacIO` I/O methods to accept `os.PathLike` objects for their HREF arguments.
  
   This represents a breaking change in terms of typing since users who are creating custom `StacIO` classes will need to update the type hints in these methods as well. However, this is not a breaking change in terms of runtime behavior, so I am inclined to include it in a minor version release.

@TomAugspurger I'd like to get your feedback on the overall approach, the idea of implementing `os.PathLike` for Links, and whether we should go farther and accept `os.PathLike` objects in more methods that work with HREFs (e.g. `Catalog.save`, etc.).

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 - Experimental support for Python 3.11 ([#731](https://github.com/stac-utils/pystac/pull/731))
+- Accept PathLike objects in `StacIO` I/O methods, `pystac.read_file` and `pystac.write_file` ([#728](https://github.com/stac-utils/pystac/pull/728))
 
 ### Removed
 
--- a/pystac/__init__.py
+++ b/pystac/__init__.py
@@ -62,7 +62,7 @@ from pystac.media_type import MediaType
 from pystac.rel_type import RelType
 from pystac.stac_io import StacIO
 from pystac.stac_object import STACObject, STACObjectType
-from pystac.link import Link, HIERARCHICAL_LINKS
+from pystac.link import Link, HIERARCHICAL_LINKS, HREF
 from pystac.catalog import Catalog, CatalogType
 from pystac.collection import (
     Collection,
@@ -76,7 +76,6 @@ from pystac.asset import Asset
 from pystac.item import Item
 from pystac.item_collection import ItemCollection
 from pystac.provider import ProviderRole, Provider
-from pystac.types import HREF
 import pystac.validation
 
 import pystac.extensions.hooks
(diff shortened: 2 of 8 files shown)


gadomski (approved):
No issues from me.

Only minor comment is that `Union[str, "os.PathLike[AnyStr]"]` ends up everywhere and is a lot to parse for a casual reader, is it worth type aliasing that to `Href` or something?

duckontheweb:
> No issues from me.
> 
> Only minor comment is that `Union[str, "os.PathLike[AnyStr]"]` ends up everywhere and is a lot to parse for a casual reader, is it worth type aliasing that to `Href` or something?

Yeah, I agree that's pretty hard to interpret. Maybe we should have a `types.py` module where we can keep some PySTAC-/STAC-specific types that we use throughout the library.

TomAugspurger:
Looks good to me to.

I vaguely recall a discussion around whether `__fspath__` was intended to represent a "proper" filesystem, or whether it could be used for potentially remote urls like we're doing here for Links. But IMO it's ok. If you do something like `open(item.links[0])`, where the link happens to be remote, you're still going to get an error.

duckontheweb:
> > No issues from me.
> > Only minor comment is that `Union[str, "os.PathLike[AnyStr]"]` ends up everywhere and is a lot to parse for a casual reader, is it worth type aliasing that to `Href` or something?
> 
> Yeah, I agree that's pretty hard to interpret. Maybe we should have a `types.py` module where we can keep some PySTAC-/STAC-specific types that we use throughout the library.

I created a `pystac.types` module that currently has the `HREF` type alias (from this PR) and the `ItemLike` type alias that used to be in `pystac.item_collection`. I think we could probably do some more work on creating more readable type aliases and moving them into `pystac.types`, but that seems better left for a separate PR.

@l0b0 Do you have any thoughts on the approach of storing type aliases in a single module?

duckontheweb:
> > @l0b0 Do you have any thoughts on the approach of storing type aliases in a single module?
> 
> I'd treat that the same as utility functions, and refer to [my answer to _Is putting general-use functions in a "helpers" file an anti-pattern or code smell?_](https://softwareengineering.stackexchange.com/a/422932/13162) 😁

Thanks. It does feel like overkill to create a separate `types` sub-module for just 2 type aliases, especially since 1 of them is only used by the `ItemCollection` class. I'm going to move `ItemLike` back where it came from and put `HREF` into the `link` sub-module since that seems like the most logical place for it.

PR #743, opened 2022-02-03
Optional HTML representations for Jupyter Notebook

DahnJ:
**Related Issue(s):** #508 

## Description
This is a second attempt (after #573) on providing rich HTML representations for (Jupyter) notebooks.

There are still many areas that should/could be improved, but I'm submitting the PR now for feedback. Any feedback is most welcomed, I'd be happy to align this with expectations and requirements from the PySTAC developers/community.

Some areas of possible improvement in this or a subsequent PR
- Catalog vs Collection (some fields such as `extent` are missing from collections since the template simply uses `extra_fields`)
- Additional templates / formatting, e.g. for providers, properties
- Providing some support for extensions (e.g. would be nice to have pretty-formatted `eo:bands` )

### Examples

To reproduce the examples, use the example [catalog](https://github.com/stac-utils/pystac/tree/main/docs/example-catalog)

```python
## Execute in a notebook
import pystac
cat = pystac.Catalog.from_file('./pystac/docs/example-catalog/catalog.json')
cat
```

#### Catalog shows children
image

#### Collection
image

#### Item with Assets
image

### Changes from #573 
Based on #573 with the following improvements:
- HTML representation is now completely optional. If Jinja isn't installed, `repr` is used.
- Catalogs show their children
- Items show their assets
- Templates refactored and improved upon

## PR Checklist


pystac/html/jinja_env.py
```diff
@@ -0,0 +1,10 @@
+jinja_env = None
+
+try:
+    from jinja2 import Environment, PackageLoader, select_autoescape
+
+    jinja_env = Environment(
```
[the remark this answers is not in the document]
DahnJ: Good point. How about wrapping all of this in a cached function? Something like:

```python
from functools import lru_cache

@lru_cache()
def get_jinja_environment():
    try:
        from jinja2 import Environment, PackageLoader, select_autoescape

        return Environment(
            loader=PackageLoader("pystac", "html"), autoescape=select_autoescape()
        )
    except ModuleNotFoundError:
        return None
```

And then only importing and using the `get_jinja_environment` function?

pystac/catalog.py
```diff
@@ -196,6 +198,13 @@ def __init__(
     def __repr__(self) -> str:
         return "<Catalog id={}>".format(self.id)
 
+    def _repr_html_(self) -> str:
+        if jinja_env:
+            template = jinja_env.get_template("Catalog.jinja2")
+            return str(template.render(catalog=self, catalog_type="Catalog"))
```
duckontheweb, line 204: I think we should rename the `catalog_type` parameter here to something else. "catalog type" in other parts of the code base refers to the values defined in the [`CatalogType` enum](https://github.com/stac-utils/pystac/blob/dfc1cc2102cdfc06b952104f9b2c801e5ee17c3b/pystac/catalog.py#L44). I would also be open to just having separate templates to make it a little more clear how the templates map to the PySTAC classes.

pystac/html/Asset.jinja2
```diff
@@ -0,0 +1,31 @@
+{% import 'Macros.jinja2' as macros %}
```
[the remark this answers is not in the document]
duckontheweb: We already have a `package_data` argument in `setup.py` to include the `py.typed` file, so that would just need to be updated to include these files as well. Something as general as `package_data={"": ["py.typed", "*.jinja2"]}` would probably be fine.

pystac/html/Catalog.jinja2
```diff
@@ -0,0 +1,26 @@
+{% import 'Macros.jinja2' as macros %}
+
+<div class="jp-RenderedHTMLCommon jp-RenderedHTML jp-mod-trusted jp-OutputArea-output">
+    {{ macros.square() }}
+    <div>
+        <details style="margin-left: 48px;">
+            <summary style="margin-bottom: 20px;">
+                <h3 style="margin-bottom: 0px; display: inline;">{{catalog_type}}: {{catalog.id}}</h3>
```
duckontheweb, line 8:
```suggestion
                <h3 style="margin-bottom: 0px; display: inline;">{{catalog.__class__.__name__}}: {{catalog.id}}</h3>
```

Should we maybe get the class name directly from the instance here? This would avoid the necessity for the `catalog_type` argument (see my other comment), and would also mean that classes inheriting from `Catalog` or `Collection` in downstream libraries (e.g. [`pystac_client.CollectionClient`](https://github.com/stac-utils/pystac-client/blob/47812c6f59e8f81956e62f2149d4becc34e7aae0/pystac_client/collection_client.py#L10)) would be represented by their true class name.

pystac/html/Macros.jinja2
```diff
@@ -0,0 +1,91 @@
...
+{%- endmacro %}
+
+
+{% macro items(parent) -%}
+    {% if parent.get_items()|list|length > 0 %}
+        <details>
+            <summary style="margin-bottom: 20px;">
+                <h4 style="margin-bottom: 0px; display: inline;">Items</h4>
+            </summary>
+            {% for item in parent.get_items() %}
```
duckontheweb, line 42: A call to `get_items` is going to resolve all Items for this catalog (not recursively) by making a network request to each `"item"` link in the catalog, which could take a pretty long time. I think we probably want to avoid that here and maybe just summarize the item link information instead (i.e. using `get_item_links` and summarizing the title, href, rel type, etc.), even though it does not provide as much detail.

Also, [in `pystac_client.CollectionClient`](https://github.com/stac-utils/pystac-client/blob/47812c6f59e8f81956e62f2149d4becc34e7aae0/pystac_client/collection_client.py#L14), which inherits from `pystac.Collection`, this will result in multiple paginated requests to the API to fetch all Items, which could take a *very* long time. That is more of a concern for that library, but I thought it was worth mentioning (cc: @matthewhanson, @gadomski).

pystac/html/Macros.jinja2
```diff
@@ -0,0 +1,91 @@
...
+                {{ item._repr_html_() }}
+            {% endfor %}
+        </details>
+    {% endif %}
+{%- endmacro %}
+
+
+{% macro children(parent) -%}
+    {% if parent.get_children()|list|length > 0 %}
```
duckontheweb, line 51: Similar to comment above, this is going to result in (possibly) multiple network requests to resolve each child as a `STACObject`. We may want to use `get_child_links` and summarize the link info instead.

pystac/html/jinja_env.py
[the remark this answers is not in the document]
duckontheweb: I'm +1 for using a cached function to retrieve this as-needed.

duckontheweb (commented):
Thanks for this @DahnJ! I really like the way it's organized; the templates are very understandable, and it's easy to trace a PySTAC class to its template.

Regarding some of the possible "points of improvement" in your description...

> * Catalog vs Collection (some fields such as extent are missing from collections since the template simply uses extra_fields)

See my review comments, but I think it makes sense to break these into separate templates. I think that would make it easier to handle some of the extra fields in `Collection` (like `extent`).

> * Additional templates / formatting, e.g. for providers, properties
> * Providing some support for extensions (e.g. would be nice to have pretty-formatted eo:bands )

I agree it would be good to flesh out some of the other classes like `Provider` and the extension objects. 

Should we also add a `jinja2` extra [here](https://github.com/stac-utils/pystac/blob/dfc1cc2102cdfc06b952104f9b2c801e5ee17c3b/setup.py#L30)? In general, users would probably be using this functionality in a Jupyter notebook or other environment where `jinja2` is already installed, but it might be nice to give them the option to install it as an extra as well.

pystac/html/jinja_env.py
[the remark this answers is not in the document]
DahnJ: I will go with the cached function approach. I will use `# type: ignore` since I can't use `Optional[Environment]` as the class hasn't been imported at function definition time.

pystac/catalog.py
DahnJ: I agree, I will go with separating the templates

pystac/html/Macros.jinja2
DahnJ: I see. I'm torn here: the fact that I can actually click through the items/children is why I wanted to write this in the first place.

Could you please give me an example of loading a STAC catalog over a network / internet, so I could test this out? I only found examples using `Catalog.from_file()` in the documentation.

duckontheweb: > I see. I'm torn here: the fact that I can actually click through the items/children is why I wanted to write this in the first place.

I agree, I'm torn as well because there is not much information in the links. Could we maybe put a limit on the number of Items that are fetched using something like: `itertools.islice(parent.get_items(), 100)` and give some visual indication that we are only showing the first 100 (or whatever number we choose) items? That way users could get a feel for the Items in the catalog while still limiting the number of requests.

> **edit**: I see that simply passing URL seems to work to some extent, e.g.
> 
> ```python
> cat = pystac.Catalog.from_file('https://cbers-stac-1-0.s3.amazonaws.com/CBERS4/catalog.json')
> ```
> 
> But for example `next(cat.get_children())` hangs forever for me, which I'm not sure if supposed to happen.

I'm not sure what's going on there; that code is running for me in a second or two:

```python
>>> import pystac
>>> cat = pystac.Catalog.from_file('https://cbers-stac-1-0.s3.amazonaws.com/CBERS4/catalog.json')
>>> next(cat.get_children())
<Collection id=CBERS4-MUX>
```

Is the issue persistent for you?

DahnJ: > Is the issue persistent for you?

Aha, I think it hangs only when I call `_repr_html_()`, otherwise it works fine. So I now have an example to try :+1: 

> I agree, I'm torn as well because there is not much information in the links. Could we maybe put a limit on the number of Items that are fetched using something like: `itertools.islice(parent.get_items(), 100)` and give some visual indication that we are only showing the first 100 (or whatever number we choose) items?

Yes, I like that, I will try it out. 

I'm a bit worried that my "hangs forever" issue is probably because the templates get constructed by fetching all items/children immediately and recursively. If that's the case, then even a limit on the number of fetched items could be too simple. 

I will look into it, looking into some lazy loading options (e.g. there [seem to be some HTML mechanisms](https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading)

DahnJ: Here is what I tried

# Loading only the first child/item
For the `CBERS` dataset, this took ~2.2s to render. Note that it's fairly nested (4 levels of nesting to get to an item). Two seconds seems reasonable to me, especially since that only happens the first time.

Displaying any additional items increases the time exponentially and is IMO not feasible. Loading two children took ~10s.

### Example
```python
import pystac
cat = pystac.Catalog.from_file('https://cbers-stac-1-0.s3.amazonaws.com/CBERS4/catalog.json')
```

### Preview
image

# Lazy loading
I looked into the pure HTML/CSS options for lazy loading, but didn't find anything useful.
- HTML: `loading="lazy"` only seems to be available for `img` and `iframe`
- CSS: `content-visibility` still loads the data immediately, only controls visibility

DahnJ: I don't see a better solution for now. Showing only the first entry is a reasonable solution IMO.

duckontheweb: Yeah, I agree this seems like the best solution for now.

duckontheweb:
@DahnJ Sorry for the long silence on this, I'm just coming back to this now. I just updated the branch to reflect the latest changes to `main`. This looks good to me after the change to only list the first Item in the Catalog/Collection summaries. There is a lint error (`black` is reformatting one of the files), but aside from that it looks ready to me.

pystac/html/jinja_env.py
```diff
@@ -0,0 +1,21 @@
...
+        )
+
+        environment.filters["first"] = lambda x: islice(x, 1)
+        environment.filters["is_nonempty_generator"] = (
+            lambda x: next(x, None) is not None
+        )
+
+        return environment
+    except ModuleNotFoundError:
+        return None
```
gadomski, line 21: This try block could be more narrow, correct?
```suggestion
    try:
        from jinja2 import Environment, PackageLoader, select_autoescape
    except ModuleNotFoundError:
        return None
    environment = Environment(
        loader=PackageLoader("pystac", "html"), autoescape=select_autoescape()
    )

    environment.filters["first"] = lambda x: islice(x, 1)
    environment.filters["is_nonempty_generator"] = (
        lambda x: next(x, None) is not None
    )

    return environment
```

TomAugspurger:
Lazy question: does this handle `pystac.ItemCollection` too? If not, it doesn't need to hold up this PR but I think it'd be great to do as a followup.

Untitled.ipynb
```diff
@@ -0,0 +1,6 @@
+{
+ "cells": [],
+ "metadata": {},
+ "nbformat": 4,
+ "nbformat_minor": 5
+}
```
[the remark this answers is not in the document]
DahnJ: No, removed

pystac/html/jinja_env.py
DahnJ: Correct, changed in https://github.com/stac-utils/pystac/pull/743/commits/07042aac10ef257ec0d84976985ca4d0907712d0

DahnJ:
> It looks like we'll also need to configure `mypy` to ignore `jinja2` imports as well (similar to [this line](https://github.com/stac-utils/pystac/blob/e5d010f94cf718a94995ea5883325faa820f50c4/mypy.ini#L8-L9) for `setuptools`.

Done in https://github.com/stac-utils/pystac/pull/743/commits/6cf3fa4f0d70a80df5e2bc8eade85e63022f4327

gadomski (approved):

DahnJ:
> Lazy question: does this handle `pystac.ItemCollection` too? If not, it doesn't need to hold up this PR but I think it'd be great to do as a followup.

I will still have a quick look at this (beyond the fact that `ItemCollection` inherits from `Collection`) tomorrow, but after that this is good to go from my side :+1:

DahnJ:
> Lazy question: does this handle `pystac.ItemCollection` too? If not, it doesn't need to hold up this PR but I think it'd be great to do as a follow-up.

I have implemented a rudimentary template for `ItemCollection` in https://github.com/stac-utils/pystac/pull/743/commits/4328a772688c49434fed92d57aca5f6dc1e0cd98 which simply lists items and `extra_fields` (if present). To my understanding, there's not much more to `ItemCollection` than that.

This is what it looks like

image


However, I have shown **all** the items. If I understood it correctly, `ItemCollection` is simply a GeoJSON `FeatureCollection` and thus cannot end up fetching the items over the network? If I misunderstood, then I can simply show the first Item as with `Catalog`/`Collection`.

cc: @duckontheweb

duckontheweb:
> If I understood it correctly, `ItemCollection` is simply a GeoJSON `FeatureCollection` and thus cannot end up fetching the items over the network?

Yes, this is correct. These Items will already be resolved in memory.

> We might want to limit the ItemCollection repr to some number of items (maybe 10?). It's common to have an ItemCollection with hundreds or thousands of items.

I agree with @TomAugspurger on this, I think it would be good to limit the number of Items displayed.

DahnJ:
> > We might want to limit the ItemCollection repr to some number of items (maybe 10?). It's common to have an ItemCollection with hundreds or thousands of items.
> 
> I agree with @TomAugspurger on this, I think it would be good to limit the number of Items displayed.

Okay, I pushed https://github.com/stac-utils/pystac/pull/743/commits/2137ff0f1a27658a8e204839163415f6708ee7c3 with the changes:

If there are `<=10` items, the behaviour stays the same:

image


But if there are over 10 items, only the first 10 are shown with a helpful message:

image

TomAugspurger (approved):
Looks great, thanks!

duckontheweb (approved):
Thanks for sticking with this one @DahnJ, this is a pretty big feature add to the library!

PR #744, opened 2022-02-03
Use type alias for temporal intervals

### LINKED ISSUE #586: Type issues with TemporalExtent
I'm getting errors from mypy when I try to create `TemporalExtent`s:

```
error: Argument 1 to "TemporalExtent" has incompatible type "List[List[datetime]]"; expected "Union[List[List[Optional[datetime]]], List[Optional[datetime]]]"
```

I'm creating them like this:

```
pystac.TemporalExtent([[start_datetime`, end_datetime]])
pystac.TemporalExtent([[start_datetime` or None, end_datetime or None]])
```

The first line throws the `mypy` error, the second does not. Am I missing something obvious? Is this just a fault in `mypy`?

`# type: Optional[datetime]` can also be applied to the variables.

duckontheweb:
**Related Issue(s):**

- Closes #586 

**Description:**

Adds 2 type variables that are used in the `TemporalExtent` class:

- `TemporalIntervals` - This is used for properly formatted [STAC temporal intervals](https://github.com/radiantearth/stac-spec/blob/master/collection-spec/collection-spec.md#temporal-extent-object) (i.e. a list of lists)
- `TemporalIntervalsLike` - This is used for the `intervals` argument to `TemporalExtent` where we allow either a list of lists or a list of optional datetimes and convert it into a list of lists.

I tried using a covariant `TypeVar` as I had initially suggested in [#586 (comment)](https://github.com/stac-utils/pystac/issues/586#issuecomment-889199716), but that did not fix the type errors. Using a `Union` type of all the possible combinations seemed to be the only way to make `mypy` happy.

Also note that this adds a test whose only purpose is type checking when we run `mypy`. If there are suggestions on a better place to put this kind of check I'm happy to change that.

**PR Checklist:**

### DIFF
--- a/pystac/collection.py
+++ b/pystac/collection.py
@@ -38,7 +38,9 @@ if TYPE_CHECKING:
 
 T = TypeVar("T")
 TemporalIntervals = Union[List[List[datetime]], List[List[Optional[datetime]]]]
-TemporalIntervalsLike = Union[TemporalIntervals, List[datetime], List[List[datetime]]]
+TemporalIntervalsLike = Union[
+    TemporalIntervals, List[datetime], List[Optional[datetime]]
+]
 
 
 class SpatialExtent:
--- a/pystac/collection.py
+++ b/pystac/collection.py
@@ -37,6 +37,8 @@ if TYPE_CHECKING:
     from pystac.provider import Provider as Provider_Type
 
 T = TypeVar("T")
+TemporalIntervals = Union[List[List[datetime]], List[List[Optional[datetime]]]]
+TemporalIntervalsLike = Union[TemporalIntervals, List[datetime], List[List[datetime]]]
 
 
 class SpatialExtent:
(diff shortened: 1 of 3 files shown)


duckontheweb:
cc: @cuttlefish

pystac/collection.py
```diff
@@ -37,6 +37,8 @@
     from pystac.provider import Provider as Provider_Type
 
 T = TypeVar("T")
+TemporalIntervals = Union[List[List[datetime]], List[List[Optional[datetime]]]]
+TemporalIntervalsLike = Union[TemporalIntervals, List[datetime], List[List[datetime]]]
```
gadomski, line 41: I'm probably just not understanding the typing property, but this looks like `List[List[datetime]]` is duplicated by this union?

gadomski (approved):
I did have one question about the type union but I'm probably not understanding it correctly so approving so I don't block if I'm wrong :-).

pystac/collection.py
duckontheweb: Oh shoot, you're right. I meant that to be:

```python
TemporalIntervalsLike = Union[TemporalIntervals, List[datetime], List[Optional[datetime]]]
```

duckontheweb: Fixed in the latest commit.

gadomski (approved):

Issue #448, opened 2021-06-16
Handling STAC extension version upgrades


duckontheweb:
Moving here from [`stac-spec` Discussions thread](https://github.com/radiantearth/stac-spec/discussions/1145).

Now that STAC extension versions are no longer tied to the core STAC version or to each other, we may run into situations where a user is working with catalogs that implement some combination of a core spec version and extension versions that are not all supported by the same version of PySTAC.

Here are a few ideas on how to handle this, but other suggestions are welcome. It would also be good to get feedback from maintainers of downstream libraries on what would work best.

### 1. Separate Extension Repos

STAC extension implementations would each have their own repo that would follow its own versioning (e.g. `pystac-extension-label`, `pystac-extension-file`, etc.). This would probably also require creating a `pystac-extensions-base` package that would be a dependency of each extension implementation.

**Pros**

* More accurately reflects the structure of the spec with extensions separate from the core spec
* Developers can mix and match any combination of extension versions they need
* No need to upgrade PySTAC version with each extension version update
* Simpler PySTAC core code base

**Cons**

* Means a lot more repos in `stac-utils` org
* Could lead to dependency conflicts if different extension implementations are using different versions of `pystac-extensions-base`
* Developers need to explicitly install each extension they want to use (or maybe we have an `all-extensions` extra in the `pystac` install that grabs all of them...)
* Sounds like a lot of work and may delay a 1.0 release

### 2. Version-specific Extension Classes

Instead of a single set of extension classes (e.g. `EOExtension`, `ItemEOExtension`, `AssetEOExtension`), we create a set of classes for each extension version (e.g. `EOExtension_1_0_0`, etc.). We still have a top-level extension class (e.g. `EOExtension`) with an `ext` method that would handle parsing the extension schema URI and delegating to the right version-specific class.

**Pros**

* Fewer repos to maintain
* Developers will continue to get extension support within PySTAC without needing to install other extension-specific dependencies
* May be able to limit the outward-facing changes if the behavior of `ext` stays mostly the same as it is now
* Might be able to implement this incrementally (i.e. only for extensions that have actually undergone an upgrade)?

**Cons**

* More complex extension code base within PySTAC (may need to create version-specific modules within each extension to keep the files from getting too bloated).
* Typing may be difficult and/or confusing (e.g. calling `EOExtension.ext` and getting an instance of `EOExtension_1_0_0` back might be non-intuitive).

### 3. Only Support Latest Version (Status Quo)

[...]

PR #748, opened 2022-02-10
Add performance benchmarks

### LINKED ISSUE #729: Establish performance testing framework and benchmarks
A few issues have come up over the past few months related to slow performance within the library. We do not currently have any benchmarks for memory usage or runtimes built into our testing suite, so it is hard to catch performance regressions or evaluate possible performance improvements.

The goal of this issue is to articulate a plan for a first pass at performance benchmarking within the library. This should include deciding which parts of the code we want to benchmark and selecting a performance benchmarking library (or libraries) to use in our testing suite.

I do not have extensive personal experience in this area, so input from others who have set up performance testing frameworks in the past is highly desired. To start the discussion, here are some initial thoughts:

## What to Benchmark?

- [`tests.test_writing.py::STACWritingTest::test_testcases`](https://github.com/stac-utils/pystac/blob/d659185aca360b9d6adc6d329d90d5320269da03/tests/test_writing.py#L114)
  
  This could be a good test of the round-trip of reading, copying, and writing a catalog for each of our test catalogs.

- `to_dict` methods
  
  #546 brought up some performance issues in the use of `to_dict` and [this comment](https://github.com/stac-utils/pystac/issues/546#issuecomment-911050197) raises some questions about how we cache STAC objects. It would be good to have some benchmarks here so we can fine-tune this in terms of both memory and speed.

- Methods for fetching items/children in catalogs
  #99 brings up the issue of efficiently fetching specific items from a catalog and it would be good to have some metrics around this

## What Tool to Use?

Given the interest in using async and multithreaded techniques to speed up performance (see #609 and #274) any library we choose should probably be able to handle asynchronous and multithreaded code. Here are some options from a brief survey of the landscape:

- [`asv` (airspeed velocity)](https://asv.readthedocs.io/en/stable/)

  Well-documented and supported speed benchmarking framework with built-in visualization capabilities. Not sure what it would look like to support asynchronous or multi-threaded code.

- [`yappi`](https://github.com/sumerc/yappi/)

  Has support for `async` and `gevent` and multithreaded processing, seems well-supported, and claims to be very fast. Only profiles timing, not memory usage (as far as I can tell).

- [`profile`/`cProfile`](https://docs.python.org/3/library/profile.html#)
 
  Already included in the standard library, but would take some work to make it support multi-threaded or async code. Only profiles timing, not memory usage. Can be combined with [`snakeviz`](https://jiffyclub.github.io/snakeviz/) for visualizing results.

- [`guppy3`](https://github.com/zhuyifei1999/guppy3/)

  Memory profiler with some [related blog posts](https://github.com/zhuyifei1999/guppy3/#usage) on how to track down memory issues.

cc: @TomAugspurger @lossyrob @matthewhanson @gadomski @scottyhq

duckontheweb:
**Related Issue(s):**

- Closes #729 

**Description:**

Uses the [airspeed velocity (`asv`)](https://asv.readthedocs.io/en/stable/index.html) library to run performance benchmarks. This configuration generally follows the strategy used by `xarray` (as documented [here](https://github.com/pydata/xarray/blob/main/asv_bench/benchmarks/README_CI.md)), with some notable exceptions:

- Increase the number of rounds to 4 (from default of 2) and the number of repetitions to between 10 & 50 (from default of 5). This seemed to eliminate a lot of the false positives with regard to performance changes. These values are set in the [`Bench` base class](https://github.com/duckontheweb/pystac/blob/c187569e0d9b32b7f49ca11080e3f93aaa374578/benchmarks/benchmarks/_base.py) and can be changed for a particular benchmark.
- Use the `--interleave-rounds` option (also seems to help with eliminating false positives)
- Run benchmarks on all supported Python versions (we may want to change this later as the benchmark suite grows and takes longer to run).

This is a work-in-progress, but I want to get feedback on the approach before fleshing out the rest.

Still to do:

- [x] Add timing benchmarks for extension implementations (maybe not all, but at least a few)
- [x] Add timing and peak memory benchmarks for reading and walking a catalog
- [x] Add timing benchmark for writing large catalogs to disk (e.g. on the order of thousands of items)
- [ ] Come up with a strategy for measuring possible performance improvements from implementing async requests. We will probably see the most impact from this when making network requests rather than local file reads, but we don't want our benchmarks dependent upon changing network latency.
- [x] Add documentation

**PR Checklist:**


duckontheweb:
@TomAugspurger Would be great to get your feedback as well.

duckontheweb:
> * Establish performance testing framework and benchmarks #729

Thanks @gadomski, I have a bit of a list started in #729. I like both of these suggestions, I'll add them to the PR.

gadomski:
I picked this PR up and, in the interest of not letting perfect be the enemy of good, implemented _most_ of @duckontheweb's suggestions:

- I added benchmarks for reading, walking, and writing large catalogs, including peak memory usage for walking a large catalog
- I added a single extension benchmark, for the projection extension. I'm not quite sure what we want to be benchmarking there, so I kept it minimal.
- I refactored the location of things to make usage a bit simpler (e.g. I want `asv dev` to work out-of-the-box)
- I removed CI benchmarking. To paraphrase @TomAugspurger from <https://tomaugspurger.github.io/posts/performance-regressions>, "This rules out running the benchmarks on ... CI  ... even if we could finish it in time, we couldn’t really trust the results." It's my personal opinion that benchmarking on CI is so finicky and untrustworthy that its not worth the complexity. In liu of a dedicated benchmark machine, IMO it's better to run benchmarks locally when the situation calls for it. At least we have a framework for benchmarking now, which we could always expand in the future to a dedicated machine if its needed.
- Docs.
- Simple check in CI to make sure benchmarks run w/o failure -- doesn't do any reporting.

I didn't implement:

- Any sort of async strategy. I think that's best tackled as part of the async improvements themselves (https://github.com/stac-utils/pystac/pull/749) -- at least, I don't think they should hold up getting _some_ sort of benchmarking into the repo.
- Extensive extension benchmarking. I'm not sure where the pain points are right now, and I don't want broken benchmarks to get in the way of solutions to #448.

docs/contributing.rst
```diff
@@ -73,6 +73,44 @@ flag to Git commit commands, as in ``git commit --no-verify``.
...
    when modifying the return type of a function used in another file.
 
+Benchmarks
+^^^^^^^^^^
+
+PySTAC uses `asv <https://asv.readthedocs.io>`_ for benchmarking. Benchmarks are
+defined in the ``./benchmarks`` directory. Due to the inherent uncertainty in
+the environment of Github workflow runners, benchmarks are not executed in CI.
+If your changes may affect performance, we recommend checking your benchmarks
+locally using the provided script:
```
pjhartzell, line 83: Might be worth being a little more explicit about what `scripts/bench` is doing. Something like:

>If your changes may affect performance, we recommend locally checking your benchmarks against the main branch benchmarks using the provided script.

That may need some wordsmithing.

benchmarks/import_pystac.py
```diff
@@ -0,0 +1,14 @@
...
+    repeat = 10
+
+    def setup(self, *args: Any, **kwargs: Any) -> None:
+        def import_pystac() -> None:
+            import pystac  # noqa: F401
+
+        self._import_pystac = import_pystac
+
+    def time_import_pystac(self) -> None:
+        self._import_pystac()
```
pjhartzell, line 14: Per the asv docs, I think this should be using [raw timing](https://asv.readthedocs.io/en/stable/writing_benchmarks.html#raw-timing-benchmarks).

pjhartzell (changes requested):
Looking good. This is a nice addition.

benchmarks/import_pystac.py
gadomski: Good catch, thanks.

pjhartzell (approved):
Looks good.

PR #749, opened 2022-02-14
Add async I/O methods [WIP]

### LINKED ISSUE #274: Large catalog async optimizations
I've been playing with using asyncio/httpx/aiofiles/aioboto3 to improve the speed of reading and writing of large catalogs (10+ million). Right now I'm just wrapping around PySTAC but think some of this could be incorporated into it. 

Opening up this ticket to see if anyone else has been working with async functions here.

### LINKED ISSUE #609: Make pystac handle item creation asynchronously
As of version 1.1.0 stac items are handeled in a synchronized way. This is fine for a few items, but has a long runtime with over hundreds of items. 

I use our own LGLN script to build the catalog up from collections and items, calling our own function for the heavy lifting, normalizing and then save the catalog in the end.

With pythons concurrent lib, I was able to fasten up the item creation from tiff files with the ThreadPoolExecutor, but having trouble with using a ProcessPoolExecutor, which would be the right way to do it, cause the task is pretty much just cpu-bound, not I/O-bound in any way.

I can see a 400% increase in item creation speed with 16 processes spawned instead of one, but have to comment out the save method in the end, which then generated nothing.

Unfortunately, the catalog.save method does not wait at all for the catalog.add_item method to return and so tries to build a dictionary from an empty list, which results in an error. 

To circumvent this, it would be really great, if pystac itself internally could handle the creation of items and adding to the catalog asynchronously, maybe with an extra parameter, to be synchronious handling still be the default, because "it just works".

```
def addImageItemsToCatalog(imageList, catalog):
    """multithreaded execution inspired by:
    https://stackoverflow.com/questions/21143162/python-wait-on-all-of-concurrent-futures-threadpoolexecutors-futures
    Performance increase was measured to be around 100%, measured with 93 stac items in 2 different catalogs,
    compared to singlethreaded.
    Could be further increased with multiprocess implementation (e.g. use "ProcessPoolExecutor" instead),
    but a way to have the "catalog.save" method be waiting on the execution has to be found, otherwise
    pyStac will fail.
    """
    executor = concurrent.futures.ProcessPoolExecutor(MULTI_THREAD_SWEET_SPOT) 
    futures = [executor.submit(writeImageFileItems, imageFile, catalog) for imageFile in imageList]
    concurrent.futures.wait(futures, timeout=None, return_when=ALL_COMPLETED)
    # VVV this is done on the main thread, without waiting, which crashes :(
    catalog.normalize_hrefs("generated/dop")
    print("Saving catalog to " + catalog.get_self_href())
    catalog.save(catalog_type=pystac.CatalogType.SELF_CONTAINED)
```

The resulting error would be the following, which I think is clear to me, because .save() does not wait for anything to be in there:
```
  File "C:\dev\repo\task-3d-2.0\prototypes\dopCatalogBuilder\main.py", line 9, in <module>
    main()
  File "C:\dev\repo\task-3d-2.0\prototypes\dopCatalogBuilder\main.py", line 5, in main
    dopCatalogBuilder.build()
  File "C:\dev\repo\task-3d-2.0\prototypes\dopCatalogBuilder\dopCatalogBuilder.py", line 100, in build
    dop_catalog = createCatalogFromAreaDictionary(dopImages, subcatalogs, catalogTitle="test raster Orthophotos",
  File "C:\dev\repo\task-3d-2.0\prototypes\dopCatalogBuilder\dopCatalogBuilder.py", line 114, in createCatalogFromAreaDictionary
    addImageItemsToCatalog(images, dopCatalog)
  File "C:\dev\repo\task-3d-2.0\prototypes\dopCatalogBuilder\dopCatalogBuilder.py", line 133, in addImageItemsToCatalog
    catalog.save(catalog_type=pystac.CatalogType.SELF_CONTAINED)
  File "C:\Users\dev\anaconda3\envs\catalogBuilder\lib\site-packages\pystac\catalog.py", line 753, in save
    child.save()
  File "C:\Users\dev\anaconda3\envs\catalogBuilder\lib\site-packages\pystac\catalog.py", line 783, in save
    self.save_object(
  File "C:\Users\dev\anaconda3\envs\catalogBuilder\lib\site-packages\pystac\stac_object.py", line 341, in save_object
    stac_io.save_json(dest_href, self.to_dict(include_self_link=include_self_link))
  File "C:\Users\dev\anaconda3\envs\catalogBuilder\lib\site-packages\pystac\collection.py", line 520, in to_dict
    d["extent"] = self.extent.to_dict()
AttributeError: 'list' object has no attribute 'to_dict'
```

duckontheweb:
## Related Issue(s)

- Closes #274 (pending additional changes)
- Closes #609 (pending additional changes)

## Description

Implements asynchronous I/O operations by adding `async` equivalents to the following `StacIO` methods:

- `read_text` / `read_text_async` (abstract methods)
- `write_text` / `write_text_async` (abstract methods)
- `read_json` / `read_json_async`
- `save_json` / `save_json_async`
- `read_stac_object` / `read_stac_object_async`

This PR also creates a new `DefaultStacIOAsync` class with default implementations of the async methods using `aiofiles` (for local file I/O) and `httpx` (for reading via HTTP). In order to take advantage of these async methods, users must install PySTAC using the `aiofiles` and `httpx` extras or create their own implementations of the async abstract methods using their libraries of choice (e.g. `aiohttp`).

I added a `Catalog.save_async` method that takes advantage of these new async I/O methods as an example of how me might implement async I/O throughout the code base. Feedback would be appreciated on the approach of having sync methods call their equivalent async method, and on the naming conventions used.

There is still quite a bit of work to do on this before we can consider the linked issues to be closed, but I would like feedback on the changes to the `Catalog` and `StacIO` APIs before I do too much more work on the rest of the code.

*This PR represents a breaking change, as it requires developers to implement the abstract `read_text_async` and `write_text_async` methods in any custom `StacIO` implementations. However, users who are using the default implementation should not need to make any changes.*

## PR Checklist

- [ ] Test performance of synchronous `StacIO` methods calling async methods vs. using separately maintained synchronous code.
- [ ] Implement asynchronous versions of remaining methods that involve I/O operations (e.g. `Catalog.walk`)
- [ ] Update tutorials and other docs to use async methods
- [ ] Merge changes from #748 and add benchmarks to test I/O performance improvements/degradations


duckontheweb:
Changed the base branch to `2.0` since this introduces breaking changes.

gadomski (commented):
Tl;dr: +0

I don't have any strong objections to the way things are currently laid out. It feels a little awkward to have all of the `RuntimeError` checks inside the `save` method, but if that's how it has to be, so it goes.

I'll mention a couple of possible other implementations that popped to mind; my understanding is pretty low for the Python async ecosystem so these are probably Very Wrong™:
- Go "async-first" and make `save` (and friends) async, and use "blocking" to describe non-async methods (e.g. `read_blocking`). This is probably a bad idea because PySTAC is so foundational and we don't want people to have to go async to just use it.
- Use two different classes for Async and Sync objects, e.g. `AsyncItem`. I think this probably breaks a lot of the type inference stuff used to detect item types?
- Is there any way to "store" asyncio information on `StacIO` to avoid all those `try/excepts`? Probably not...just something that I thought of while reading tokio's tutorial on bridging w/ sync code: https://tokio.rs/tokio/topics/bridging.

schwehr:
I've never used asyncio, so I don't have a lot to offer in review.

duckontheweb:
Thanks @gadomski!

> It feels a little awkward to have all of the `RuntimeError` checks inside the `save` method, but if that's how it has to be, so it goes.

I agree. I was going to abstract this into a separate `_run_coroutine` method, but I didn't want to add that complexity until there was consensus on the general approach.

> * Go "async-first" and make `save` (and friends) async, and use "blocking" to describe non-async methods (e.g. `read_blocking`). This is probably a bad idea because PySTAC is so foundational and we don't want people to have to go async to just use it.

I thought about this, too, and was wary for the same reasons. I suspect that the majority of our users are not as familiar with async in Python and might be at least intimidated by seeing the methods they are used to using switched to async.

> * Use two different classes for Async and Sync objects, e.g. `AsyncItem`. I think this probably breaks a lot of the type inference stuff used to detect item types?

I had toyed with this idea a bit, too. I think the general recommendation for async programming is to put sync and async I/O operations in different classes, but as you mentioned that might cause problems with some of our STAC object type sniffing. It also would make it a little trickier to take advantage of async operations within synchronous methods (like [here](https://github.com/duckontheweb/pystac/blob/2187cdff8391831fb089955015bc21c30fe45551/pystac/catalog.py#L767)) for parallelizing operations. 

That said, I'm open to other ways of structuring this. One thought would be to have an `AsyncItem` class as you suggested but only have it contain the I/O operations and make it available within the normal `Item` class (maybe in an `Item.async_io` property). That way our normal STAC object detection methods could still return the normal classes, and we could isolate the async operations to a separate class. If that seems worth pursuing I can put in an alternate PR with that structure.

> * Is there any way to "store" asyncio information on `StacIO` to avoid all those `try/excepts`? Probably not...just something that I thought of while reading tokio's tutorial on bridging w/ sync code: https://tokio.rs/tokio/topics/bridging.

I had originally structured the `StacIO` changes by having a separate `AsyncStacIO` class that was available as `StacIO.async_io` in the normal `StacIO` instances and could be used by `StacIO` to do async operations. I ran into a some problems with making the sync and async methods available within the methods on `STACObject`, `Catalog`, etc., but those might be alleviated if we separate out the async logic as suggested above. This makes me think it might be worth a separate PR to explore structuring the the async code that way so we have a good point of comparison.

sgillies:
When I was updating Planet's stactools plugin I noticed that stactools already depends on aiohttp. Maybe it would be worth considering using that instead of httpx?

PR #760, opened 2022-03-03
Add helper methods for RFC 3339 datetime handling

philvarner:
**Related Issue(s):** n/a

**Description:**

There are a number of places in the STAC ecosystem where datetimes are converted two and from RFC 3339 strings. Many of these places are doing this incorrectly. These changes aim to add a correct implementations to:

- rfc3339_str_to_datetime: parse ONLY a valid RFC 3339 string to a datetime
- str_to_interval: parse an interval string into a datetime tuple
- now_in_utc: create a datetime with UTC set as the datetime (since utcnow() inexplicably does not do this)
- now_to_rfc3339_str: create an RFC3339 string representing now

My primary motivation is to replace the numerous uses of the stac_pydantic constant `DATETIME_RFC339`(sic) with uses of these methods. Other libraries that depend on pystac will also use them, and there are likely several places even within pystac that these could be used instead of the current code.

One downside to this is that the tests need the pytest parameterized decorator to run, and this doesn't work in the class-based test modules that unittest requires. I ran the existing tests with pytest and it seems to run them all fine, so we could change to using pytest to run the test without needing to convert all the tests from unittest to pytest.

**PR Checklist:**

- [X] Code is formatted (run `pre-commit run --all-files`)

### DIFF
--- a/pystac/utils.py
+++ b/pystac/utils.py
@@ -329,12 +329,26 @@ def datetime_to_str(dt: datetime, timespec: str = "auto") -> str:
 def str_to_datetime(s: str) -> datetime:
     """Converts a string timestamp to a :class:`datetime.datetime` instance using
     :meth:`dateutil.parser.parse` under the hood. The input string may be in any
-    format :std:doc:`supported by the parser <parser>`.
+    format :std:doc:`supported by the parser <parser>`. This includes many formats
+    including ISO 8601 and RFC 3339.
 
     Args:
         s (str) : The string to convert to :class:`datetime.datetime`.
+
+    Returns:
+        str: The :class:`datetime.datetime` represented the by the string.
     """
     return dateutil.parser.isoparse(s)
+
+
+def now_in_utc() -> datetime:
+    """Returns a datetime value of now with the UTC timezone applied"""
+    return datetime.now(timezone.utc)
+
+
+def now_to_rfc3339_str() -> str:
+    """Returns an RFC 3339 string representing now"""
+    return datetime_to_str(now_in_utc())
 
 
 def geometry_to_bbox(geometry: Dict[str, Any]) -> List[float]:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
+- Additional util methods `now_in_utc` and `now_to_rfc3339_str` ([#760](https://github.com/stac-utils/pystac/pull/760))
 - Add `media_type` and `role` filtering to Item and Collection `get_assets()` method ([#936](https://github.com/stac-utils/pystac/pull/936))
 - `Asset.has_role` ([#936](https://github.com/stac-utils/pystac/pull/936))
 - Enum MediaType entry for flatgeobuf ([discussion](https://github.com/flatgeobuf/flatgeobuf/discussions/112#discussioncomment-4606721)) ([#938](https://github.com/stac-utils/pystac/pull/938))
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -7,6 +7,7 @@ import unittest
 from datetime import datetime, timedelta, timezone
 from typing import Optional
 
+import pytest
 from dateutil import tz
 
 from pystac import utils
@@ -14,6 +15,8 @@ from pystac.utils import (
     is_absolute_href,
     make_absolute_href,
     make_relative_href,
+    now_in_utc,
+    now_to_rfc3339_str,
     str_to_datetime,
 )
 from tests.utils import TestCases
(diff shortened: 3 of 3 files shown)


philvarner:
It may be more appropriate to put str_to_interval in stac-fastapi, since the datetime interval format is only used by query parameters.

philvarner:
looks like i need to write some stubs for ciso8601

duckontheweb:
> looks like i need to write some stubs for ciso8601

@philvarner I think it would also be fine to just ignore these in the `mypy.ini` file as we've done for [other libraries without stubs](https://github.com/stac-utils/pystac/blob/0692ffeaa870df17fa9f43bc5b15ddbbf98362d7/mypy.ini#L8-L9).

philvarner:
> > looks like i need to write some stubs for ciso8601
> 
> @philvarner I think it would also be fine to just ignore these in the `mypy.ini` file as we've done for [other libraries without stubs](https://github.com/stac-utils/pystac/blob/0692ffeaa870df17fa9f43bc5b15ddbbf98362d7/mypy.ini#L8-L9).

Okay, I'll do that. Weird thing is that it looks like ciso8601 has stubs defined. I haven't used mypy that much, so I don't really know how the stubs are supposed to be resolved, etc.

setup.py
```diff
@@ -25,6 +25,7 @@
     python_requires=">=3.7",
     install_requires=[
         "python-dateutil>=2.7.0",
+        "ciso8601~=2.2.0",
```
duckontheweb, line 28: We've traditionally been *very* stingy about adding new dependencies to PySTAC. My feeling is that this one is worth it, but could we could drop `python-dateutil` if we are introducing `ciso8601`?

@gadomski @lossyrob @schwehr Curious to get your thoughts as well.

duckontheweb:
> > > looks like i need to write some stubs for ciso8601
> > 
> > 
> > @philvarner I think it would also be fine to just ignore these in the `mypy.ini` file as we've done for [other libraries without stubs](https://github.com/stac-utils/pystac/blob/0692ffeaa870df17fa9f43bc5b15ddbbf98362d7/mypy.ini#L8-L9).
> 
> Okay, I'll do that. Weird thing is that it looks like ciso8601 has stubs defined. I haven't used mypy that much, so I don't really know how the stubs are supposed to be resolved, etc.

Yeah, that is peculiar. I'll look into this a bit more and see if there is a better way to solve this besides ignoring the import.

philvarner:
> > > > looks like i need to write some stubs for ciso8601
> > > 
> > > 
> > > @philvarner I think it would also be fine to just ignore these in the `mypy.ini` file as we've done for [other libraries without stubs](https://github.com/stac-utils/pystac/blob/0692ffeaa870df17fa9f43bc5b15ddbbf98362d7/mypy.ini#L8-L9).
> > 
> > 
> > Okay, I'll do that. Weird thing is that it looks like ciso8601 has stubs defined. I haven't used mypy that much, so I don't really know how the stubs are supposed to be resolved, etc.
> 
> Yeah, that is peculiar. I'll look into this a bit more and see if there is a better way to solve this besides ignoring the import.

I needed to add it to the requirements-test.txt and setup.py validation requirements.

setup.py
philvarner: To give a bit of context to this, ciso8601 is the only library I've found in any language that correctly parses RFC 3339 dates. I'd though about replacing python-dateutil instead, but that would likely break existing behavior because the python-dateutil parsing methods are *so* permissive in what they accept -- it's very much an attempt to convert anything that looks like a string representing a datetime into a `datetime.datetime`.

philvarner: Also, ciso8601 introduces no transitive dependencies, as it has none.

schwehr: I am slightly against it as a requirement.  

- [It's C code](https://github.com/closeio/ciso8601/blob/master/module.c)... so that's a weightier dependency to have
- It's not that commonly used
- But at least [it is MIT licensed](https://github.com/closeio/ciso8601/blob/master/LICENSE), so no license concerns from my point-of-view

If there was a way to have it be an optional install that didn't make the testing situation a nightmare, that would be good.

philvarner: I removed `str_to_interval`, since I think that is very specific to api impls and should be in stac-fastapi instead.

philvarner: I've been thinking about this more, and my underlying concern here is that the use of dateutil.parser.parse in `str_to_datetime` (which is used everywhere) is too lenient, and means that many datetime strings that are not even close to being RFC3339 are parsed without error. For example, `dateutil.parser.parse("January 1, 2047 at 8:21:00AM")` returns a valid datetime. Postel's law is relevant here, but as it is, I think pystac is being much too liberal in what it accepts as a datetime string. 

Another option would be to use `dateutil.parser.isoparse` that only parses ISO8601 datetimes. I haven't personally tested this against my edge case strings, but will do that.

philvarner: So there's two issues here:

1. should the existing str_to_datetime be changed to be more strict
2. should a precise RFC3339 method (useful for validation) be included? Should a correct datetime->str method be included for use by downstream users?

gadomski: > Curious to get your thoughts as well.

I'm with @schwehr, if we're going to incorporate a non-wheel-providing dependency that requires a compiler, I'd like to move slowly and ensure there aren't any Python-only alternative solutions.

To address your questions @philvarner, from my perspective:

> should the existing str_to_datetime be changed to be more strict

PySTAC is used a part of production systems (where we it to be conservative) and by data consumers in development/research environments (where we want it to be liberal). I'd be in favor of a opt-in strictness on the consumption side (i.e. `str -> datetime`) to ensure that our users in the science community (e.g.) don't suddenly have to wrangle w/ datetime parsing errors because the data provider they've been using has been Doing It Wrong™.

> should a precise RFC3339 method (useful for validation) be included?

Yes. Validation should be strict as hell right? :-)

> Should a correct datetime->str method be included for use by downstream users?

Yes. Production should be as strict as possible IMO.

> Another option would be to use dateutil.parser.isoparse that only parses ISO8601 datetimes. I haven't personally tested this against my edge case strings, but will do that.

IMO this would be a much more attractive alternative if we can make it work.

setup.py
```diff
@@ -25,9 +25,13 @@
...
         'typing_extensions >= 3.7; python_version < "3.8"',
     ],
-    extras_require={"validation": ["jsonschema>=3.0"], "orjson": ["orjson>=3.5"]},
+    extras_require={
+        "validation": ["jsonschema>=3.0", "pytest"],
```
gadomski, line 32:
```suggestion
        "validation": ["jsonschema>=3.0"],
```

Don't need `pytest` as a distributed dependency.

tests/test_utils2.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def test_now_functions() -> None:
+    now1 = now_in_utc()
+    now2 = now_in_utc()
+
+    assert now1 < now2
+    assert now1.tzinfo == timezone.utc
+
+    rfc3339_str_to_datetime(now_to_rfc3339_str())
```
gadomski, line 74: Could be cool to break this out into its own micropackage to assess various libraries, do you have anything like that already @philvarner? If not, I might have a couple of cycles...

requirements-test.txt
```diff
@@ -13,5 +13,10 @@ types-orjson==3.6.2
 
 pre-commit==2.17.0
 
+pytest~=7.0.1
+pytest-cov~=3.0.0
+
+ciso8601~=2.2.0
+
```
gadomski, line 20:
```suggestion
```

Unnecessary, because already in `setup.py`.

setup.py
lossyrob: IMO accepting datetimes liberally is a feature, not a bug. The purpose of PySTAC is to make working with STAC easier, not leave people scratching their heads on what the exact correct ISO8601 formatted string is to make PySTAC accept their datetime string (which I know I would be). Adding a more strict version of str_to_datetime does seem like a great addition; however like Pete mentions we should for sure not break existing usage of PySTAC in the name of strictness, even if we wanted to (which as I said, I wouldn't be in favor of strictness).

For the write site (datetime_to_str), that should always produce valid datetimes to the standard. Any issues solved around that is a good change. "Lenient on input, strict on output" is a good mantra for this library IMO.

I think there's likely a way to solve this that doesn't include the addition of a C dependency. I think we'd have to weight whatever additional capability against the addition of a dependency, which as others have said we've been pretty careful not to do. From what I can tell the work to fix issues with the datetime_to_str method in a Python-only way would be worth excluding this additional dependency and saving the potential work by the user base for any future issues that dependency may cause.

lossyrob:
> One downside to this is that the tests need the pytest parameterized decorator to run, and this doesn't work in the class-based test modules that unittest requires. I ran the existing tests with pytest and it seems to run them all fine, so we could change to using pytest to run the test without needing to convert all the tests from unittest to pytest.

I'd be happy with a change to pytest personally; however I wonder if that should be a larger PR that does a wholesale conversion rather than have multiple testing frameworks in the test. You could use a `unittest.TestCase` with `self.subTest` to test over each of the cases; for example here: https://github.com/stac-utils/pystac/blob/0426bd1475d5686a74cae33e795a03c033a716ba/tests/validation/test_validate.py#L39-L40

As it stands there's not a change to `scripts/test` to use pytest, so these unit tests wouldn't be run as part of CI.

setup.py
philvarner: 1. rewrote to remove the pytest dependency. I should have just done it that way from the beginning. 
2. changed the new method to use dateutil's ISO8601 parsing method. Another option would be to add a parameter "strict" to `str_to_datetime` that defaults to False for now, and maybe change it to True in a future version with appropriate notification of the the change.

The strictness should probably be somewhere between accepting "January 1, 2047 at 8:21:00AM" and requiring a perfect RFC3339 string.

pystac/utils.py
```diff
@@ -329,14 +329,48 @@ def datetime_to_str(dt: datetime, timespec: str = "auto") -> str:
...
+    Args:
+        s (str) : The string to convert to :class:`datetime.datetime`.
+
+    Returns:
+        datetime: The datetime represented by the ISO8601 (RFC 3339) formatted string.
+
+    Raises:
+        ValueError: If the string is not a valid RFC 3339 string.
     """
     return dateutil.parser.isoparse(s)
```
gadomski, line 361: Without ciso8601, these two functions are the same. I'd suggest removing `iso8601_to_datetime`.

gadomski (changes requested):

setup.py
gadomski: The scope of this PR has been pared down dramatically since this conversation, to the point where it might end up just adding two free helper functions to `utils`. I'm going to consider this conversation useful but resolved.

pjhartzell (approved):
Looks good to me.

philvarner:
Got this merged in just under a year!

Issue #783, opened 2022-03-31
Create Grid Extension from code in stactools sentinel2 package


philvarner:
Grid Extension can be created from the file grid.py in this PR https://github.com/stactools-packages/sentinel2/pull/26

duckontheweb:
Closed by #799

PR #799, opened 2022-04-29
add Grid Extension support

philvarner:
**Related Issue(s):** 

- #783

**Description:**

Add Grid Extension support <https://github.com/stac-extensions/grid/>

**PR Checklist:**

- [X] Code is formatted (run `pre-commit run --all-files`)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,9 @@
 [metadata]
 version = attr: pystac.version.__version__
+
+[tool:pytest]
+minversion = 6.0
+addopts = -ra -q
+testpaths =
+    tests
+asyncio_mode = auto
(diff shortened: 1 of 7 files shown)


setup.cfg
```diff
@@ -1,2 +1,9 @@
 [metadata]
 version = attr: pystac.version.__version__
+
+[tool:pytest]
+minversion = 6.0
+addopts = -ra -q
+testpaths =
+    tests
+asyncio_mode = auto
```
duckontheweb, line 9: I wonder if we should leave these changes for a future PR that more comprehensively adds support for `pytest`. I'm in support of moving to `pytest` in general, but given that we don't include it in `requirements-test.txt` it seems out of place to have configuration for it here.

duckontheweb (commented):
Thanks for this @philvarner . All of the extension-related changes look good to me, I just had a question regarding whether we want to add the `pytest` configuration as part of this PR.

setup.cfg
philvarner: I think this config is all that needs to be included for pytest to work (and not even that, since this just a bit more optimal). I think it's a separate question as to whether we make it the default/recommended way to run the tests.

duckontheweb (approved):

PR #808, opened 2022-05-13
Return None from Asset.get_absolute_href when owning Item self HREF is NOne

### LINKED ISSUE #754: `Asset.get_absolute_href()` should return `None` but doesn't
When STAC item doesn't have a self link and asset `href` is not absolute then I expect `.get_absolute_href()` to return `None`, instead it returns absolute path in the form `{absolute_path_to_the_current_working_directory}/{relative_path_of_the_asset}`.

```python
from datetime import datetime
import pystac

item = pystac.Item("id", None, None, datetime(2022, 2, 22), {})
item.add_asset("b1", pystac.Asset("b1.tiff"))
assert item.get_self_href() is None
assert item.assets["b1"].get_absolute_href() is None
```

Looking at the code it only returns `None` for "detached" assets.
https://github.com/stac-utils/pystac/blob/2d175d11223738778b54dfd37451134b0c291d44/pystac/asset.py#L102-L103

I think code above should be something along these lines:

```python
if self.owner is not None:
   base = self.owner.get_self_href()
   if base is not None:
       return utils.make_absolute_href(self.href, base)
```

duckontheweb:
**Related Issue(s):

* Closes #754 

**Description:**

`Asset.get_absolute_href` now returns `None` when the owning `Item` has no self HREF.

**PR Checklist:**

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -10,6 +10,3 @@ ignore_missing_imports = True
 
 [mypy-setuptools.*]
 ignore_missing_imports = True
-
-[mypy-sphinx.util]
-ignore_missing_imports = True
--- a/pystac/asset.py
+++ b/pystac/asset.py
@@ -92,7 +92,8 @@ class Asset:
         """Gets the absolute href for this asset, if possible.
 
         If this Asset has no associated Item, and the asset HREF is a relative path,
-            this method will return None.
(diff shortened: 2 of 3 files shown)


duckontheweb:
cc: @Kirill888

gadomski (approved):

Issue #789, opened 2022-04-13
Raster schema should be updated from v1.0.0 to v1.1.0


philvarner:

duckontheweb:
Closed by #809 . Thanks @pjhartzell !

PR #809, opened 2022-05-16
update raster extension to v1.1.0

pjhartzell:
**Related Issue(s):** #789 

**Description:**

Makes "nan", "inf", and "-inf" valid values for the `nodata` property in `raster:bands`.

- update schema uri
- align the `nodata` type with the updated schema, which allows for numeric and "nan", "inf", and "-inf" strings.
- update tests and test data. For the test data, the `raster:bands` object added to the `B09` asset is from the sentinel example in the Raster Extension repo.

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@
 
 ### Changed
 
+- Updated Raster Extension from v1.0.0 to v1.1.0 ([#809](https://github.com/stac-utils/pystac/pull/809))
+
 ### Fixed
 
 - "How to create STAC catalogs" tutorial ([#775](https://github.com/stac-utils/pystac/pull/775))
--- a/pystac/extensions/raster.py
+++ b/pystac/extensions/raster.py
@@ -10,7 +10,7 @@ from pystac.extensions.base import (
 )
 from pystac.utils import StringEnum, get_opt, get_required, map_opt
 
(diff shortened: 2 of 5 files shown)


pjhartzell:
Thanks @gadomski for pointing me at #448. Beyond the relevant discussion on how to handle extension updates in the future, it looks like an extension hook should be added to this PR to conform to the current method of handling extension version updates.

pjhartzell:
In this case, the only change between v1.0.0 and v1.1.0 is the allowance of certain strings for the `nodata` value. So updating STAC from v1.0.0 to v1.1.0 only needs the schema URI in the `extensions_list` to change from v1.0.0 to v1.1.0. 

However, [the `migrate` method is only called](https://github.com/stac-utils/pystac/blob/1261e6058e8e9137d5c7c7f16e0d18e888a6d9cf/pystac/extensions/hooks.py#L50) when `stac_version` differs from the current value. So adding an extension hook only migrates the URI if the imported STAC is <1.0.0.

duckontheweb (approved):
@pjhartzell Thanks so much for working on this, these changes all look good to me. Your comment on the migration only needing to update the schema URI is right on, so I think this is ready to be converted from Draft to PR and merged.

duckontheweb:
@pjhartzell Any objection to me converting this from a Draft?

PR #826, opened 2022-06-08
Deepcopy Asset.extra_fields when cloning

### LINKED ISSUE #820: Asset: clone() doesn't clone extra_fields
If you do asset.clone() you get a shallow copy, but as extra_field is explicitily a member of the class, I'd expect that it at least also get cloned.

The following code adds `abc` to both asset and mask, which is somewhat unexpected:
```py
fields = {}
asset = Asset(
    href=href,
    extra_fields=fields,
)
mask = asset.clone()
mask.extra_fields["abc"] = True
```

duckontheweb:
**Related Issue(s):**

- Closes #820 

**Description:**

Uses `deepcopy` on `Asset.extra_fields` when cloning so that mutating the cloned Asset's extra fields does not mutate the original asset.

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 
 ### Fixed
 
+- Mutating `Asset.extra_fields` on a cloned `Asset` also mutated the original asset ([#826](https://github.com/stac-utils/pystac/pull/826))
 - "How to create STAC catalogs" tutorial ([#775](https://github.com/stac-utils/pystac/pull/775))
 - Add a `variables` argument, to accompany `dimensions`, for the `apply` method of stac objects extended with datacube ([#782](https://github.com/stac-utils/pystac/pull/782))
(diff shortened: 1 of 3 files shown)


gadomski (approved):

duckontheweb:
> Thank you @duckontheweb I'd also be fine with a shallow copy of Asset.extra_fields because that is what the documentation says, but also fine with a deep copy if documented properly.

@m-mohr Which documentation are you referring to here? The `Asset.clone` docs  just say `"Clones this asset.`", which seems vague enough to cover both cases. I'll update that to be more specific, but is there another place where it specifically suggests we're doing a shallow copy?

m-mohr:
Oh yeah, sorry, I was confusing this with the description of `clone` for Items

duckontheweb:
I updated the `Asset.clone` docs and added a regression test to be sure `Asset.roles` is also not mutated on clones.

PR #834, opened 2022-06-28
Preserve Collection assets on clone

### LINKED ISSUE #747: Collection assets removed on clone()
Hello! First thanks for this wonderful library, it's been super helpful

I might be misunderstanding, but I think there is an issue with Collections losing their assets when you run `.clone()` because the attribute is not carried over when constructing a new instance. I saw some discussion about placing asset info in the `Collection.extra_fields` attribute (ref https://github.com/stac-utils/pystac/issues/293) but my understanding is that Collections now support having assets (e.g., there's a `Collection.add_asset`)

It looks to me like the asset attribute needs to be included in this call,
https://github.com/stac-utils/pystac/blob/main/pystac/collection.py#L556-L568
but was maybe omitted as the STAC spec evolved.

Example,
```python
>>> import pystac
>>> coln = pystac.read_file("https://planetarycomputer.microsoft.com/api/stac/v1/collections/landsat-8-c2-l2")
# nice thumbnail!
>>> coln.assets
{'thumbnail': <Asset href=https://ai4edatasetspublicassets.blob.core.windows.net/assets/pc_thumbnails/landsat.png>}
# works
>>> coln.to_dict()['assets']
{'thumbnail': {'href': 'https://ai4edatasetspublicassets.blob.core.windows.net/assets/pc_thumbnails/landsat.png', 'type': 'image/png', 'title': 'Landsat 8 C2', 'roles': ['thumbnail']}}
# broken
>>> coln.clone().assets
{}
```

If this makes sense I'm happy to add a PR with a fix and tests

duckontheweb:
**Related Issue(s):**

- Closes #747 

**Description:**

Adds an `assets` argument to `Item.__init__` and `Collection.__init__` that accepts a dictionary mapping strings to `Asset` objects, which allows us to attach Assets to these objects as part of instantiation. 

Also preserves Collection assets when using `Collection.clone` and adds regression test.

cc: @cholden-ag

**PR Checklist:**

### DIFF
--- a/tests/test_collection.py
+++ b/tests/test_collection.py
@@ -271,6 +271,22 @@ class CollectionTest(unittest.TestCase):
         with self.assertRaises(pystac.STACTypeError):
             _ = pystac.Collection.from_dict(catalog_dict)
 
+    def test_clone_preserves_assets(self) -> None:
+        path = TestCases.get_path("data-files/collections/with-assets.json")
+        original_collection = Collection.from_file(path)
+        assert len(original_collection.assets) > 0
+
+        cloned_collection = original_collection.clone()
+
+        original_assets_dict = {
+            k: asset.to_dict() for k, asset in original_collection.assets.items()
+        }
+        cloned_assets_dict = {
+            k: asset.to_dict() for k, asset in cloned_collection.assets.items()
+        }
+
+        self.assertDictEqual(original_assets_dict, cloned_assets_dict)
+
 
 class ExtentTest(unittest.TestCase):
     def setUp(self) -> None:
--- a/tests/test_collection.py
+++ b/tests/test_collection.py
@@ -275,17 +275,20 @@ class CollectionTest(unittest.TestCase):
         path = TestCases.get_path("data-files/collections/with-assets.json")
         original_collection = Collection.from_file(path)
         assert len(original_collection.assets) > 0
+        assert all(
+            asset.owner is original_collection
+            for asset in original_collection.assets.values()
+        )
 
         cloned_collection = original_collection.clone()
 
-        original_assets_dict = {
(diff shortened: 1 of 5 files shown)


cholden-ag:
Sorry to leave the PR to my issue hanging @duckontheweb and thanks for adding the fix. I think this looks good for fixing the `clone()` behavior on Collections, and having `assets` in the constructor will also improve the user experience for creating some Items/Collections since it'll take care of running `add_asset`

tests/test_collection.py
```diff
@@ -271,6 +271,22 @@ def test_from_invalid_dict_raises_exception(self) -> None:
...
+
+        original_assets_dict = {
+            k: asset.to_dict() for k, asset in original_collection.assets.items()
+        }
+        cloned_assets_dict = {
+            k: asset.to_dict() for k, asset in cloned_collection.assets.items()
+        }
+
+        self.assertDictEqual(original_assets_dict, cloned_assets_dict)
+
```
gadomski, line 289: Should probably put a regression test in for `Item` as well?

gadomski (changes requested):

tests/test_collection.py
duckontheweb: We sort of implicitly test that the assets don't get dropped [here](https://github.com/stac-utils/pystac/blob/a6fffa59f4c6b6433858a8569b35647798593976/tests/test_item.py#L224-L232), but it would probably be good to make it more explicit. I can also update the Collection test to check that the owner is set properly to match the Item test.

duckontheweb: Updated both tests to make them match in dcd21c0f0c73b4fdf078e2842e9a94cb2cbb6452.

gadomski (approved):

PR #848, opened 2022-07-18
Use isoparse instead of parse for datetimes

### LINKED ISSUE #819: `pystac.read_file` returns temporal extents with `tzinfo=tzlocal()` depending on the environment.
Depending on the environment (I think the value of `TZ`, maybe other things), the timezones on pystac Collections can be `tzlocal` rather than `utc`.

```python
>>> pystac.read_file("https://planetarycomputer.microsoft.com/api/stac/v1/collections/sentinel-2-l2a").extent.temporal.intervals[0][0].tzinfo
tzlocal()
```

Here's a demonstration of the environment `TZ` influencing the output:

```console
❯ TZ=UTC python -c 'import pystac; print(pystac.read_file("https://planetarycomputer.microsoft.com/api/stac/v1/collections/sentinel-2-l2a").extent.temporal.intervals[0][0].tzinfo)'
tzlocal()

❯ TZ="US/Central" python -c 'import pystac; print(pystac.read_file("https://planetarycomputer.microsoft.com/api/stac/v1/collections/sentinel-2-l2a").extent.temporal.intervals[0][0].tzinfo)'
tzutc()
```

Since the STAC spec says that all datetimes should be UTC, I think that the datetimes should be tz-aware. If python-dateutil happens to return a tzlocal timezone, I think pystac can assume that local = UTC and localize it to UTC.

Upstream issue in python-dateutil: https://github.com/dateutil/dateutil/issues/349

xref https://github.com/stac-utils/pystac/pull/760. I'm not sure if this would fix this behavior or not (cc. @philvarner)

duckontheweb:
**Related Issue(s):**

- Closes #819 

**Description:**

Updates `pystac.utils.str_to_datetime` to use `dateutil.parser.isoparse` instead of `dateutil.parser.parse` to ensure UTC timestamps stay in UTC even if the system timezone is set to UTC as well. There were also a couple of places in the `Item` and `Collection` classes where we were using `dateutil.parser.parse` directly instead of using `pystac.utils.str_to_datetime`. These have all been updated to use `str_to_datetime`.

**PR Checklist:**

### DIFF
--- a/tests/test_collection.py
+++ b/tests/test_collection.py
@@ -4,6 +4,7 @@ import os
 import json
 from datetime import datetime
 from dateutil import tz
+import time
 import tempfile
 
 import pystac
@@ -412,6 +413,46 @@ class ExtentTest(unittest.TestCase):
         )
 
         self.assertDictEqual(extent_dict, extent.to_dict())
+
+    def test_utc_datetimes_are_not_localized(self) -> None:
+        extent_dict = {"interval": [["2015-06-27T10:25:31Z", None]]}
+
+        prev_tz = os.environ.get("TZ")
+
+        # Test without TZ environment variable set
+        with self.subTest(tz=None):
+            if "TZ" in os.environ:
+                del os.environ["TZ"]
+                time.tzset()
+            extent = TemporalExtent.from_dict(extent_dict)
+            first_datetime = extent.intervals[0][0]
+            assert first_datetime is not None
+            self.assertIs(first_datetime.tzinfo, tz.tzutc())
+            self.assertIsNot(first_datetime.tzinfo, tz.tzlocal())
+
+        # Test with TZ environment variable set to UTC
+        with self.subTest(tz="UTC"):
+            os.environ["TZ"] = "UTC"
+            time.tzset()
+            extent = TemporalExtent.from_dict(extent_dict)
+            first_datetime = extent.intervals[0][0]
+            assert first_datetime is not None
+            self.assertIs(first_datetime.tzinfo, tz.tzutc())
+            self.assertIsNot(first_datetime.tzinfo, tz.tzlocal())
+
+        # Test with TZ environment variable set to UTC
+        with self.subTest(tz="US/Central"):
+            os.environ["TZ"] = "US/Central"
+            time.tzset()
+            extent = TemporalExtent.from_dict(extent_dict)
+            first_datetime = extent.intervals[0][0]
+            assert first_datetime is not None
+            self.assertIs(first_datetime.tzinfo, tz.tzutc())
+            self.assertIsNot(first_datetime.tzinfo, tz.tzlocal())
+
+        if prev_tz is not None:
+            os.environ["TZ"] = prev_tz
+            time.tzset()
 
 
 class CollectionSubClassTest(unittest.TestCase):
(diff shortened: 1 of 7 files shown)


tests/test_item.py
```diff
@@ -20,6 +22,46 @@ def get_example_item_dict(self) -> Dict[str, Any]:
...
+        with self.subTest(tz="UTC"):
+            os.environ["TZ"] = "UTC"
+            time.tzset()
+            item = Item.from_dict(item_dict)
+            datetime = item.datetime
+            assert datetime is not None
+            self.assertIs(datetime.tzinfo, tz.tzutc())
+            self.assertIsNot(datetime.tzinfo, tz.tzlocal())
+
+        # Test with TZ environment variable set to UTC
```
gadomski, line 51:
```suggestion
        # Test with TZ environment variable set to US/Central
```

Or just remove the comments, it's clear (to me at least) what's going on here.

tests/test_item.py
```diff
@@ -20,6 +22,46 @@ def get_example_item_dict(self) -> Dict[str, Any]:
             item_dict: Dict[str, Any] = json.load(f)
         return item_dict
 
+    def test_utc_datetimes_are_not_localized(self) -> None:
```
gadomski, line 25: To me, these tests should belong in a `test_utils.py::test_str_to_datetime` -- there's no item-specific functionality being checked, correct? This item test would then check the obvious regression case (tz="UTC"). That's just my test layout preference, though, not a required change (i.e. feel free to ignore).

tests/test_collection.py
```diff
@@ -413,6 +414,46 @@ def test_to_from_dict(self) -> None:
 
         self.assertDictEqual(extent_dict, extent.to_dict())
 
+    def test_utc_datetimes_are_not_localized(self) -> None:
```
gadomski, line 417: Same comment as in `test_item.py`, these branch checks could be moved to `test_utils.py` and then this could just check the basic regression case.

gadomski (changes requested):

tests/test_item.py
duckontheweb: Yeah, good point. Putting them here actually helped me catch the fact that we weren't using `str_to_datetime` everywhere that we were parsing timestamps, but now that that is fixed we probably don't need these here as a regression test.

duckontheweb: Fixed in 6997000e40f248ea8779f53d5b41c3f535aa9a51

tests/test_collection.py
duckontheweb: Fixed in 6997000e40f248ea8779f53d5b41c3f535aa9a51

tests/test_item.py
duckontheweb: Removed in 6997000e40f248ea8779f53d5b41c3f535aa9a51

duckontheweb:
@TomAugspurger I verified that this fixes the original issue in `pystac.read_file` as well:

```console
$ TZ=UTC python -c 'import pystac; print(pystac.read_file("https://planetarycomputer.microsoft.com/api/stac/v1/collections/sentinel-2-l2a").extent.temporal.intervals[0][0].tzinfo)'
tzutc()
```

TomAugspurger (approved):
Thanks!

gadomski (approved):

PR #878, opened 2022-08-18
Attribute `code` has type `str` but is used as type `None`

luca-digrazia:
**Related Issue(s):** #
Type error found with Pyre

**Description:**
"filename": "tests/extensions/test_grid.py"
"warning_type": "Incompatible attribute type [8]"
"warning_message": " Attribute `code` declared in class `GridExtension` has type `str` but is used as type `None`."
"warning_line": 89
"fix": None to ""

### DIFF
--- a/tests/extensions/test_grid.py
+++ b/tests/extensions/test_grid.py
@@ -92,6 +92,18 @@ class GridTest(unittest.TestCase):
             # that does not conform to the type signature.
             # https://github.com/stac-utils/pystac/pull/878#discussion_r957352232
             GridExtension.ext(self.item).code = None  # type: ignore
+        with self.assertRaises(ValueError):
+            # First segment has to be all caps
+            # https://github.com/stac-utils/pystac/pull/878#discussion_r957354927
+            GridExtension.ext(self.item).code = "this-is-not-a-grid-code"
+        with self.assertRaises(ValueError):
+            # Folks might try to put an epsg code in
+            # https://github.com/stac-utils/pystac/pull/878#discussion_r957355415
+            GridExtension.ext(self.item).code = "4326"
+        with self.assertRaises(ValueError):
+            # Folks might try to put an epsg code in
+            # https://github.com/stac-utils/pystac/pull/878#discussion_r957355415
+            GridExtension.ext(self.item).code = "EPSG:4326"
 
     def test_extension_not_implemented(self) -> None:
         # Should raise exception if Item does not include extension URI
--- a/tests/extensions/test_grid.py
+++ b/tests/extensions/test_grid.py
@@ -1,4 +1,6 @@
 """Tests for pystac.extensions.grid."""
+# This is for the type checking on GridTest.test_clear_code
+# mypy: warn_unused_ignores=False
 
 import datetime
 from typing import Any, Dict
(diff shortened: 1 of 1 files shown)


gadomski (commented):
This seems like a false positive -- the `code` property returns `Optional[str]`: https://github.com/stac-utils/pystac/blob/69fcf3f2137045bd45ac5cf6cbd2fdad869eb169/pystac/extensions/grid.py#L68-L71

luca-digrazia:
> This seems like a false positive -- the `code` property returns `Optional[str]`:
> 
> https://github.com/stac-utils/pystac/blob/69fcf3f2137045bd45ac5cf6cbd2fdad869eb169/pystac/extensions/grid.py#L68-L71

Thanks, for the answer, but I think it is not the method call in line 89.

gadomski:
Thanks, I misunderstood what was going on. It looks like the line you're trying to fix is intentionally triggering a `ValueError`, so I think the correct fix might be to change the signature of the setter and the underlying helper function, e.g.:

```diff
diff --git a/pystac/extensions/grid.py b/pystac/extensions/grid.py
index 928897e..3b0bd6f 100644
--- a/pystac/extensions/grid.py
+++ b/pystac/extensions/grid.py
@@ -17,7 +17,7 @@ CODE_REGEX: str = r"[A-Z]+-[-_.A-Za-z0-9]+"
 CODE_PATTERN: Pattern[str] = re.compile(CODE_REGEX)
 
 
-def validated_code(v: str) -> str:
+def validated_code(v: Optional[str]) -> str:
     if not isinstance(v, str):
         raise ValueError("Invalid Grid code: must be str")
     if not CODE_PATTERN.fullmatch(v):
@@ -71,7 +71,7 @@ class GridExtension(
         return self._get_property(CODE_PROP, str)
 
     @code.setter
-    def code(self, v: str) -> None:
+    def code(self, v: Optional[str]) -> None:
         self._set_property(CODE_PROP, validated_code(v), pop_if_none=False)
 
     @classmethod
```

Requesting Phil's review since he added this in #799.

luca-digrazia:
Up :)

tests/extensions/test_grid.py
```diff
@@ -86,7 +86,7 @@ def test_clear_code(self) -> None:
         GridExtension.ext(self.item).apply(code)
 
         with self.assertRaises(ValueError):
-            GridExtension.ext(self.item).code = None
+            GridExtension.ext(self.item).code = ""
```
philvarner, line 89: either attempting to set this to None or empty string should raise a ValueError, so the `None` case should remain instead of being replaced. Also some tests for non-compliant code strings should be added. There is a regex defined in GridExtension as to what they need to match.

philvarner (changes requested):
see comment

tests/extensions/test_grid.py
gadomski: Should we update the setter signature to take an `Optional[str]` so we don't make type checkers angry? No change to functionality (still raises `ValueError`).

> Also some tests for non-compliant code strings should be added.

Do you have a couple off the top of your head? Just stuff like `this-is-not-a-grid-code`?

philvarner: or accidentally put the epsg number in there.

gadomski (approved):

Issue #886, opened 2022-09-19
RFC: Add a mechanism for passing authorization headers to child links


john-dupuy:
I recently discovered this project and it's awesome! I am dealing with a STAC catalog that requires authentication and I noticed that there is no mechanism for me to pass an auth header for when I want to walk a STAC catalog. 

As a simple scenario consider the following:

```python
from pystac import Catalog
catalog =  Catalog.from_dict(requests.get(<stac-url-requiring-auth>, headers={"Authorization": "api-key <my-api-key>"}))
catalog.validate_all()
```
In this situation, we will get an exception at the first link with an exception:
```
urllib.error.HTTPError: HTTP Error 401: Unauthorized
```
coming from this line:
```
pystac/stac_io.py", line 292, in read_text_from_href
    with urlopen(href) as f:
```

It would be great if we could add an auth header when we create the STAC object. As a workaround I created a class like 
```python
class StacIOWithAuth(DefaultStacIO):
    """Override 'read_text_from_href' so that we can add authn (api-key) to the requests that pystac makes."""

    def __init__(self, api_key):
        self.api_key = api_key

    def read_text_from_href(self, href: str) -> str:
        """Reads file as a UTF-8 string.

        If ``href`` has a "scheme" (e.g. if it starts with "https://") then this will
        use :func:`urllib.request.urlopen` to open the file and read the contents;
        otherwise, :func:`open` will be used to open a local file.

        Args:

            href : The URI of the file to open.
        """
        parsed = safe_urlparse(href)
        href_contents: str
        if parsed.scheme != "":
            try:
                req = Request(href, headers={"Authorization": f"api-key {self.api_key}"})
                with urlopen(req) as f:
                    href_contents = f.read().decode("utf-8")
            except HTTPError as e:
                raise Exception("Could not read uri {}".format(href)) from e
        else:
            with open(href, encoding="utf-8") as f:
                href_contents = f.read()
        return href_contents
```

And then just set `catalog._stac_io = StacIOWithAuth(<api_key>)`.

Thanks for reading - I'm happy to take this on as a first issue if no one has bandwidth to tackle it.

gadomski:
Your approach makes sense and is the currently-supported way of doing custom IO. Note that there is a [StacIO.set_default](https://pystac.readthedocs.io/en/stable/api/stac_io.html#pystac.stac_io.StacIO.set_default) method, which you can use to enable your custom StacIO without touching the "private" `_stac_io` attribute.

[...]

PR #889, opened 2022-09-23
Add a mechanism for adding headers to StacIO requests

john-dupuy:
**Related Issue(s):** #886

**Description:**
This PR aims to address #886 by adding a mechanism for adding headers for StacIO requests.

The suggested approach for using this:
```python
from pystac import StacIO
stac_io = StacIO.default()
stac_io.headers = {"Authorization": "<some-authn-header>"}

catalog = Catalog.from_file("<some-href-that-requires-authn>", stac_io=stac_io)
```
You could also just directly instantiate the `StacIO` or `DefaultStacIO` class, with headers included as an arg. 

**PR Checklist:**

### DIFF
--- a/tests/test_stac_io.py
+++ b/tests/test_stac_io.py
@@ -1,3 +1,4 @@
+import json
 import os
 import unittest
 import tempfile
@@ -103,14 +104,13 @@ class StacIOTest(unittest.TestCase):
     def test_headers_stac_io(self, urlopen_mock: unittest.mock.MagicMock) -> None:
         stac_io = DefaultStacIO(headers={"Authorization": "api-key fake-api-key-value"})
 
-        try:
-            # note we don't care if this raises an exception, we just want to make
-            # sure urlopen was called with the appropriate headers
-            pystac.Catalog.from_file(
-                "https://example.com/catalog.json", stac_io=stac_io
-            )
-        except Exception:
-            pass
+        catalog = pystac.Catalog("an-id", "a description").to_dict()
+        # required until https://github.com/stac-utils/pystac/pull/896 is merged
+        catalog["links"] = []
+        urlopen_mock.return_value.__enter__.return_value.read.return_value = json.dumps(
+            catalog
+        ).encode("utf-8")
+        pystac.Catalog.from_file("https://example.com/catalog.json", stac_io=stac_io)
 
         request_obj = urlopen_mock.call_args[0][0]
         self.assertEqual(request_obj.headers, stac_io.headers)
--- a/tests/test_stac_io.py
+++ b/tests/test_stac_io.py
@@ -98,3 +98,19 @@ class StacIOTest(unittest.TestCase):
                 str(excinfo.exception),
                 f'Found duplicate object name "key" in {src_href}',
             )
+
+    @unittest.mock.patch("pystac.stac_io.urlopen")
+    def test_headers_stac_io(self, urlopen_mock: unittest.mock.MagicMock) -> None:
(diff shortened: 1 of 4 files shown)


john-dupuy:
@gadomski can you take a look?

gadomski:
> @gadomski can you take a look?

Yes, it's in my queue, thanks for the bump. Right now I'm maintaining PySTAC in my off time so responses may be a bit slower, apologies.

tests/test_stac_io.py
```diff
@@ -98,3 +98,19 @@ class ReportingStacIO(DefaultStacIO, DuplicateKeyReportingMixin):
...
+            # note we don't care if this raises an exception, we just want to make
+            # sure urlopen was called with the appropriate headers
+            pystac.Catalog.from_file(
+                "https://example.com/catalog.json", stac_io=stac_io
+            )
+        except Exception:
+            pass
+
+        request_obj = urlopen_mock.call_args[0][0]
+        self.assertEqual(request_obj.headers, stac_io.headers)
```
gadomski, line 116: I played around with the mocks, and I think we can get rid of the blanket try/except by setting the `read` return value. Can you update your PR to look something like this:

```diff
diff --git a/tests/test_stac_io.py b/tests/test_stac_io.py
index 8609f4b..5314a96 100644
--- a/tests/test_stac_io.py
+++ b/tests/test_stac_io.py
@@ -1,6 +1,8 @@
+import json
 import os
-import unittest
 import tempfile
+import unittest
+import unittest.mock
 
 import pystac
 from pystac.stac_io import StacIO, DefaultStacIO, DuplicateKeyReportingMixin
@@ -103,14 +105,13 @@ class StacIOTest(unittest.TestCase):
     def test_headers_stac_io(self, urlopen_mock: unittest.mock.MagicMock) -> None:
         stac_io = DefaultStacIO(headers={"Authorization": "api-key fake-api-key-value"})
 
-        try:
-            # note we don't care if this raises an exception, we just want to make
-            # sure urlopen was called with the appropriate headers
-            pystac.Catalog.from_file(
-                "https://example.com/catalog.json", stac_io=stac_io
-            )
-        except Exception:
-            pass
+        catalog = pystac.Catalog("an-id", "a description").to_dict()
+        # required until https://github.com/stac-utils/pystac/pull/896 is merged
+        catalog["links"] = []
+        urlopen_mock.return_value.__enter__.return_value.read.return_value = json.dumps(
+            catalog
+        ).encode("utf-8")
+        pystac.Catalog.from_file("https://example.com/catalog.json", stac_io=stac_io)
 
         request_obj = urlopen_mock.call_args[0][0]
         self.assertEqual(request_obj.headers, stac_io.headers)
```

gadomski (changes requested):

tests/test_stac_io.py
john-dupuy: Thanks for the suggestion, updated!

gadomski (approved):
I've tested this with a real-world STAC server that requires authorization, and it appears to work. Thanks for the contribution! I'm going to request @matthewhanson's review as well, because he brought up the point in a side channel that you _could_ just use [pystac-client](https://pystac-client.readthedocs.io/en/stable/), which already supports authorization. In my opinion, this PR is still useful, as it doesn't add any extra dependencies to add functionality, but would like Matt to weigh in as well.

matthewhanson:
Thanks for the PR @john-dupuy .  As @gadomski said, you could use pystac-client here which handles header auth and works fine on static catalogs. However I agree that since it requires no dependencies and is a simple addition it doesn't hurt to include it in PySTAC as well.

matthewhanson (approved):

Issue #183, opened 2020-09-08
Creating a collection without links has an unexpected side-effect


jisantuc:
Creating a collection without any links fills in a `root` link with a `null` href, but `null` is not a valid value for `href`:

```python
>>> collection = Collection(
    collection_id,
    "Example collection",
    Extent(SpatialExtent([[0, 0, 1, 1]]), TemporalExtent([[datetime.now(), None]])),
)

>>> collection.links[0].get_href()
>>>
```

Neither [Use of Links](https://github.com/radiantearth/stac-spec/blob/master/best-practices.md#use-of-links) nor the [collection spec](https://github.com/radiantearth/stac-spec/blob/master/collection-spec/collection-spec.md) says that I _have_ to have links. One place where I definitely don't want a `root` link is when I'm just trying to create a `Collection` to POST to Franklin. Having a surprise link that Franklin correctly said was no good was a bit of a surprise.

PR #896, opened 2022-10-04
Create valid dict without root href

gadomski:
**Description:**

Catalogs and Collections always have a root href. If the root Catalog or Collection (e.g. itself) does not have a self href, the dictionary serialization included a root link with href set to `null`, which is invalid. This fix removes the root link if it doesn't have a valid href.

Not required for items because they don't have a root link by default.

I discovered this problem when trying to write-then-read a very simple Collection using `save_object`, but the problem proved to be in the dict serialization.

**PR Checklist:**

- ~[ ] Documentation has been updated to reflect changes, if applicable~


gadomski:
Whoops, my markdown linter added some spaces to the CHANGELOG. If those are unwanted noise I can remove, LMK.

pjhartzell (approved):
I think the changes address #183 from a practical standpoint. The creation of null href root links seems embedded in pystac (editing `catalog.py` such that a root link that will have a null href is not created causes multiple test failures).

gadomski:
> editing catalog.py such that a root link that will have a null href is not created causes multiple test failures

Yup, I found the same thing -- PySTAC kind of expects a `root`, whether it has a href or not.

PR #900, opened 2022-10-13
Normalize hrefs while skipping unresolved links

### LINKED ISSUE #284: Normalize URLs and set link types for newly added entities

gadomski:
**Related Issue(s):**
- Closes #284

**Description:**

Adds a `skip_unresolved` argument to `Catalog.normalize_hrefs` and `Catalog.normalize_and_save`. Instead of using `Catalog.get_items` and `Catalog.get_children` (which resolve their links), we re-implement with additional logic to conditionally skip unresolved links. We considered pushing `skip_unresolved` all the way down into `Link.get_stac_object`, but this seemed like overkill, since the motivating use-case for `skip_unresolved` is to avoid walking the whole tree when normalizing (#284).

Because `save` already skips unresolved links, this enables `normalize_and_save` to only touch objects that have been added to the tree or modified (and therefore resolved).

Note this isn't exactly the solution proposed by @matthewhanson in #284, since the hrefs newly-added entities aren't normalized. These changes are only confined to `normalize`.

This is a feature-adding, backwards compatible change that could be incorporated in a MINOR release.

cc @l0b0 

**PR Checklist:**

### DIFF
--- a/pystac/catalog.py
+++ b/pystac/catalog.py
@@ -572,6 +572,7 @@ class Catalog(STACObject):
         catalog_type: Optional[CatalogType] = None,
         strategy: Optional[HrefLayoutStrategy] = None,
         stac_io: Optional[pystac.StacIO] = None,
+        skip_unresolved: bool = False,
     ) -> None:
         """Normalizes link HREFs to the given root_href, and saves the catalog.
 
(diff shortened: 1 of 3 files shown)


matthewhanson:
This looks fine to me, I think it's a good change - in most cases of adding new items to a catalog you shouldn't need to resolve anything that isn't newly added (i.e., resolved).

But I'm not sure I follow this:

```
Note this isn't exactly the solution proposed by @matthewhanson in https://github.com/stac-utils/pystac/issues/284, since the hrefs newly-added entities aren't normalized. These changes are only confined to normalize.
```

What do you mean by newly-added entities aren't normalized?

gadomski:
> What do you mean by newly-added entities aren't normalized?

If I understood #284 correctly, you were proposing a change where

```python
collection.add_item(item)
```

would auto-magically normalize the item href (without having to call `collection.normalize_hrefs()`). This PR does _not_ do that.

pystac/catalog.py
```diff
@@ -639,16 +656,27 @@ def process_catalog(
...
-
-            for child in cat.get_children():
-                setter_funcs.extend(process_catalog(child, new_root, is_root=False))
+            for link in cat.get_links():
+                if skip_unresolved and not link.is_resolved():
+                    continue
+                elif link.rel == pystac.RelType.ITEM:
+                    link.resolve_stac_object(root=self.get_root())
+                    if isinstance(link.target, pystac.Item):
+                        setter_funcs.append(process_item(link.target, new_root))
```
pjhartzell, line 671: What are the implications if the link target is not a pystac.Item? Was the rel type incorrect? Right now it looks like we silently move on.

pystac/catalog.py
```diff
@@ -639,16 +656,27 @@ def process_catalog(
...
+                elif link.rel == pystac.RelType.CHILD:
+                    link.resolve_stac_object(root=self.get_root())
+                    if isinstance(link.target, pystac.Catalog) or isinstance(
+                        link.target, pystac.Collection
+                    ):
```
pjhartzell, line 676: Same question as above.

pjhartzell (changes requested):

pystac/catalog.py
gadomski: Correct. This is a slight change in behavior -- previously, we were processing the target as if it were an Item, without any check: https://github.com/stac-utils/pystac/blob/254040480ed1fcfd14599b197d12649b15de649d/pystac/catalog.py#L432-L440

In the interest of changing as little as possible, maybe I should keep the old behavior, aka just pass the target to `process_item` without checking whether it's an item? What do you think?

pjhartzell: Probably a better chance that something will fail if it is not an Item. I _think_ that is preferable than silently skipping it.

pjhartzell (approved):

PR #918, opened 2022-11-10
Add start and end datetimes to `Item.__init__`

### LINKED ISSUE #710: request for consistent datetime intake for Item
The `datetime` argument to `pystac.Item` expects `datetime` objects, while the `properties` argument expects a dictionary with string type for `start_datetime` and `end_datetime` when these are required. That inconsistency doesn't seem necessary, and likely leads to extra work handling conversions and time-zone ambiguity.

Why not let the `datetime` argument handle both cases, and not ask users to mess with the properties?

To elaborate, the code below works, but probably shouldn't. But rathern than setting `datetime=None`, I'd like to set `datetime=tspan` and skip setting properties.

```
import pystac
import pandas as pd

tspan = pd.to_datetime(('2022-01-01T00:00:00 -05:00', '2022-01-31T00:00:00 -05:00'))

c = pystac.Collection(id='Test', description=None, extent=None)
i = pystac.Item(
    id='Test-Item',
    geometry={'type': 'Polygon', 'coordinates': (
        ((1, 0),(1, 1),(0, 1),(0, 0),(1, 0)),
    )},
    bbox=(0, 0, 1, 1),
    datetime=tspan[0],
    properties={'start_datetime': tspan[0].isoformat(), 'end_datetime': tspan[1].isoformat()},
)
c.add_item(i)
c.update_extent_from_items()
```

Would you be interested in a PR that makes the `datetime` argument handle one **or** a two-tuple datetime and sets properties as appropriate? Fair warning ... I'd have to learn `typing`.

### LINKED ISSUE #905: CommonMetadata: If datetime is None, a start_datetime and end_datetime must be supplied in the properties.
The following code leads to an error, which is cumbersome because the intended way of adding common metadata doesn't work:

```py
        item = Item(
            stac_extensions=[],
            id=id,
            properties=properties,
            geometry=constants.GEOMETRY,
            bbox=constants.BBOX,
            datetime=None,
            collection=collection,
        )

        common_item = CommonMetadata(item)
        common_item.title = f"Land Cover Map of {year}"
        common_item.gsd = constants.GSD
        common_item.start_datetime = start_datetime
        common_item.end_datetime = end_datetime
```

I get: 
> Invalid Item: If datetime is None, a start_datetime and end_datetime must be supplied in the properties.

It neglects the purpose of the Common Metadata class a bit...

gadomski:
**Related Issue(s):** 

- Closes #905 
- Closes #710 (though with a different solution than the one proposed there)

**Description:**

It's pretty awkward to provide `start_datetime` and `end_datetime` as strings in `properties`, when you can provide a datetime object as `datetime`. This PR adds two optional arguments to set `start_datetime` and `end_datetime` in the constructor. These values will override anything set in `properties`.

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - `AsIsLayoutStrategy` ([#919](https://github.com/stac-utils/pystac/pull/919))
 - `__geo_interface__` for items ([#885](https://github.com/stac-utils/pystac/pull/885))
 - Optional `strategy` parameter to `catalog.add_items()` ([#967](https://github.com/stac-utils/pystac/pull/967))
+- `start_datetime` and `end_datetime` arguments to the `Item` constructor ([#918](https://github.com/stac-utils/pystac/pull/918))
 
 ### Removed
 
--- a/pystac/item.py
+++ b/pystac/item.py
@@ -40,8 +40,12 @@ class Item(STACObject):
             where n is the number of dimensions. Could also be None in the case of a
             null geometry.
         datetime : Datetime associated with this item. If None,
-            a start_datetime and end_datetime must be supplied in the properties.
+            a start_datetime and end_datetime must be supplied.
         properties : A dictionary of additional metadata for the item.
+        start_datetime : Optional start datetime, part of common metadata. This value
+            will override any `start_datetime` key in properties.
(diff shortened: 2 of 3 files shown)


pjhartzell (approved):
Looks good.

Issue #873, opened 2022-08-15
Add `role`, `media_type` filtering arguments to `Item.get_assets`


TomAugspurger:
`Item.get_links` includes some arguments to filter which links are returned by `rel` and `media_type`:

https://github.com/stac-utils/pystac/blob/876b60a8547d9307b564cfb372c1237b324c555b/pystac/stac_object.py#L115-L119

`Item.get_assets` would be more useful if it could filter on `roles` and `media_type`. We could replace

```python
import pystac

item = pystac.read_file("https://planetarycomputer.microsoft.com/api/stac/v1/collections/jrc-gsw/items/90E_30Nv1_3_2020")

cog_assets = {
    asset_key: asset
    for asset_key, asset in item.assets.items()
    if asset.media_type == pystac.MediaType.COG
}
```

with `cog_assets = item.get_assets(media_type=pystac.MediaType.COG)`

Since `roles` is a list, we might need a flag to control whether you need any / all of the roles specified to be returned. But as a start, I think allowing a single `role` and returning any assets with that would be best.

PR #936, opened 2023-01-04
Add filter to  Item and Collection `get_assets` method

pjhartzell:
**Related Issue(s):** 
#873 

**Description:**
Adds ability to filter Item and Collection Assets by media_type and role via `get_assets()` method.

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
+- Add `media_type` and `role` filtering to Item and Collection `get_assets()` method ([#936](https://github.com/stac-utils/pystac/pull/936))
 - Update Grid Extension support to v1.1.0 and fix issue with grid:code prefix validation ([#925](https://github.com/stac-utils/pystac/pull/925))
 - Adds custom `header` support to `DefaultStacIO` ([#889](https://github.com/stac-utils/pystac/pull/889))
 - Python 3.11 checks in CI ([#908](https://github.com/stac-utils/pystac/pull/908))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 - Add `media_type` and `role` filtering to Item and Collection `get_assets()` method ([#936](https://github.com/stac-utils/pystac/pull/936))
+- `Asset.has_role` ([#936](https://github.com/stac-utils/pystac/pull/936))
 - Update Grid Extension support to v1.1.0 and fix issue with grid:code prefix validation ([#925](https://github.com/stac-utils/pystac/pull/925))
 - Adds custom `header` support to `DefaultStacIO` ([#889](https://github.com/stac-utils/pystac/pull/889))
 - Python 3.11 checks in CI ([#908](https://github.com/stac-utils/pystac/pull/908))
--- a/pystac/collection.py
+++ b/pystac/collection.py
@@ -692,19 +692,14 @@ class Collection(Catalog):
             Dict[str, Asset]: A dictionary of assets that match ``media_type``
                 and/or ``role`` if set or else all of this collection's assets.
         """
-        assets = dict(self.assets.items())
-        if media_type is not None:
-            assets = {
-                key: asset
-                for key, asset in assets.items()
-                if asset.media_type == media_type
-            }
-        if role is not None:
-            assets = {
-                key: asset
-                for key, asset in assets.items()
-                if asset.roles is not None and role in asset.roles
-            }
+        if media_type is None and role is None:
+            return dict(self.assets.items())
+        assets = dict()
+        for key, asset in self.assets.items():
+            if (media_type is None or asset.media_type == media_type) and (
+                role is None or (asset.roles is not None and role in asset.roles)
+            ):
+                assets[key] = asset
         return assets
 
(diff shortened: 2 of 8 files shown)


pystac/collection.py
```diff
@@ -675,13 +675,37 @@ def from_dict(
...
+                for key, asset in assets.items()
+                if asset.media_type == media_type
+            }
+        if role is not None:
+            assets = {
+                key: asset
+                for key, asset in assets.items()
+                if asset.roles is not None and role in asset.roles
+            }
+        return assets
```
gadomski, line 708: This is a little nitpicky, but we're iterating over the assets as many as three times; once to create the new dict, then twice if both `media_type` and `role` are set. Iterating once seems preferable (with a short circuit):

```python
if media_type is None and role is None:
    return dict(self.assets.items())
assets = dict()
for key, asset in self.assets.items():
    if (media_type is None or asset.media_type == media_type) and (role is None or role in asset.roles):
        assets[key] = asset
return assets
```

pystac/item.py
```diff
@@ -221,13 +221,37 @@ def set_datetime(self, datetime: Datetime, asset: Optional[Asset] = None) -> Non
...
+                for key, asset in assets.items()
+                if asset.media_type == media_type
+            }
+        if role is not None:
+            assets = {
+                key: asset
+                for key, asset in assets.items()
+                if asset.roles is not None and role in asset.roles
+            }
+        return assets
```
gadomski, line 254: Same as the comment on the collection method.

gadomski (changes requested):

pystac/collection.py
pjhartzell: Nope, not nitpicky. That's cleaner. We do need to augment the condition to guard against a `None` value in `asset.roles`:

```python
if (media_type is None or asset.media_type == media_type) and (role is None or (asset.roles is not None and role in asset.roles)):
```

That's getting to be a mouthful. We could simplify `asset.roles is not None and ...` to just `asset.roles and ...`, but that's not as clear (to me).

pystac/collection.py
```diff
@@ -675,13 +675,32 @@ def from_dict(
...
+            Dict[str, Asset]: A dictionary of assets that match ``media_type``
+                and/or ``role`` if set or else all of this collection's assets.
         """
-        return dict(self.assets.items())
+        if media_type is None and role is None:
+            return dict(self.assets.items())
+        assets = dict()
+        for key, asset in self.assets.items():
+            if (media_type is None or asset.media_type == media_type) and (
+                role is None or (asset.roles is not None and role in asset.roles)
```
gadomski, line 700: I agree that this is getting a little wild. What do you think about a new `Asset.has_role` method? E.g.

```suggestion
            if (media_type is None or asset.media_type == media_type) and (role is None or asset.has_role(role)):
```

and then at https://github.com/stac-utils/pystac/blob/017ddfd8592e822ecc6905b5ea3ed0cd121798dc/pystac/asset.py:

```python
def has_role(self, role: str) -> bool:
    if self.roles is None:
        return False
    else:
        return role in self.roles

Comment (pjhartzell) [reply pystac/collection.py]: Yeah, that works for me. Done.

File: CHANGELOG.md [hunk 4]
```diff
@@ -4,6 +4,7 @@
 
 ### Added
 
+- Add `media_type` and `role` filtering to Item and Collection `get_assets()` method ([#936](https://github.com/stac-utils/pystac/pull/936))
```
gadomski, line 7:
```suggestion
- Add `media_type` and `role` filtering to Item and Collection `get_assets()` method ([#936](https://github.com/stac-utils/pystac/pull/936))
- `Asset.has_role` ([#936](https://github.com/stac-utils/pystac/pull/936))
```

gadomski (changes requested):

gadomski (approved):

pystac/collection.py
```diff
@@ -675,13 +675,32 @@ def from_dict(
 
         return collection
 
-    def get_assets(self) -> Dict[str, Asset]:
-        """Get this item's assets.
+    def get_assets(
+        self,
+        media_type: Optional[Union[str, pystac.MediaType]] = None,
+        role: Optional[Union[str, pystac.MediaType]] = None,
```
gadomski, line 681:
```suggestion
        role: Optional[str] = None,
```

pystac/item.py
```diff
@@ -221,13 +221,32 @@ def set_datetime(self, datetime: Datetime, asset: Optional[Asset] = None) -> Non
         else:
             asset.extra_fields["datetime"] = datetime_to_str(datetime)
 
-    def get_assets(self) -> Dict[str, Asset]:
+    def get_assets(
+        self,
+        media_type: Optional[Union[str, pystac.MediaType]] = None,
+        role: Optional[Union[str, pystac.MediaType]] = None,
```
gadomski, line 227:
```suggestion
        role: Optional[str] = None,
```

gadomski (changes requested):

pystac/collection.py
pjhartzell: Yeesh. Good catch. Fixed.

pystac/item.py
pjhartzell: Fixed.

gadomski (approved):

PR #939, opened 2023-01-05
Switch to pytest

### LINKED ISSUE #931: Convert to pytest
As mentioned in https://github.com/stac-utils/stactools/issues/218#issuecomment-1022450823, PySTAC is currently _not_ using **pytest**, but it probably should.

This could be as simple as changing over the requirements-test.txt and the testing scripts, but it could also include more in-depth updates to use pytest features. In particular, [pytest-vcr](https://pytest-vcr.readthedocs.io/en/latest/) could speed up/robustify network tests (we get occasional CI errors from 500s from CMR, e.g.: https://github.com/stac-utils/pystac/actions/runs/3434301808/jobs/5725483518).

pjhartzell:
**Related Issue(s):**
- Closes #931

**Description:** 
- Switches from unittest to pytest.
- Adds pytest parametrizations in place of for loops and subTests.
- The tests directory was removed from the coverage measurement to better represent the pystac code coverage. This results in a lower coverage percentage, so the benchmark (`fail_under`) was lowered from 94% to 90%.
- Adds tests for the `html` module.

**PR Checklist:**


pjhartzell:
This the minimum set of changes to use pytest. Per #931 we may want to look at using more pytest features.

scripts/test
```diff
@@ -10,10 +10,10 @@ echo
 if [[ -z ${CI} || -n ${CHECK_COVERAGE} ]]; then
     echo " -- RUNNING UNIT TESTS (WITH COVERAGE) --"
     # Test suite with coverage enabled
-    coverage run -m unittest discover tests
+    coverage run -m pytest tests
```
gadomski, line 13: We should use [pytest-cov](https://pypi.org/project/pytest-cov/) to run coverage.

gadomski (changes requested):
Looks good. A couple of notes / changes:

1. While I think we should avoid any huge test refactors in this PR, there are a couple of places where **pytest** gets us some quick wins. E.g. https://github.com/stac-utils/pystac/blob/d28bd4387fc6093b03d600a32ab7647e151610a4/tests/test_writing.py#L114-L126 could be made a lot simpler (and report a lot better) using [pytest's parameterized tests](https://docs.pytest.org/en/7.1.x/example/parametrize.html). Could you update that test, and any other similar ones, to use parameterization?
2. I _don't_ think it's worth going through and updating every test to use [pytest's fixtures](https://docs.pytest.org/en/6.2.x/fixture.html) right now, but could you open an issue to capture that future task?
3. Same as above but for **pytest-vcr**, we could robustify some of our tests by using that instead of hitting the network every time, but that can be its own task that we do after this PR is merged. Could you open an issue for that as well?

Thanks!

pjhartzell:
A comment here about including `tests` (and not just `pystac`) in coverage. My initial reaction is that I want (and expect) to see coverage for the package of interest, i.e., `pystac`, only. 

However, I see that `tests` was explicitly added by @l0b0 in https://github.com/stac-utils/pystac/commit/3e5fed638e69fb51cf5322571d8ccef7b6b90c9b. Any thoughts from the broader community on this?

l0b0:
> However, I see that `tests` was explicitly added by @l0b0 in [3e5fed6](https://github.com/stac-utils/pystac/commit/3e5fed638e69fb51cf5322571d8ccef7b6b90c9b). Any thoughts from the STAC community on this?

I believe that was in response to a test (or test utility, I don't remember) which was not being run. Test code should always have 100% coverage, no exceptions, so it shouldn't really clutter up any reports.

So, basically, while "meta-coverage" it answers a different question than production code coverage, it's still a useful thing to know.

gadomski:
> So, basically, while "meta-coverage" it answers a different question than production code coverage, it's still a useful thing to know.

To me, it's confusing to include test code coverage in the "core" coverage report. If we do want to check for non-run tests, IMO that should be part of a separate coverage run. I also am personally 👍🏽  for dropping it as a part of CI -- if we're worried about non-run tests, its not too hard to do a one-off coverage run of just the test suite.

gadomski (approved):
Since we're removing `--cov=tests`, I opened https://github.com/stac-utils/pystac/issues/951 to capture some non-run test lines I found from a manual run.

pjhartzell:
> Since we're removing `--cov=tests`

I haven't removed that yet, but, yeah, 👍  on the new issue. I'll remove the tests directory coverage from the `.coveragerc` file.

pjhartzell:
@gadomski 
After removing the tests directory from coverage, the coverage percentage dropped to 90.42% --> I set the `fail_under` value in `.coveragerc` to 90.4.

gadomski:
> After removing the tests directory from coverage, the coverage percentage dropped to 90.42% --> I set the fail_under value in .coveragerc to 90.4.

Yeah, I think that's a good thing -- the tests directory was giving us an over-estimate of project coverage.

There's two options here
1. Merge with "non-passing" coverage, and open an issue to raise coverage above 90%
2. Up the coverage before merge

IMO if there's an obvious place we can add a test or two to this PR to get coverage over 90%, that'll make CI happy w/o dropping our threshold too low.

gadomski (approved):

pjhartzell:
- Coverage was at 94.13%. The fail under was set to 94.0%.
- After removing the tests directory, coverage is now at 90.42%. I set a new fail under to 90.4%. We can make that a cleaner number (90.0%) if desired.

>IMO if there's an obvious place we can add a test or two to this PR to get coverage over 90%, that'll make CI happy w/o dropping our threshold too low.

We are still above 90%. I think there was some number confusion. Hopefully, the above bullets clear that up.

gadomski:
Coverage looks like its failing: https://github.com/stac-utils/pystac/actions/runs/3905719122/jobs/6673063889#step:10:64.

pjhartzell:
Oops. Yeah, just saw that. Will look into this.

tests/html/test_html.py
```diff
@@ -0,0 +1,28 @@
...
+from pystac.html.jinja_env import get_jinja_env
+from tests.utils import TestCases
+
+
+def test_item_html() -> None:
+    item_stac_href = TestCases.get_path("data-files/item/sample-item.json")
+    item_html_href = TestCases.get_path("data-files/html/sample-item.html")
+    item = pystac.Item.from_file(item_stac_href)
+    item.set_self_href(None)
+    assert item._repr_html_() == open(item_html_href).read()
```
gadomski, line 15: I'm a little uneasy about failing tests if the HTML representation changes just a little bit. What do you think about just checking that the `_repr_html_()` produces valid HTML, and nothing more?

gadomski (changes requested):

.github/workflows/continuous-integration.yml
```diff
@@ -46,6 +46,7 @@ jobs:
         run: |
           pip install --upgrade pip
           pip install -r requirements-test.txt
+          pip install jinja2
```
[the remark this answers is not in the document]
pjhartzell: It's currently in `requirements-docs.txt`...I don't think we want it in two requirements files.

tests/html/test_html.py
pjhartzell: Yeah, that makes sense. I don't like monolothic comparisons either.

.github/workflows/continuous-integration.yml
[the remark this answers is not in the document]
gadomski: Why not? If we need it for tests and we need it for docs ...

We have `check-dev-dependencies` to make sure there isn't a version conflict: https://github.com/stac-utils/pystac/blob/7cb4b69851d36a0f6456b2ee2addc697c4a92c26/.github/workflows/continuous-integration.yml#L148-L161

tests/html/test_html.py
pjhartzell: I added a simple html parser in place of the html string comparison. Uncovered what appears to be a bug in the `_html_repr()` - see #955.

.github/workflows/continuous-integration.yml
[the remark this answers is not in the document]
pjhartzell: Just seemed bad form. But if we have controls in place, then all good. I added the necessary dependencies to `requirement-dev.txt` file.

gadomski (approved):

PR #966, opened 2023-01-26
Filter links with only media_type

### LINKED ISSUE #937: Media type argument in `Item.get_links` method is ignored
The `media_type` argument is ignored when filtering the returned links with the `Item.get_links` method. It looks like it is short-circuited by the absence of a `rel` argument.

https://github.com/stac-utils/pystac/blob/e77cd0aa82b34803539ddaff310beadba9ca01e2/pystac/stac_object.py#L125-L133

I would expect to be able to filter on either (or both) `rel` and `media_type`. For example, I would expect the following to return the single link with an HTML media type, but it returns all links.

```
import pystac

item = pystac.read_file("https://planetarycomputer.microsoft.com/api/stac/v1/collections/landsat-c2-l2/items/LC09_L2SR_083122_20221230_02_T2")

html_links = item.get_links(media_type=pystac.MediaType.HTML)
for link in html_links:
    print(link.to_dict())

# Output:
# {'rel': 'collection', 'href': 'https://planetarycomputer.microsoft.com/api/stac/v1/collections/landsat-c2-l2', 'type': 'application/json'}
# {'rel': 'parent', 'href': 'https://planetarycomputer.microsoft.com/api/stac/v1/collections/landsat-c2-l2', 'type': 'application/json'}
# {'rel': <RelType.ROOT: 'root'>, 'href': 'https://planetarycomputer.microsoft.com/api/stac/v1/', 'type': <MediaType.JSON: 'application/json'>, 'title': 'Microsoft Planetary Computer STAC API'}
# {'rel': 'self', 'href': 'https://planetarycomputer.microsoft.com/api/stac/v1/collections/landsat-c2-l2/items/LC09_L2SR_083122_20221230_02_T2', 'type': 'application/geo+json'}
# {'rel': 'cite-as', 'href': 'https://doi.org/10.5066/P9OGBGM6', 'title': 'Landsat 8-9 OLI/TIRS Collection 2 Level-2'}
# {'rel': 'via', 'href': 'https://landsatlook.usgs.gov/stac-server/collections/landsat-c2l2-sr/items/LC09_L2SR_083122_20221230_20230101_02_T2_SR', 'type': 'application/json', 'title': 'USGS STAC Item'}
# {'rel': 'preview', 'href': 'https://planetarycomputer.microsoft.com/api/data/v1/item/map?collection=landsat-c2-l2&item=LC09_L2SR_083122_20221230_02_T2', 'type': 'text/html', 'title': 'Map of item'}
```

pjhartzell:
**Related Issue(s):**
- Closes #937 

**Description:**
Enables filtering with `get_links()` on either `rel`, `media_type`, or both.

**PR Checklist:**

### DIFF
--- a/pystac/stac_object.py
+++ b/pystac/stac_object.py
@@ -83,23 +83,30 @@ class STACObject(ABC):
 
     def get_single_link(
         self,
-        rel: Union[str, pystac.RelType],
+        rel: Optional[Union[str, pystac.RelType]] = None,
         media_type: Optional[Union[str, pystac.MediaType]] = None,
     ) -> Optional[Link]:
-        """Get single link that matches the given ``rel`` and, optionally,
-        ``media_type``. If ``media_type`` is ``None``, then the link is
-        matched only on the ``rel`` value.
+        """Get a single :class:`~pystac.Link` instance associated with this
+        object.
 
         Args:
-             rel : The :class:`~pystac.Link` ``rel`` to match on.
-             media_type: The :class:`~pystack.MediaType` ``media_type`` to match on
-        """
+            rel : If set, filter links such that only those
+                matching this relationship are returned.
+            media_type: If set, filter the links such that only
+                those matching media_type are returned
 
+        Returns:
+            Optional[:class:`~pystac.Link`]: First link that matches ``rel``
+                and/or ``media_type``, or else the first link associated with
+                this object.
+        """
+        if rel is None and media_type is None:
+            return next(iter(self.links), None)
         return next(
             (
                 link
                 for link in self.links
-                if link.rel == rel
+                if (rel is None or link.rel == rel)
                 and (media_type is None or link.media_type == media_type)
             ),
             None,
--- a/pystac/stac_object.py
+++ b/pystac/stac_object.py
@@ -86,7 +86,7 @@ class STACObject(ABC):
         rel: Union[str, pystac.RelType],
         media_type: Optional[Union[str, pystac.MediaType]] = None,
     ) -> Optional[Link]:
-        """Get single link that match the given ``rel`` and, optionally,
+        """Get single link that matches the given ``rel`` and, optionally,
         ``media_type``. If ``media_type`` is ``None``, then the link is
         matched only on the ``rel`` value.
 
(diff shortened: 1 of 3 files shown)


pystac/stac_object.py
```diff
@@ -86,7 +86,7 @@ def get_single_link(
         rel: Union[str, pystac.RelType],
```
gadomski, line 86:
```suggestion
        rel: Optional[Union[str, pystac.RelType]],
```

We check for `rel is None` below, so we should declare that it could be optional.

pystac/stac_object.py
```diff
@@ -86,7 +86,7 @@ def get_single_link(
...
         media_type: Optional[Union[str, pystac.MediaType]] = None,
     ) -> Optional[Link]:
-        """Get single link that match the given ``rel`` and, optionally,
+        """Get single link that matches the given ``rel`` and, optionally,
```
gadomski, line 89:
```suggestion
        """Get single link that matches the given ``rel`` and
```

Since both are optional, this "optionally" doesn't make much sense.

pystac/stac_object.py
```diff
@@ -86,7 +86,7 @@ def get_single_link(
...
         ``media_type``. If ``media_type`` is ``None``, then the link is
         matched only on the ``rel`` value.
```
gadomski, line 91:
```suggestion
        ``media_type``. If ``media_type`` is ``None``, then the link is
        matched only on the ``rel`` value, and if ``rel`` is ``None``, then
        the link is matched only on the ``media_type`` value.
```

gadomski (changes requested):

pystac/stac_object.py
pjhartzell: Other than the doc string edit, I didn't monkey with the `get_single_link`. But maybe we should.

gadomski: Whoops, that's my fault -- I didn't track that the docstring was for a different function. Disregard my comment.

pystac/stac_object.py
gadomski: Same as above, disregard.

pystac/stac_object.py
pjhartzell: It seems appropriate to modify this function, though, to make it consistent with `get_links`. I'm going to implement.

gadomski (approved):

PR #967, opened 2023-01-26
Make `layout_strategy` default value consistent

### LINKED ISSUE #965: Make layout strategy arguments consistent
As described in https://github.com/stac-utils/pystac/pull/919#issuecomment-1403752882, there's some alignment needed in how layout strategies are passed in to methods. Specifically:

- `Catalog.add_item` has `None` for a default layout strategy, but functionally it defaults to `BestPracticesLayoutStrategy`
- `Catalog.add_item`'s strategy argument is undocumented
- `Catalog.add_items` could/should take a similar strategy argument
- Same for `Collection` methods of the same name

pjhartzell:
**Related Issue(s):** #
- Closes #965 

**Description:**
- Updates docstrings to make users aware of the default layout strategy.
- Adds an optional `strategy` parameter to the catalog `add_items()` method.
- Updates docstring formatting in several places to correct rendering in the API docs.

**PR Checklist:**

### DIFF
--- a/pystac/catalog.py
+++ b/pystac/catalog.py
@@ -230,7 +230,7 @@ class Catalog(STACObject):
         self,
         child: Union["Catalog", "Collection_Type"],
         title: Optional[str] = None,
-        strategy: HrefLayoutStrategy = BestPracticesLayoutStrategy(),
+        strategy: Optional[HrefLayoutStrategy] = None,
     ) -> None:
         """Adds a link to a child :class:`~pystac.Catalog` or
         :class:`~pystac.Collection`. This method will set the child's parent to this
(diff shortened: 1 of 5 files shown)


pjhartzell:
@gadomski This went a bit deeper than I initially perceived, which gives me some pause about the changes.

pystac/catalog.py
```diff
@@ -230,7 +230,7 @@ def add_child(
         self,
         child: Union["Catalog", "Collection_Type"],
         title: Optional[str] = None,
-        strategy: Optional[HrefLayoutStrategy] = None,
+        strategy: HrefLayoutStrategy = BestPracticesLayoutStrategy(),
```
gadomski, line 233: Yeah, constructing as the default value is generally dangerous -- it leaves us open to shared-mutable-object concerns. We probably want to keep this as `None` default, and just document the behavior better.

gadomski (changes requested):

pystac/catalog.py
pjhartzell: Erf. Yeah. Thanks. Will update.

gadomski (approved):

PR #972, opened 2023-01-31
Preserve relative links when using `from_file`

### LINKED ISSUE #838: Collection link points to local path
While I see a lot of code that normalizes asset hrefs and make them relative whenever required, it seems I always end up with a local file path for the collection link. I think the collection link should be normalized as well and ideally be relative to the self link.

gadomski:
**Related Issue(s):**
- Closes #838 

**Description:**

When reading Catalogs/Collections from files, a user should be able to preserve any relative links e.g. to children. Otherwise, re-saving the Catalog/Collection may produce surprising results (e.g. filesystem-absolute links where previously they were relative, breaking portability).

This is currently a draft PR demonstrating the problem. The fix will likely be adding a `catalog_type` argument to `from_file`, but we should explore other options as well, since adding a new argument to such a widely used classmethod is not ideal.

**PR Checklist:**

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -8,15 +8,15 @@ from .utils import ARBITRARY_BBOX, ARBITRARY_EXTENT, ARBITRARY_GEOM
 
 
 @pytest.fixture
-def test_catalog() -> Catalog:
+def catalog() -> Catalog:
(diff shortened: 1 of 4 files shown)


gadomski:
Closing, as this test wasn't exactly the issue from OP.

PR #983, opened 2023-02-10
Generalize return types from classmethods

### LINKED ISSUE #862: alternative constructors shouldn't hardcode their return types
Currently, alternative constructors like `Collection.from_dict` define the return type as `"Collection"`. https://github.com/stac-utils/pystac/blob/2e76a877b02eccf4d8e8ec6ad7b4aa7cf09aa268/pystac/collection.py#L607-L615

This makes is hard for subclasses to properly override the methods, since they'll be returning the subclasses type.

See something like https://stackoverflow.com/a/44644576/1889400 or https://mypy.readthedocs.io/en/stable/more_types.html?highlight=classmethod#precise-typing-of-alternative-constructors.

### LINKED ISSUE #863: `clone` shouldn't hard-code return type
In methods like https://github.com/stac-utils/pystac/blob/2e76a877b02eccf4d8e8ec6ad7b4aa7cf09aa268/pystac/collection.py#L100-L102, `.clone()` uses the class itself.

```python
class Extent:
    def clone(self):
        return Extent(...)
```

Instead of hard-coding `Extent(...)` in the return there, we should use `type(self)(...)` to ensure that a subclass is returned.

xref https://github.com/stac-utils/pystac/issues/862, which also relates to subclasses.

gadomski:
**Related Issue(s):**

- Closes #862 
- Closes #863 

**Description:**

Use `TypeVar` for alternate constructors, e.g. `from_dict` and `from_file`. The key `cast` that enables the whole inheritance tree to work is https://github.com/stac-utils/pystac/compare/issues/862-return-types?expand=1#diff-3f0f0b208e42045d63bb010a3b963e4178a8cf882f75f73d8392828b2190eafcR525.

Includes some sidecar fixes:
- Allow subclass instantiation in `clone` in a couple of places where it was missing, closing #863 
- Fix up a couple more `from __future__ import annotations` opportunities that were missed in #962 
- Begin using simple test fixtures, now that we have **pytest** from #939 
- Fix a couple of lints that I flagged locally with [ruff](https://github.com/charliermarsh/ruff)

**PR Checklist:**


pystac/collection.py
```diff
@@ -97,7 +99,8 @@ def clone(self) -> SpatialExtent:
         Returns:
             SpatialExtent: The clone of this object.
         """
-        return SpatialExtent(
+        cls = self.__class__
+        return cls(
```
pjhartzell, line 103: If subclassed, I see how `cls = self.__class__` is the subclass. But are we concerned with the return being typed as the parent class (`SpatialExtent`) rather than the subclass? I'm trying to reconcile this with [#862](https://github.com/stac-utils/pystac/issues/862).

pystac/validation/__init__.py
```diff
@@ -164,7 +164,7 @@ class RegisteredValidator:
     def get_validator(cls) -> STACValidator:
         if cls._validator is None:
             try:
-                import jsonschema
+                import jsonschema as _  # noqa
```
pjhartzell, line 167: Why? Just to indicate it is unused in this module?

pystac/common_metadata.py
```diff
@@ -23,17 +25,17 @@ class CommonMetadata:
         properties : Dictionary of attributes that is the Item's properties
```
pjhartzell, line 25: How is `properties` an `Arg` here?

pjhartzell (changes requested):
Looks good. Just a few questions.

pystac/validation/__init__.py
gadomski: Yeah, my [ruff](https://github.com/charliermarsh/ruff) linter was complaining about it, so this just makes it more explicit that it isn't being used.

pystac/common_metadata.py
pjhartzell: >IMO a fix doesn't belong in this PR -- we should open another issue, or find an existing documentation issue and link this problem in.

Yep, that's fine. I'm just not seeing a `properties` attribute or an argument/parameter named `properties` for `__init__()`, so not sure why it's in the docstring at all.

gadomski: I thinks it's referencing the `object` attribute -- for sure it's not right.

pjhartzell (approved):

PR #985, opened 2023-02-14
Always use ``from datetime import datetime``

jsignell:
**Description:** 
Just a minor tidy. The datetime import varied across the codebase so I took the opportunity to consolidate around `from datetime import datetime`. 

**PR Checklist:**

### DIFF
--- a/pystac/item.py
+++ b/pystac/item.py
@@ -28,6 +28,7 @@ from pystac.utils import (
 T = TypeVar("T", bound="Item")
 
 if TYPE_CHECKING:
+    # avoids conflicts since there are also kwargs and attrs called `datetime`
     from datetime import datetime as Datetime
 
 
--- a/pystac/item.py
+++ b/pystac/item.py
@@ -1,9 +1,8 @@
 from __future__ import annotations
 
 from copy import copy, deepcopy
-from datetime import datetime as Datetime
 from html import escape
-from typing import Any, Dict, List, Optional, Type, TypeVar, Union, cast
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Type, TypeVar, Union, cast
 
 import pystac
 from pystac import STACError, STACObjectType
@@ -28,6 +27,9 @@ from pystac.utils import (
 
 T = TypeVar("T", bound="Item")
 
+if TYPE_CHECKING:
+    from datetime import datetime as Datetime
+
 
 class Item(STACObject):
     """An Item is the core granular entity in a STAC, containing the core metadata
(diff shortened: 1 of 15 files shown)


pystac/item.py
```diff
@@ -28,6 +27,9 @@
 
 T = TypeVar("T", bound="Item")
 
+if TYPE_CHECKING:
+    from datetime import datetime as Datetime
```
gadomski, line 31: Any reason not to keep the lowercase version here, as done elsewhere?

```suggestion
    from datetime import datetime
```

gadomski (changes requested):
👍🏽 to Doing Things One Way, thanks! I have one inline question, and then one top-level change requested and one top-level question. Change:

- I generally like to record these sorts of decisions in a style guide -- since PySTAC doesn't have one (yet), can you either (a) create one w/ this decision, or (b) add a style section to https://github.com/stac-utils/pystac/blob/main/docs/contributing.rst and add this decision?

And question:

- Any reason why you chose to consolidate around `from datetime import datetime`, instead of `import datetime`? I _usually_ prefer classes/constants without their namespace, and free functions with a single namespace, but since `datetime.datetime` is a lowercase class I find myself just `import datetime`ing more often. I don't have a problem with choosing this way, I was just curious if there was any deep background behind the decision (no problem if not).

pystac/item.py
jsignell: Oh yeah I should have mentioned this. I _did_ try changing this to `from datetime import datetime` but I ran into conflicts since there are also kwargs and attrs called `datetime` and in the file.

jsignell:
>   * I generally like to record these sorts of decisions in a style guide -- since PySTAC doesn't have one (yet), can you either (a) create one w/ this decision, or (b) add a style section to https://github.com/stac-utils/pystac/blob/main/docs/contributing.rst and add this decision?

Absolutely. Good call.

>   * Any reason why you chose to consolidate around `from datetime import datetime`, instead of `import datetime`? I _usually_ prefer classes/constants without their namespace, and free functions with a single namespace, but since `datetime.datetime` is a lowercase class I find myself just `import datetime`ing more often. I don't have a problem with choosing this way, I was just curious if there was any deep background behind the decision (no problem if not).

Yeah I did a brief google to see if there was some consensus around which style is preferable. I have also seen `import datetime as dt` - which seems like it has a bit of a following: https://import-as.github.io/import/datetime.html. I ended up going with `from datetime import datetime` based on the datetime docs: https://docs.python.org/3/library/datetime.html. I am definitely not tied to this decision, just in it for the consistency.

pystac/item.py
gadomski: Ok, can you then throw a comment on that import w/ that explanation? If we're documenting our decisions w/ a style guide we should call out deviations from that guide.

gadomski:
> I ended up going with from datetime import datetime based on the datetime docs: https://docs.python.org/3/library/datetime.html.

I think this is a great rationale, and would be good to include in the style guide writeup. Thanks!

> I have also seen import datetime as dt - which seems like it has a bit of a following:

We all have our personal preferences, and I'm over on the side of "don't shorten names", so much the point that I do `import numpy` instead of `import numpy as np`. But that's me :-) ... long story short I'm 👍🏽 on `from datetime import datetime` and 👎🏽 on `import datetime as dt`.

gadomski (approved):

Issue #949, opened 2023-01-12
Avoid network calls in tests suite


pjhartzell:
Make test suite more robust by eliminating network calls. There is at least [one network call](https://github.com/stac-utils/pystac/blob/d00ac702e2d19d3b36fd3d066c5310579c1e12fe/tests/test_catalog.py#L116) that can be avoided by using something like [pytest-vcr](https://pytest-vcr.readthedocs.io/en/latest/) once the switch to `pytest` (#939) is complete.

PR #986, opened 2023-02-14
Add `RetryStacIO`

### LINKED ISSUE #958: New feature: StacIO that retries network requests
## Summary

Add a `RetryStacIO` (probably inheriting from `DefaultStacIO`) that retries network requests in a configurable way. It'll probably be modeled on [urllib3](https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.Retry), though I am _not_ suggesting we add **urllib3** as a dependency. E.g.

```python
from pystac import Item
from pystac.stac_io import RetryStacIO
url = "https://planetarycomputer.microsoft.com/api/stac/v1/collections/aster-l1t/items/AST_L1T_00310012006175412_20150516104359"
item = Item.from_file(url, RetryStacIO())  #  This will retried with default settings
item = Item.from_file(url, RetryStacIO(total=3, backoff_factor=2))   # Retry can be configured

# If you want to enable retries for all PySTAC operations
from pystac import StacIO
StacIO.set_default(RetryStacIO)
```

## Motivation

When doing a large number of operations with PySTAC, sometimes in parallel, it's possible to overwhelm servers or otherwise get transient errors. My specific example involved `Item.to_dict` -- I forgot to specify `transform_hrefs=False` (clapback to https://github.com/stac-utils/pystac/issues/546#issuecomment-911050197), and did this ~1 million times:

```python
d = item.to_dict(include_self_href=False)
```

Under the hood, this does a network request to resolve the root, and those requests would sometimes error out. I solved my specific problem by disabling href transforms:

```python
d = item.to_dict(include_self_href=False, transform_hrefs=False)
```

But a retry-enabled PySTAC would be useful in any large-scale/batch processing context.

## Alternatives

We could add **urllib3** as a dependency and use it's `Retry` explicitly. I don't think that's a terrible thing -- **urllib3** is widely used and popular. **requests** is built on it, so adding **urllib3** could represent a gradual slide towards just biting the bullet and adding **requests** too.

gadomski:
**Related Issue(s):**

- Closes #958 

**Description:**

Though I initially thought I could do this without **urllib3**, as I was implementing I decided it was silly to just copy the functionality from **urllib3**, since it's so commonly on systems w/ PySTAC. So I added it as an `extras_require`. The **urllib3** typing story is a bit of a mess, so I had to use a couple of `# type: ignore`s. Testing retries is more trouble than its worth IMO, so I only included smoke tests for reading with the new class.

**Background:**

[**requests** uses **urllib3**](https://github.com/psf/requests/blob/15585909c3dd3014e4083961c8a404709450151c/setup.py#L64), so I consider **urllib3** a pretty no-brainer library to optionally depend on -- it's everywhere.

**Sidecar:**

- Includes two small doc fixes I found along the way
- Casually adds `pystac-vcr` as a test dependency, with an eye towards #949 
- I rearranged `requirements-test.txt` because its arbitrary layout was getting to me ... now it's just a sorted list

**PR Checklist:**


pjhartzell (approved):
Looks good.

Issue #872, opened 2022-08-10
table extension Column is not JSON serializable


martinfleis:
When trying to use the `table` extension, I think I hit a bug (or I don't know how to use it). See the example:

```py
from datetime import datetime

import pystac
from pystac.extensions.table import TableExtension, Column

catalog = pystac.Catalog(
    id="test_catalog",
    description="Test catalog",
    )

# create Item
item = pystac.Item(id='countries',
                 geometry=None,
                 bbox=[0, 1, 1, 2],
                 datetime=datetime.utcnow(),
                 href="./data.json",
                 properties={},
                 )

# add column metadata
tab_ext = TableExtension.ext(item, add_if_missing=True)
tab_ext.columns = [
    Column({"name": 'col_1', "type": 'str'}),
    Column({"name": 'col_2', "type": 'byte_array'}),
]

# add asset
item.add_asset(
    key='data',
    asset=pystac.Asset(
        title="Dataset root",
        href="./data.parquet",
        media_type="application/x-parquet"
    )
)

# add to collection
catalog.add_item(item)

catalog.normalize_hrefs("stac/")
catalog.save(catalog_type=pystac.CatalogType.SELF_CONTAINED)
```
The last line raises

```py
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Untitled-1.ipynb Cell 4 in <cell line: 1>()
----> 1 catalog.save(catalog_type=pystac.CatalogType.SELF_CONTAINED)

...

TypeError: Object of type Column is not JSON serializable
```
I  believe that this mirrors the issue #605 and change I made on this branch (https://github.com/stac-utils/pystac/compare/main...martinfleis:pystac:table_serialize) fixes it. Happy to turn it into a PR if that is an actual bug and not my misunderstanding of the API.

m-mohr:
I'm running into the same issue...

PR #992, opened 2023-02-15
Make table extension columns serializable

jsignell:
**Related Issue(s):** #872

**Description:**
The table extension was not serializable because the list of columns was not getting serialized properly. This PR takes inspiration from the original recommendation (thanks @martinfleis!) but pushes the change further back into the mixin class. 

**Note:**
The `_set_property` functionality seems pretty useful overall and I was wondering if it should be use more thoroughly or if there is a reason not to use it. 

**PR Checklist:**


tests/extensions/test_table.py
```diff
@@ -59,3 +59,21 @@ def test_should_raise_exception_when_passing_invalid_extension_object(
...
+        ]
+        tab_ext.columns = columns
+        item.save_object()
+        self.assertTrue(all(isinstance(c, Column) for c in tab_ext.columns))
+        self.assertTrue(
+            all(
+                before.properties == after.properties
+                for before, after in zip(columns, tab_ext.columns)
+            )
+        )
```
gadomski, line 79:
```suggestion
 def test_item_with_table_extension_is_serilalizable_and_roundtrips() -> None:
      item = pystac.Item.from_file(self.example_uri)
      # add column metadata
      tab_ext = TableExtension.ext(item, add_if_missing=True)
      columns = [
          Column({"name": "col_1", "type": "str"}),
          Column({"name": "col_2", "type": "byte_array"}),
      ]
      tab_ext.columns = columns
      item.save_object()
      assert all(isinstance(c, Column) for c in tab_ext.columns)
      assert (
          all(
              before.properties == after.properties
              for before, after in zip(columns, tab_ext.columns)
          )
      )
```

Since #939 we have **pytest**; while we haven't backported our existing tests away from `unittest.TestCase` format yet, I think it's worth starting to define new tests as free functions to reduce our future work. I've opened https://github.com/stac-utils/pystac/issues/993 to capture the need to switch over the whole test suite (no small thing).

gadomski (changes requested):
Agreed that the new functionality in the base class is useful; previous fixes (e.g. https://github.com/stac-utils/pystac/blob/781972855e2dbd31b71b5e4787fcec59185d3e09/pystac/extensions/raster.py#L693-L697) do the same thing, but at the extension level. IMO it'd be worth opening a new issue after this PR lands to convert any other extension-specific `_set_property` magic to use this new base-class functionality.

tests/extensions/test_table.py
```diff
@@ -59,3 +59,21 @@ def test_should_raise_exception_when_passing_invalid_extension_object(
...
+    def test_item_with_table_extension_is_serilalizable_and_roundtrips(self) -> None:
+        item = pystac.Item.from_file(self.example_uri)
+        # add column metadata
+        tab_ext = TableExtension.ext(item, add_if_missing=True)
+        columns = [
+            Column({"name": "col_1", "type": "str"}),
+            Column({"name": "col_2", "type": "byte_array"}),
+        ]
+        tab_ext.columns = columns
+        item.save_object()
```
gadomski, line 72: One other change please -- currently this overwrites the test data file, leading to a git diff after running tests. Coupled with the change to the free function (https://github.com/stac-utils/pystac/pull/992/files#r1107712387), could you use [tmp_path](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html) to save the new item to a temporary location, so we don't modify the code tree? Thanks!

tests/extensions/test_table.py
jsignell: :100: Oh thank god. I am way more comfortable in pytest.

gadomski (approved):

Issue #831, opened 2022-06-24
Item from_dict throws KeyError if fields don't exist, disallowing read of invalid Item JSON


philvarner:
The Item.from_dict method uses the dictionary "pop" method for retrieving most fields from the source dict, resulting in a KeyError if they don't exist.

The conceptual issue here is that from_dict doesn't allow the creation on an invalid Item from an invalid JSON/dict representation, whereas manually invoking the __init__ method of Item is typed to and does allow the creation on invalid items:

```
    def __init__(
        self,
        id: str,
        geometry: Optional[Dict[str, Any]],
        bbox: Optional[List[float]],
        datetime: Optional[Datetime],
        properties: Dict[str, Any],
        stac_extensions: Optional[List[str]] = None,
        href: Optional[str] = None,
        collection: Optional[Union[str, Collection]] = None,
        extra_fields: Optional[Dict[str, Any]] = None,
```

If the from_dict method is intended to disallow the creation of an Item from an invalid dict representation, it should throw a specific exception to this effect, not the accidental and non-contextualize KeyError.

jsignell:
I can take a look at this. It seems like adding a default to these `pop`s would be a sane and low-impact place to start.

Issue #991, opened 2023-02-15
Retrieve MAXAR Open Data -  STACTypeError: JSON does not represent a STAC object


giswqs:
I am trying to retrieve the [MAXAR Open Data Catalog](https://stacindex.org/catalogs/maxar-open-data-catalog-ard-format#/) using the the url  `https://maxar-opendata.s3.amazonaws.com/events/catalog.json`. However, I can't get it to work with pystac or pystac-client. What's the recommended way to read a stac catalog like this?

I noticed that the [stac-browser website](https://radiantearth.github.io/stac-browser/#/external/maxar-opendata.s3.amazonaws.com/events/catalog.json) was able to crawl the collections and items using the same catalog URL shown above. How can we do this using Python?

```python
import pystac
url = 'https://maxar-opendata.s3.amazonaws.com/events/catalog.json'
pystac.read_file(url)
```

> STACTypeError: JSON does not represent a STAC object.

```python
from pystac_client import Client
url = 'https://maxar-opendata.s3.amazonaws.com/events/catalog.json'
client = Client.open(url)
```

> ClientTypeError: Could not open Client (href=https://maxar-opendata.s3.amazonaws.com/events/catalog.json), expected type=Catalog, found type=None

drwelby:
I build those catalogs, so let me know if I need to adjust anything.

gadomski:
Looks like that endpoint is invalid STAC:

```shell
$ curl -s https://maxar-opendata.s3.amazonaws.com/events/catalog.json | jq '{ type: .type, stac_version: .stac_version }'
{
  "type": null,
  "stac_version": "1.0.0"
}
```

[`type` is a required field in STAC v1.0](https://github.com/radiantearth/stac-spec/blob/master/catalog-spec/catalog-spec.md#catalog-fields). STAC v0.6 [did not require the `type` field](...), so one way to open the catalog would be to change the `stac_version` attribute, e.g.:

```shell
$ curl -s https://maxar-opendata.s3.amazonaws.com/events/catalog.json | jq '.stac_version = "0.6.0"' > catalog.json && python -c "import pystac; print(pystac.read_file('catalog.json'))"
<Catalog id=maxar-opendata>
```

However, the core issue should be files w/ MAXAR, as it looks like their endpoint is not producing valid STAC.

drwelby:
Root collection now has a "type".

gadomski:
@drwelby needs to be `Catalog` (it's currently `catalog`).

[...]

PR #994, opened 2023-02-15
Add defaults to all the `Item.from_dict` pops

jsignell:
**Related Issue(s):** #831

**Description:**
The easy solution to better error handling is to just defer until the init. So I added pop defaults to all the dict fields _even if they are required_

**PR Checklist:**

### DIFF
--- a/pystac/item.py
+++ b/pystac/item.py
@@ -423,6 +423,9 @@ class Item(STACObject):
         migrate: bool = False,
         preserve_dict: bool = True,
     ) -> T:
+        if preserve_dict:
+            d = deepcopy(d)
+
         if migrate:
             info = identify_stac_object(d)
             d = migrate_to_latest(d, info)
@@ -432,34 +435,39 @@ class Item(STACObject):
                 f"{d} does not represent a {cls.__name__} instance"
             )
 
-        if preserve_dict:
-            d = deepcopy(d)
+        # some fields are passed through to __init__
+        pass_through_fields = [
+            "id",
+            "geometry",
+            "bbox",
+            "stac_extensions",
+            "collection",
+        ]
 
-        id = d.pop("id")
-        geometry = d.pop("geometry")
-        properties = d.pop("properties")
-        bbox = d.pop("bbox", None)
-        stac_extensions = d.get("stac_extensions")
-        collection_id = d.pop("collection", None)
-
+        # some fields need some parsing or special handling
+        parse_fields = ["links", "assets", "properties"]
+        links = d.get("links", [])
+        assets = d.get("assets", {})
+        properties = d.get("properties", {})
         datetime = properties.get("datetime")
         if datetime is not None:
             datetime = str_to_datetime(datetime)
-        links = d.pop("links")
-        assets = d.pop("assets")
 
-        d.pop("type")
-        d.pop("stac_version")
+        # some fields are excluded from __init__ entirely
+        exclude_fields = ["type", "stac_version"]
+
+        # get all the fields that are not passed through, parsed, or excluded
+        extra_fields = {
+            k: v
+            for k, v in d.items()
+            if k not in [*pass_through_fields, *parse_fields, *exclude_fields]
+        }
 
         item = cls(
-            id=id,
-            geometry=geometry,
-            bbox=bbox,
+            **{k: d.get(k) for k in pass_through_fields},  # type: ignore
             datetime=datetime,
             properties=properties,
-            stac_extensions=stac_extensions,
-            collection=collection_id,
-            extra_fields=d,
+            extra_fields=extra_fields,
             href=href,
             assets={k: Asset.from_dict(v) for k, v in assets.items()},
         )
@@ -499,6 +507,12 @@ class Item(STACObject):
 
     @classmethod
     def matches_object_type(cls, d: Dict[str, Any]) -> bool:
+        for field in ("type", "stac_version"):
+            if field not in d:
+                raise pystac.STACTypeError(
+                    f"{d} does not represent a {cls.__name__} instance"
+                    f"'{field}' is missing."
+                )
         return identify_stac_object_type(d) == STACObjectType.ITEM
 
     @property
--- a/tests/test_item.py
+++ b/tests/test_item.py
@@ -8,6 +8,7 @@ from copy import deepcopy
 from typing import Any, Dict, Optional
 
 import dateutil.relativedelta
+import pytest
 
 import pystac
 import pystac.serialization.common_properties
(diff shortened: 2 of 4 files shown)


jsignell:
Just noticed the comment on the other PR about pytest. I'll update this now

pystac/item.py
```diff
@@ -435,21 +435,21 @@ def from_dict(
...
             datetime = str_to_datetime(datetime)
-        links = d.pop("links")
-        assets = d.pop("assets")
+        links = d.pop("links", [])
+        assets = d.pop("assets", {})
 
-        d.pop("type")
-        d.pop("stac_version")
+        d.pop("type", None)
+        d.pop("stac_version", None)
```
gadomski, line 452: I _think_ we're assuming that these are populated by `migrate_to_latest` on L428 -- can you either
1. add a comment to that effect,
2. add an assertion (or other error) just to make _sure_ they're populated, or
3. correct me?

Thanks!

gadomski (changes requested):

pystac/item.py
jsignell: Oh! I just added these kind of blindly. I think that you are right that type and version _have_ to be provided otherwise you don't make it past `cls.matches_object_type`. But I also don't see the harm in adding None here since we are just dropping them on the floor anyway.

gadomski: I guess part of the point of this exercise is to allow `Item.__init__` to do its sort-of-validation (e.g. checking for `datetime` OR `start_datetime` OR `end_datetime`, etc), and by switching from `KeyError` to silently dropping, we lose a bit of validation IF (somehow) a dict made it through without a `type`.

Again, I don't think it can make it through, but I often like to at least `assert` these sorts of things to make it clear to my future self, who might not remember that `matches_object_type` adds/catches those.

gadomski:
I'm going to link this with https://github.com/stac-utils/pystac/issues/831, even though it doesn't implement the fix _exactly_ as recommended by @philvarner. IMO we don't want to doing psuedo-validation in multiple places, so if there is going to be validity checks, they should probably be in the constructor (not in `from_dict`). If folks want to do more checking in the constructor (e.g. ensuring that `bbox` is defined if `geometry` is), that should be done in a separate PR.

gadomski:
@philvarner I don't think the current code is that far from what you're describing. The `Item` constructor is doing some validation (e.g. the datetime stuff), so all we'd need to add is More Validation™ to get to your goal of a nice, strict, constrained class.

I'm not sure of the value of `MaybeItem` -- I think `Dict[str, Any]` is basically a `MabyeItem`. I don't quite see the utility in this (which is what I think you mean by `MabyeItem`):

```python
class MabyeItem:
    type: Optional[Any]
    stac_version: Optional[Any]
    ...
```

Am I misunderstanding `MabyeItem`?

philvarner:
I was thinking of it a bit more strict than that -- the fields value are all Optional, but of the correct type. 

class MaybeItem:
    type: Optional[str]
    stac_version: Optional[str]
    ...

The use case here is the same one that Item is currently failing on -- I want to read an invalid item and then presumably do something with it while in an invalid state or modify it to be valid. When I'm modifying it, it's a nicer experience to modify the object and get the typechecking on it rather than operating on an dict[str, Any] with no checking.

gadomski:
> When I'm modifying it, it's a nicer experience to modify the object and get the typechecking on it rather than operating on an dict[str, Any] with no checking.

Fair, and makes sense. There is a tradeoff w/ complexity b/c we'd be adding another, slightly ambiguous item to the namespace. I do see the value in providing _some_ sort of support for invalid STAC -- without it, we end up with problems like https://github.com/stac-utils/pystac/issues/991, where it's totally opaque to a user what the problem is w/ a STAC read operation.

My instinct is to try to ring-fence the `Maybe*` classes in the IO space (e.g. keep them in `pystac.io`), since you'd presumably be using them exclusively with stuff you find "in the wild" (i.e. things you read in). When you're making STAC yourself, you shouldn't need those potentially-invalid structures, right?

pystac/item.py
jsignell: I added an error raise to `matches_object_type` for the case where the dict is invalid. Let me know if that feels weird. I really wanted to add the context of what field is missing that makes the dict invalid as an `Item`

pystac/item.py
```diff
@@ -423,6 +423,9 @@ def from_dict(
         migrate: bool = False,
         preserve_dict: bool = True,
     ) -> T:
+        if preserve_dict:
```
philvarner, line 426: not a fan of the re-assign to a method param name, especially one just named `d`

pystac/item.py
```diff
@@ -431,25 +434,22 @@ def from_dict(
...
         datetime = properties.get("datetime")
         if datetime is not None:
             datetime = str_to_datetime(datetime)
-        links = d.pop("links")
-        assets = d.pop("assets")
-
-        d.pop("type")
-        d.pop("stac_version")
+        links = d.pop("links", [])
+        assets = d.pop("assets", {})
```
philvarner, line 453: I think a more clear approach here would be to use `in` or `get` for accessing all of the fields that are expected, and then explicitly removing all of the ones that we don't want to be in extra_fields in one go, like:

```
for k in ["type", "stac_version", ....]:
  d.pop(k, None)
```

Reason is, I think it's unclear to someone reading this that the pops are both checking that the field is defined **and** modifying the dict to be set in the extra_fields. Separating the operations to validate the fields and modify the fields is less likely to cause bugs in the future if someone adds other checks that use pop or get, but wasn't thinking about how that effects the extra_fields.

tests/test_item.py
```diff
@@ -436,3 +437,15 @@ def from_dict(
             return super().from_dict(d)
 
     _ = CustomItem.from_dict(test_item.to_dict())
+
```
philvarner, line 440: add a test for when `type` is not in the dict

pystac/item.py
```diff
@@ -431,25 +434,22 @@ def from_dict(
             raise pystac.STACTypeError(
                 f"{d} does not represent a {cls.__name__} instance"
             )
+        # These fields must exist on the dict for `matches_object_type` to pass
+        d.pop("type")
```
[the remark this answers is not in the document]
philvarner: I see now that these will always exist if they passed `matches_object_type`, but I think this could be more clear to someone (like me!) reading this code if we didn't mix pop and get, both with and without defaults and removed all the fields in one place.

philvarner (changes requested):
several comments.

pystac/item.py
jsignell: I just moved this block up in the method. As it was prior the dict would be altered if `migrate=True` which seemed to go against the intent of the kwarg.

pystac/item.py
jsignell: I totally agree that it is not clear from reading this the first time what is going on. I also didn't originally understand that the data is being paired down and then fed to extra_fields. I actually think this could do with more of a reorg not that I think about it.

pystac/item.py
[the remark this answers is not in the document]
jsignell: Yeah hopefully it feels clearer now.

pystac/item.py
philvarner: Very nice.

philvarner (approved):

PR #1006, opened 2023-02-23
Add deprecation warnings

### LINKED ISSUE #843: Surface warnings when accessing deprecated collections #156
https://github.com/stac-extensions/version offers a mechanism for marking a collection or item as deprecated.

We at the Planetary Computer are deprecating our `landsat-8-c2-l2` collection in favor of `landsat-c2-l2`. I would love it if querying a deprecated collection or item would surface a (Python) Depracation / FutureWarning

```python
>>> catalog = pystac_client.Client.open("https://planetarycomputer.microsoft.com/api/stac/v1")
>>> landsat = catalog.get_collection("landsat-8-c2-l2")
FutureWarning("The collection 'landsat-8-c2-l2' is deprecated. Use 'landsat-c2-l2' instead.")
```

pystac-client would build the Collection. If it detects the `version` extension and it's marked as deprecated, then you'd look for a `latest` link and use it as the alternative.

Ideally something similar with `search`, but I don't know 100% know how that would work. We probably don't want to emit 100 warnings for a search with 100 deprecated items.

This should probably be easily configurable by an environment variable and perhaps a global variable. Users can do stuff with the warnings registry, but that's a bit unweildy.

re DeprecationWarning vs. FutureWarning, I don't know what best practice is these days. Previous versions of Python automatically silenced DeprecationWarnings that weren't running under `__main__` so I've always used FutureWarning for stuff I want users to see. That might have changed in recent versions of python.

(Transferred from https://github.com/stac-utils/pystac-client/issues/156)

pjhartzell:
**Related Issue(s):**

- Closes #843 

**Description:**

- Inserts logic into the Item and Collection `from_dict()` methods that checks for the version extension and, if it exists, issues a `DeprecatedWarning` if the version extension `deprecated` field is set to `True`.
- Provides an `ignore_deprecated` context manager to suppress the `DeprecatedWarning`.

This PR originally used logic outside the version extension to check for the `deprecated` field on assets. This was removed in favor of working within the version extension's current deprecation checking ability.

**PR Checklist:**

### DIFF
--- a/benchmarks/item.py
+++ b/benchmarks/item.py
@@ -15,7 +15,8 @@ class ItemBench(Bench):
 
         self.stac_io = StacIO.default()
 
-        self.item_path = get_data_path("item/sample-item-asset-properties.json")
+        # using an item with many assets to better test deserialization timing
+        self.item_path = get_data_path("eo/eo-sentinel2-item.json")
         with open(self.item_path) as src:
             self.item_dict = json.load(src)
         self.item = Item.from_file(self.item_path)
--- a/pystac/collection.py
+++ b/pystac/collection.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import warnings
 from copy import deepcopy
 from datetime import datetime
 from html import escape
@@ -23,7 +24,7 @@ import pystac
 from pystac import CatalogType, STACObjectType
 from pystac.asset import Asset
 from pystac.catalog import Catalog
-from pystac.errors import STACTypeError
+from pystac.errors import DeprecatedWarning, ExtensionNotImplemented, STACTypeError
 from pystac.html.jinja_env import get_jinja_env
 from pystac.layout import HrefLayoutStrategy
 from pystac.link import Link
@@ -618,6 +619,8 @@ class Collection(Catalog):
         migrate: bool = False,
         preserve_dict: bool = True,
     ) -> C:
+        from pystac.extensions.version import CollectionVersionExtension
+
         if migrate:
             info = identify_stac_object(d)
             d = migrate_to_latest(d, info)
@@ -677,6 +680,17 @@ class Collection(Catalog):
 
         if root:
             collection.set_root(root)
+
+        try:
+            version = CollectionVersionExtension.ext(collection)
+            if version.deprecated:
+                warnings.warn(
+                    f"The collection '{collection.id}' is deprecated.",
+                    DeprecatedWarning,
+                )
+            # Collection asset deprecation checks pending version extension support
+        except ExtensionNotImplemented:
+            pass
 
         return collection
 
--- a/pystac/item.py
+++ b/pystac/item.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import warnings
 from copy import copy, deepcopy
 from html import escape
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Type, TypeVar, Union, cast
@@ -9,6 +10,7 @@ from pystac import STACError, STACObjectType
 from pystac.asset import Asset
 from pystac.catalog import Catalog
 from pystac.collection import Collection
+from pystac.errors import DeprecatedWarning, ExtensionNotImplemented
 from pystac.html.jinja_env import get_jinja_env
 from pystac.link import Link
 from pystac.serialization import (
@@ -423,6 +425,8 @@ class Item(STACObject):
         migrate: bool = False,
         preserve_dict: bool = True,
     ) -> T:
+        from pystac.extensions.version import ItemVersionExtension
+
         if preserve_dict:
             d = deepcopy(d)
 
@@ -482,6 +486,17 @@ class Item(STACObject):
 
         if root:
             item.set_root(root)
+
+        try:
+            version = ItemVersionExtension.ext(item)
+            if version.deprecated:
+                warnings.warn(
+                    f"The item '{item.id}' is deprecated.",
+                    DeprecatedWarning,
+                )
+            # Item asset deprecation checks pending version extension support
+        except ExtensionNotImplemented:
+            pass
 
         return item
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - `STACObject.remove_hierarchical_links` and `Link.is_hierarchical` ([#999](https://github.com/stac-utils/pystac/pull/999))
 - `extra_fields` to `AssetDefinition` in the item assets extension ([#1003](https://github.com/stac-utils/pystac/pull/1003))
 - `Catalog.fully_resolve` ([#1001](https://github.com/stac-utils/pystac/pull/1001))
+- A `DeprecatedWarning` when deserializing an Item or Collection to a STAC object via the `from_dict()` method ([1006](https://github.com/stac-utils/pystac/pull/1006))
 
 ### Removed
 
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ See the [documentation page](https://pystac.readthedocs.io/en/latest/) for the l
 
 ## Developing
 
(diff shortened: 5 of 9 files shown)


pjhartzell:
>but I'm a little worried about the performance implications

Yeah, me too. I forgot about our benchmark suite! I'll dig in and see what we can compare.

pjhartzell:
I switched to the `tests/data-files/eo/eo-sentinel2-item.json` for the Item bench --> it has 17 assets. After running `./scripts/bench`, the output from `asv compare main HEAD` for timing the `item.from_dict()` methods is below. I did this twice.

**First Run:**

```
       before           after         ratio
     [f78637a9]       [6a1b1cf7]
     <main>           <issues/843-deprecation-warnings>
          286±1μs          286±1μs     1.00  item.ItemBench.time_item_from_dict [Prestons-MacBook-Pro.local/virtualenv-py3.9]
          285±2μs          286±1μs     1.00  item.ItemBench.time_item_from_dict [Prestons-MacBook-Pro.local/virtualenv-py3.9-orjson]
```

**Second Run:**
```
       before           after         ratio
     [f78637a9]       [6a1b1cf7]
     <main>           <issues/843-deprecation-warnings>
          283±2μs          283±2μs     1.00  item.ItemBench.time_item_from_dict [Prestons-MacBook-Pro.local/virtualenv-py3.9]
          281±2μs          282±2μs     1.00  item.ItemBench.time_item_from_dict [Prestons-MacBook-Pro.local/virtualenv-py3.9-orjson]
```

No substantial effect, it seems.

pjhartzell:
Some notes. @gadomski 
Bug =  Need to check if the deprecated field is `True`. Currently only checking for existence.
Perf = Only run deprecated checks if `version` extension exists in the `stac_extensions` list.
Refactor = Might it be cleaner (and more useful) to add an `is_deprecated()` method on the Asset, Collection, and Catalog classes? The deprecation checks in the `from_dict()` methods currently occur _after_ a class has been created, so we would have access to the proposed `is_deprecated()` method. We'd still have to loop, but the logic moved out.

gadomski (changes requested):
Looks good. It'd be nice to use the version extension instead of accessing `"deprecated"` directly -- if I recall correctly, we chatted offline about doing that first. Correct me if I'm wrong about that, @pjhartzell?

I think we should also add a section to the docs talking about deprecated objects and how to silence the warning.

pyproject.toml
```diff
@@ -1,2 +1,7 @@
 [tool.isort]
 profile = "black"
+
+[tool.pytest.ini_options]
+markers = [
+    "vcr: tests that use a vcrpy cassette"
+]
```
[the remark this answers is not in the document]
pjhartzell: I _was_ getting a `PytestUnknownMarkWarning`. But not anymore. I removed this.

benchmarks/item.py
```diff
@@ -15,7 +15,7 @@ def setup(self) -> None:
 
         self.stac_io = StacIO.default()
 
-        self.item_path = get_data_path("item/sample-item-asset-properties.json")
+        self.item_path = get_data_path("eo/eo-sentinel2-item.json")
```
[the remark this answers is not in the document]
pjhartzell: Done.

pjhartzell:
@gadomski We now have a pared-down implementation that uses the Version extension to check only Items and Collections for deprecation. Contains a new `get_version` method on the extension mixin class. 

Just want to raise a flag that adding `get_version` may be opening a can of worms. It causes one to start looking at how the other extension methods do not account for differing versions. 
- For example, if your item contains a different version than in `SCHEMA_URI`, you can have two versions of the same extension in `stac_extensions` via a call to the `add_to` method.
- The `remove_from` method only works for the hard-coded version in `SCHEMA_URI`
- `validate_has_extension` only checks for the version in `SCHEMA_URI`. 

We're opening the door to the reality that different extension versions exist and may be present on a STAC object, but there is some decent refactor required to handle that reality.

pystac/collection.py
```diff
@@ -678,6 +681,17 @@ def from_dict(
         if root:
             collection.set_root(root)
 
+        if (
+            CollectionVersionExtension.has_extension(collection)
+            and CollectionVersionExtension(collection).deprecated
+        ):
```
gadomski, line 687: Sorry to nitpick this, but I think the more "Pythonic" way would be something like:

```suggestion
        try:
            version = CollectionVersionExtension.ext(collection)
        except ExtensionNotImplemented:
             pass
        else:
            if version.deprecated:
                warnings.warn(
                    f"The collection '{collection.id}' is deprecated.",
                    DeprecatedWarning,
                )
```

pystac/extensions/base.py
```diff
@@ -132,13 +135,47 @@ def remove_from(cls, obj: S) -> None:
...
+            )
+            uri = next(
+                (uri for uri in obj.stac_extensions if uri.startswith(uri_prefix)), None
+            )
+            if uri is None:
+                return None
+            else:
+                return Path(safe_urlparse(uri).path).parts[-2]
+        else:
+            return None
```
gadomski, line 178: I think these are good changes, but w/ my suggested change to https://github.com/stac-utils/pystac/pull/1006/files#diff-28af67c49c05ff66bbfd9f3ab4487687c657a3c09d640b93ec9401b76ffb9665R687 we won't need them. Per your comment https://github.com/stac-utils/pystac/pull/1006#issuecomment-1452804629, maybe these should be part of a larger reimagining of the extension detection architecture?

pystac/item.py
```diff
@@ -483,6 +487,17 @@ def from_dict(
...
+        if (
+            ItemVersionExtension.has_extension(item)
+            and ItemVersionExtension(item).deprecated
+        ):
+            warnings.warn(
+                f"The item '{item.id}' is deprecated.",
+                DeprecatedWarning,
+            )
+            # Asset deprecation checks can be added once supported by the
+            # version extension.
```
gadomski, line 499: Same as for the collection, I think it would be more "Pythonic" to just `ext` and catch the exception.

tests/extensions/test_base.py
```diff
@@ -0,0 +1,117 @@
...
+    Path("local", "path", "custom_extension", "v1.0.0", "schema.json")
+)
+
+
+class TestVersionHttpSchemaUri:
+    def test_get_version(self, item: pystac.Item) -> None:
+        item.stac_extensions = [
+            "https://stac-extensions.github.io/projection/v2.5.0/schema.json"
+        ]
+        assert ProjectionExtension.get_version(item) == "v2.5.0"
```
gadomski, line 21:
```suggestion
        assert ProjectionExtension.get_version(item) == "2.5.0"
```

IMO the version should exclude the leading `v`, reference https://semver.org/#is-v123-a-semantic-version.

gadomski (changes requested):
TIL `recwarn`, I like that a lot! Sorry for being a 🧇 on the `has_extension` stuff -- if we don't keep the `get_version` changes in this PR, let's make sure to put it into another branch/draft PR so we don't lose it, it's useful stuff.

pystac/collection.py
pjhartzell: No need to apologize, your suggestion looks good. Implemented.

pystac/extensions/base.py
pjhartzell: I've removed this. Agreed, if nothing else, a good exercise in thinking about what would be required to handle different extension versions.

pystac/item.py
pjhartzell: Yep, done.

pjhartzell:
>I think we should also add a section to the docs talking about deprecated objects and how to silence the warning.

Agreed.

gadomski (approved):
Thanks for making the change to `try/except`, that reads much cleaner to me. I did a local benchmark run and no significant change. I also added another test section to see what happens when there's no extension url -- works fine.

PR #1016, opened 2023-02-28
When instantiating ItemCollections clone items by default

### LINKED ISSUE #859: ItemCollection constructor should clone items in iterator by default
Currently, if an ItemCollection is constructed with an iterator of Item objects, those objects may be mangled and become unusable by operation in ItemCollection. This can be avoided by setting the clone parameter to True. This change would make the clone parameter default to True to have correct behavior by default, and allow a client to set clone to False if they desired the higher performance of not cloning the Items.

jsignell:
**Related Issue(s):**

- closes #859 

**Description:**

NOTE this PR changes the behavior when adding item collections. After this PR adding item collections will no longer dedupe items. 

**PR Checklist:**

### DIFF
--- a/pystac/item_collection.py
+++ b/pystac/item_collection.py
@@ -48,9 +48,9 @@ class ItemCollection(Collection[pystac.Item]):
             :class:`~ItemCollection`.
         clone_items : Optional flag indicating whether :class:`~pystac.Item` instances
             should be cloned before storing in the :class:`~ItemCollection`. Setting to
-            ``True`` ensures that changes made to :class:`~pystac.Item` instances in
-            the :class:`~ItemCollection` will not mutate the original ``Item``, but
-            will result in slower instantiation. Defaults to ``False``.
+            ``False`` will result in faster instantiation, but changes made to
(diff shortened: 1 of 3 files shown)


pystac/item_collection.py
```diff
@@ -125,12 +125,8 @@ def __add__(self, other: object) -> "ItemCollection":
...
             return NotImplemented
 
-        combined = []
-        for item in self.items + other.items:
-            if item not in combined:
-                combined.append(item)
-
-        return ItemCollection(items=combined, clone_items=False)
+        combined = [*self.items, *other.items]
+        return ItemCollection(items=combined)
```
gadomski, line 129:
```suggestion
        return ItemCollection(items=combined, clone_items=True)
```

Since we're calling our function, IMO it's best to be explicit about the behavior we want.

gadomski (changes requested):
Needs a CHANGELOG entry.

pystac/item_collection.py
gadomski: Nope, I agree, `+` should mirror the default. Hopefully we don't have to change the default again but you never know. Thanks!

gadomski (approved):

PR #1027, opened 2023-03-06
Release v1.7.0

gadomski:
**Description:**

Includes a new `RELEASING.md` checklist and some CHANGELOG reorganization.

**PR Checklist:**


pystac/extensions/eo.py
```diff
@@ -545,7 +545,7 @@ def migrate(
                         obj["stac_extensions"] = []
                     if view.SCHEMA_URI not in obj["stac_extensions"]:
                         obj["stac_extensions"].append(view.SCHEMA_URI)
-                    if not "view:{}".format(field) in obj["properties"]:
```
philvarner, line 548: should this behavior change be in the release PR?

gadomski: Don't believe this is a behavior change, just a move of `not` to be `not in` (recommended by ruff)...I think that's just a lint?

philvarner: I didn't look at it enough to see it was the same, but why's that only showing up now?

gadomski: I ran `ruff . --fix` to see if it picked something up. I thought it was better to fold into an existing PR rather than make it Its Own Thing.

Maybe I should create a PR to add ruff instead?

philvarner: I think it should be it's own commit b/c this is the release pr.

If we're going to use ruff (which I am 👍🏻 on), it should be part of pre-commit and the ci build.

gadomski: It is it's own commit? cf6c17e222f1faf84fc0cfe7b1020f9413c6d9c5

> If we're going to use ruff (which I am 👍🏻 on), it should be part of pre-commit and the ci build.

https://github.com/stac-utils/pystac/issues/1030

philvarner: I meant PR. But whatever, :shipit:

philvarner (approved):

Issue #1040, opened 2023-03-14
Use black to format notebooks


jsignell:
I noticed that the notebook don't have black formatting enabled. I think it can be added to the pre-commit config.

jsignell:
I will open a PR shortly

PR #1034, opened 2023-03-08
Replace flake8 and isort with ruff

gadomski:
**Related Issue(s):**

- Closes #1030

**Description:**

[ruff](https://github.com/charliermarsh/ruff) is fast and has [good momentum](https://github.com/charliermarsh/ruff#whos-using-ruff). See https://beta.ruff.rs/docs/faq/#which-tools-does-ruff-replace for the list of tools that **ruff** can replace.

**PR Checklist:**

- [x] Code is formatted (run `pre-commit run --all-files`)
- [x] Tests pass (run `scripts/test`)
- [x] Documentation has been updated to reflect changes, if applicable
- [x] This PR maintains or improves overall codebase code coverage.
- [x] Changes are added to the [CHANGELOG](https://github.com/stac-utils/pystac/blob/main/CHANGELOG.md). See [the docs](https://pystac.readthedocs.io/en/latest/contributing.html#changelog) for information about adding to the changelog.

PR #1041, opened 2023-03-14
Use black to format notebooks

jsignell:
**Related Issue(s):**

- #1040 

**PR Checklist:**


.pre-commit-config.yaml
```diff
@@ -11,6 +11,14 @@ repos:
         stages: [commit]
         types: [python]
 
+      - id: black-jupyter
+        name: black-jupyter
+        entry: black
+        language: python
```
gadomski, line 17: I'm not quite sure of the lineage of `language: system` for the other pre-commit hooks, but we should _probably_ be consistent?

gadomski (commented):
I'm assuming the notebook changes are all formatting, no functional changes (read: I didn't look at them all)?

.pre-commit-config.yaml
jsignell: Yeah I tried it with `system` and it didn't like that plus the additional dependencies field. I have never seen `language: system` before, but I was hesitant about changing all the other pre-commit actions.

gadomski: Hmm...okay no worries. Per https://github.com/stac-utils/pystac/pull/1034#issuecomment-1468032168, I might be taking some time soon to write up some standards docs for this sort of stuff (e.g. **pystac-client** doesn't use `language: system`), which will be a chance to revisit how we set up the **pre-commit** hooks. So I think it's fine to do it the way that works now, and then we can bring everything in line later when we hit all the repos. Thanks.

jsignell:
> I'm assuming the notebook changes are all formatting, no functional changes (read: I didn't look at them all)?

Yeah it's almost all replacing `'` with `"`

gadomski (approved):

PR #1042, opened 2023-03-14
Update PR template since precommit runs on the diff by default

jsignell:
**Description:**

I don't think it is good practice to encourage people to run `pre-commit run --all-files`. It kind of undermines the beauty of pre-commit which is that it runs by default on the diff on each commit.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -6,7 +6,7 @@
 
 **PR Checklist:**
 
-- [ ] Code is formatted (run `pre-commit run --all-files`)
+- [ ] `pre-commit` hooks pass locally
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -6,7 +6,7 @@
 
 **PR Checklist:**
 
-- [ ] Code is formatted (run `pre-commit run --all-files`)
+- [ ] Code is formatted (`pre-commit` is installed)
```
gadomski, line 9:
```suggestion
- [ ] `pre-commit` hooks pass locally
```

The **pre-commit** hooks aren't just formatting, so maybe this text is more correct? What do you think?

gadomski (approved):

Issue #400, opened 2021-06-02
Runtime network loading of FIELDS_JSON_URL and mixed type enum


schwehr:
@volaya @duckontheweb @lossyrob and #264 

Introduced here [pystac/summaries.py#L40](https://github.com/stac-utils/pystac/blob/d2bebb349aaa6a3254118d4c93b763c1be57c6bc/pystac/summaries.py#L40), FIELDS_JSON_URL causes the default path to go load the field definitions from a remote file.   I am not a fan of having an additional code path (there was already the validation schema fetches) outside of testing:

- That file is not versioned with pystac
- It incurs a runtime cost of fetching 24089 bytes across a network.  For some folks this might be an issue.
- It's another code path that I have to patch out for hermetic testing
- It blocks using using this code when offline or sandboxed
- If something is wrong with that file, it's not obvious how to approach fixing it
- What if the file moves?

I would much rather have a script to rebuild a python file that contains just the resulting table that is only the parts used here.

Currently it's this:

```python
FIELDS_JSON_URL = ( 
   "https://cdn.jsdelivr.net/npm/@radiantearth/stac-fields/fields-normalized.json"
)

@lru_cache(maxsize=None)
def _get_fields_json(url: str) -> Dict[str, Any]:
    return pystac.StacIO.default().read_json(url)

# ...

class Summarizer:
    # ...
    def __init__(self, fields: Optional[str] = None):
        fieldspath = fields or FIELDS_JSON_URL
        try:
            jsonfields = _get_fields_json(fieldspath)
        except:
            if fields is None:
                raise Exception(
                    "Could not read fields definition file at "
                    f"{fields} or it is invalid.\n"
                    "Try using a local fields definition file."
                )
            else:
                raise
        self._set_field_definitions(jsonfields)

    def _set_field_definitions(self, fields: Dict[str, Any]) -> None:
        self.summaryfields: Dict[str, SummaryStrategy] = {}
        for name, desc in fields["metadata"].items():
            if isinstance(desc, dict):
                strategy_value = desc.get("summary", True)
                try:
                    strategy: SummaryStrategy = SummaryStrategy(strategy_value)
                except ValueError:
                    strategy = SummaryStrategy.DEFAULT
                if strategy != SummaryStrategy.DONT_SUMMARIZE:
                    self.summaryfields[name] = strategy
            else:
                self.summaryfields[name] = SummaryStrategy.DEFAULT
```

[...]

PR #1045, opened 2023-03-15
Load file

jsignell:
**Related Issue(s):**

- #400 

**Description:**

This PR adds a step to the release process that pulls the remote file and puts it in a particular location where it can be read from the pystac code and bundled up with the code when packaging. 

I did not try to tackle the question of cleaning up the Enum because of the option to pass in custom fields. I don't see an easy way to clean this up without breaking backwards compat. 

**PR Checklist:**


pystac/summaries.py
```diff
@@ -25,6 +26,12 @@
     from pystac.collection import Collection
     from pystac.item import Item
 
+here = Path(__file__).parent.resolve()
+
+# Every time pystac is released this file gets pulled from
+# https://cdn.jsdelivr.net/npm/@radiantearth/stac-fields/fields-normalized.json
+FIELDS_JSON_PATH = Path(here, "static/fields-normalized.json")
```
gadomski, line 33: I _think_ it's safer to use `importlib_resources`, e.g. https://github.com/stactools-packages/noaa-cdr/blob/7dc626225e78f701bc723e33b88f1e5f4b455d7b/src/stactools/noaa_cdr/ocean_heat_content/constants.py#L40-L44.

scripts/pull-static
gadomski, line 1: Can we just `curl` and reduce the new scripts from two to one?

scripts/pull-static
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -e
+
+echo "Pulling fields-normalized.json from cdn"
+python "$(dirname $0)/pull_static.py"
+
+echo "Committing changes"
+git add "$(dirname $(dirname $0))/pystac/static"
+git commit -m "Update fields-normalized.json"
```
gadomski, line 10: If there's no changes, this will fail, right? I _think_ we want it to succeed ... it's ok if the file doesn't need to be udpated.

gadomski (changes requested):

pystac/summaries.py
jsignell: OMGGGG I did not know about that. The main difference in that approach is that in the dev version you would get the latest verision aka the file would not live in git. But that seems like a fine compromise for a much more standardized approach.

jsignell: oh wait I misunderstood. I thought this was something that you could just put in the setupy.py and have it pulled into the package data. Ignore me.

scripts/pull-static
gadomski: Hm, yeah thanks for reminding me that I was a little unsure of that as well :-) ... I'd lean towards leaving the git commit out, and just making it part of the normal incantions you do when releasing (CHANGELOG update, script update, etc, then just commit it all in one or whatever).

pystac/summaries.py
```diff
@@ -118,18 +125,7 @@ class Summarizer:
     summaryfields: Dict[str, SummaryStrategy]
 
     def __init__(self, fields: Optional[str] = None):
-        fieldspath = fields or FIELDS_JSON_URL
-        try:
-            jsonfields = _get_fields_json(fieldspath)
-        except Exception as e:
-            if fields is None:
```
jsignell, line 125: This special error was only for the case where pulling the default remote file failed.

setup.py
```diff
@@ -23,7 +23,7 @@
     package_data={"": ["py.typed", "*.jinja2"]},
     py_modules=[splitext(basename(path))[0] for path in glob("pystac/*.py")],
     python_requires=">=3.8",
-    install_requires=["python-dateutil>=2.7.0"],
+    install_requires=["importlib-resources>=5.12.0", "python-dateutil>=2.7.0"],
```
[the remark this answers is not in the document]
jsignell: No worries. It looked to me like `.files` is only available after python 3.9: https://docs.python.org/3/library/importlib.resources.html#importlib.resources.files? I'll keep poking around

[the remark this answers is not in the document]
jsignell: Yeah I think importlib-resources is necessary for 3.8 support. But NEP29 is coming up for Python 3.8 in April so it shouldn't be in there for too long :)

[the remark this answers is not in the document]
gadomski: We can `python_version` gate the `install_requires` too right?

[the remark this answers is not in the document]
jsignell: yeah we _could_ that feels pretty strange to me though. I think I'd prefer to always include it and then remove it when that version of python isn't supported anymore.

scripts/pull-static
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -e
+
+DEST="$(dirname $(dirname $0))/pystac/static/fields-normalized.json"
+
+echo "Pulling fields-normalized.json from cdn to ${TARGET_DIR}"
+curl https://cdn.jsdelivr.net/npm/@radiantearth/stac-fields/fields-normalized.json | jq -c '{metadata: .metadata}' > ${DEST}
```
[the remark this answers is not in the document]
jsignell: I added a note about it in RELEASING

scripts/pull-static
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -e
+
+DEST="$(dirname $(dirname $0))/pystac/static/fields-normalized.json"
+
+echo "Pulling fields-normalized.json from cdn to ${TARGET_DIR}"
```
pjhartzell, line 7: Uninvited guest here. What's `TARGET_DIR`?

scripts/pull-static
```diff
@@ -0,0 +1,11 @@
...
+
+set -e
+
+VERSION="1.0.0-rc.1"
+SRC="https://cdn.jsdelivr.net/npm/@radiantearth/stac-fields@$VERSION/fields-normalized.json"
+HERE=$(dirname "$0")
+DEST=$(dirname "$HERE")/pystac/static/fields-normalized.json
+
+echo "Pulling fields-normalized.json from cdn to $DEST"
+curl "$SRC" | jq -c '{metadata: .metadata}' > "$DEST"
```
gadomski, line 11: Why are we only pulling the `.metadata` key here? I didn't see anything about pulling out the `.metadata` key in the original code (or maybe I missed it).

scripts/pull-static
gadomski, line 1: I took the liberty of updating the script to address Preston's comment (https://github.com/stac-utils/pystac/pull/1045#discussion_r1139547456) and use a versioned url, as requested in the original ticket (https://github.com/stac-utils/pystac/issues/400#issuecomment-859608341).

gadomski (changes requested):
I took the liberty of updating the CHANGELOG. One note -- when #400 is closed by this PR, we should create an issue to capture the second part of @schwehr's ticket (mixed type enum). I'll update the PR description to include a checkbox for this task.

scripts/pull-static
jsignell: too much refactoring :facepalm:

scripts/pull-static
jsignell: Only the metadata is used in the code so I figured why keep the rest around? I'm fine either way though.

jsignell:
>  we should create an issue to capture the second part of @schwehr's ticket (mixed type enum)

The reason for the mixed type enum is that there are mixed types in the original fields.json. So the enum is just smoothing over an oddity in the upstream definition. I don't really see how we can do anything about it in pystac itself as long as we still want to accept arbitrary fields json.

gadomski (approved):

pystac/summaries.py
```diff
@@ -84,21 +91,25 @@ def __repr__(self) -> str:
...
+        return jsonfields
     return pystac.StacIO.default().read_json(url)
 
 
 class SummaryStrategy(Enum):
     ARRAY = "v"
     RANGE = "r"
     SCHEMA = "s"
-    DONT_SUMMARIZE = False
+    DO_NOT_SUMMARIZE = False
```
gadomski, line 112: On second thought, can we pull this out of this PR? _Technically_ this is a breaking change so should go in a v2 milestone.

gadomski (changes requested):

pystac/summaries.py
jsignell: yes I can take this out of this PR

gadomski (commented):
LGTM ... I also re-added `FIELDS_JSON_URL` as a deprecated field, w/ a plan to remove in v2.

Because we're adding a new module, this is technically a feature-adding, backwards compatible change, so needs to be in v1.8. I'm going to hold off on merging until #1049 lands and we release v1.7.1.

gadomski (approved):

PR #1049, opened 2023-03-16
Fix __geo_interface__

### LINKED ISSUE #1048: Item __geo_interface__ returns the wrong represenation
The Item __geo_interface__ is currently:

```
    @property
    def __geo_interface__(self) -> Optional[Dict[str, Any]]:
        """Returns this item's geometry.

        https://gist.github.com/sgillies/2217756

        Returns:
            Dict[str, Any]: The Item's geometry
        """
        return deepcopy(self.geometry)
```

However, this should be returning a JSON-typed dict of a Feature for the Item, rather than the Item's geometry.

gadomski:
**Related Issue(s):**

- Closes #1048 

**Description:**

`__geo_interface__` should be the feature, not just the geometry. I misread the protocol on my first implementation. I made the choice to just call `to_dict` with all the dials set to no to make it as quick and light as possible. An alternative implementation would be to build a new dict with only the `__geo_interface__` required fields (`type`, `bbox`, `properties`, `geometry`, `coordinates`).

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Use [ruff](https://github.com/charliermarsh/ruff) instead of **isort** and **flake8** ([#1034](https://github.com/stac-utils/pystac/pull/1034))
 
+### Fixed
+
+- Item `__geo_interface__` now correctly returns a Feature, rather than only the Geometry ([#1049](https://github.com/stac-utils/pystac/pull/1049))
+
 ## [v1.7.0]
 
 ### Added
--- a/pystac/item.py
+++ b/pystac/item.py
@@ -531,12 +531,14 @@ class Item(STACObject):
         return identify_stac_object_type(d) == STACObjectType.ITEM
 
     @property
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 - Use [ruff](https://github.com/charliermarsh/ruff) instead of **isort** and **flake8** ([#1034](https://github.com/stac-utils/pystac/pull/1034))
 
+### Fixed
+
+- Item's `__geo_interface__` ([#1049](https://github.com/stac-utils/pystac/pull/1049))
```
philvarner, line 11:
```suggestion
- Item `__geo_interface__` now correctly returns a Feature, rather than only the Geometry ([#1049](https://github.com/stac-utils/pystac/pull/1049))
```

philvarner:
I think returning every field (instead of just the required GeoJSON Feature fields) is the correct behavior.

philvarner (approved):

PR #1074, opened 2023-03-31
Remove text around to_dict and JSON

### LINKED ISSUE #1047: Expectations of to_dict methods in returning JSON-typed values, regarding geometry
Currently, the to_dict method returns a "shallow" dict representation of the object. For example, if the Item has a geometry that has a `coordinates` value that is a nested tuple of numbers , the resulting dict's `geometry` key points to that same tuple. However, I think a more common assumption is that to_dict returns a JSON-typed representation of the Item, e.g., with geometry being nested arrays and numbers, regardless of what the internal representation is. Maybe there could be a flag on to_dict to determine this behavior, or another method like `to_geojson_dict`?

mypy checking doesn't catch this because geometry is typed as `Optional[Dict[str,Any]]`, so it can't catch it. (Union types would help here, allowing the stricter `Optional[Dict[str,str|list[list|float]]`)

The current docstring for to_dict is `Generate a dictionary representing the JSON of this serialized object.` , which is wrong wrt to the current behavior if non-JSON-compliant tuples are used. 

The context behind this is we had an Item that (unintentionally) had tuple geometry coordinate values, and used .to_dict() to create a value to pass to stac-validator validation, expecting that we were getting a JSON-style dict. The error message was:

```
Exception: STAC Item validation failed. Error: {‘type’: ‘Polygon’, ‘coordinates’: (((-91.00013888888888, 43.00013888888889), (-91.00013888888888, 44.00013888888889), (-92.00013888888888, 44.00013888888889), (-92.00013888888888, 43.00013888888889), (-91.00013888888888, 43.00013888888889)),)} is not valid under any of the given schemas. Error is in geometry .
```

because the tuple geometry doesn't match the schema of nested arrays of numbers, and then gets `str` serialized to a value of nested tuples of numbers. 

Relatedly, `__geo_interface__` will return an invalid value if tuple coordinates are used.

gadomski:
**Related Issue(s):**

- Closes #1047

**Description:**

Our `to_dict` methods do not always return objects that are strictly valid as JSON, e.g. they may contain tuples instead of lists. This PR updates all docstrings to remove references to returning valid JSON.

**PR Checklist:**

- [ ] (once approved) Create a new issue to capture the useful discussion in #1047 around (de)serialization libraries

### DIFF
--- a/pystac/asset.py
+++ b/pystac/asset.py
@@ -112,10 +112,10 @@ class Asset:
             return None
 
     def to_dict(self) -> Dict[str, Any]:
-        """Generate a dictionary representing the JSON of this Asset.
+        """Returns this Asset as a dictionary.
 
         Returns:
-            dict: A serialization of the Asset that can be written out as JSON.
+            dict: A serialization of the Asset.
         """
 
         d: Dict[str, Any] = {"href": self.href}
--- a/pystac/collection.py
+++ b/pystac/collection.py
@@ -86,10 +86,10 @@ class SpatialExtent:
         self.extra_fields = extra_fields or {}
 
     def to_dict(self) -> Dict[str, Any]:
-        """Generate a dictionary representing the JSON of this SpatialExtent.
+        """Returns this spatial extent as a dictionary.
 
         Returns:
-            dict: A serialization of the SpatialExtent that can be written out as JSON.
+            dict: A serialization of the SpatialExtent.
         """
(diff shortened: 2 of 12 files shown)


philvarner (approved):

gadomski:
Created https://github.com/stac-utils/pystac/issues/1092 to capture the serialization discussion.

Issue #541, opened 2021-07-08
RFC: Passing around mutable references


l0b0:
`Collection.get_assets` and `Item.get_assets` return `dict(self.assets.items())`. This does a shallow copy of the assets:

```python
>>> assets = {"a": []}
>>> bar = dict(assets.items())
>>> assets["a"].append("entry")
>>> bar
{'a': ['entry']}
```

I wonder if the intention was `deepcopy(self.assets)` or simply `self.assets` instead?

There are also a few other places where it does similar things, like `map(apply_asset_mapper, item.assets.items())`, which I'm even less sure about. But I would expect the intent to be, in general, either always or never return deep copies of anything which could possibly contain mutable references.

gadomski:
xref https://github.com/stactools-packages/cop-dem/issues/11#issuecomment-1472243905

PR #1087, opened 2023-04-07
Return deepcopy of `assets` from `get_assets`

jsignell:
**Related Issue(s):**

- #541

**Description:**

~Instead of making a shallow copy, just return the object itself.~ Always create a new dict and deepcopy all values

I grepped around for other places where `dict` or `.copy` are used, and this was all I found

**PR Checklist:**

### DIFF
--- a/pystac/collection.py
+++ b/pystac/collection.py
@@ -725,15 +725,12 @@ class Collection(Catalog):
             Dict[str, Asset]: A dictionary of assets that match ``media_type``
             and/or ``role`` if set or else all of this collection's assets.
         """
-        if media_type is None and role is None:
-            return dict(self.assets.items())
-        assets = dict()
-        for key, asset in self.assets.items():
-            if (media_type is None or asset.media_type == media_type) and (
-                role is None or asset.has_role(role)
-            ):
-                assets[key] = asset
-        return assets
+        return {
+            k: deepcopy(v)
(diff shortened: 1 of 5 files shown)


pystac/collection.py
```diff
@@ -726,7 +726,7 @@ def get_assets(
             and/or ``role`` if set or else all of this collection's assets.
         """
         if media_type is None and role is None:
-            return dict(self.assets.items())
+            return self.assets
```
gadomski, line 729: This feels a little inconsistent to me, since sometimes we get a reference to the actual object (if `media_type` and `role` are `None`) and sometimes we get a wholly new object. We should either return one or the other, and since we can't return a slice view of the interior object (from the filters) we probably have to deepcopy?

gadomski (changes requested):

pystac/collection.py
jsignell: Yeah that makes sense. I am pretty ambivalent about the usefulness of a method like `get_assets` anyways. It feels more natural to just access the assets directly via the attribute. I guess the point is to make it easier to filter?

gadomski: Yeah I think the filter's the main purpose -- I rarely if every use `get_assets` myself for exactly that reason, and that I don't know if I'm getting a mutable reference or a copy.

Issue #1084, opened 2023-04-06
Expand support for multiple extension uris


gadomski:
In https://github.com/stac-utils/pystac/pull/1081 we added support for older schema URIs to the `ProjectionExtension`. This behavior should be generalized so other extensions can use it easily. Ideas:

- Add `get_schema_uris()` to the base mixin
- Refactor the base `has_extension()` to check against `get_schema_uris()` instead of `get_schema_uri()`
- Backport any extensions that have older-but-compatible schema uris (thinking maybe `label` at least?)

This doesn't solve the problem, it's more of a band-aid -- if a newer version of an extension is incompatible with the Python structure, we don't have a mechanism to "migrate".

xref #448 which is our tracking issue for the general problem of extension versions.

PR #1088, opened 2023-04-07
Add MGRS Extension from code in stactools-package for sentinel2

jsignell:
**Related Issue(s):**

- #778

**Description:**

Code from: https://github.com/stactools-packages/sentinel2/blob/main/src/stactools/sentinel2/mgrs.py
Example from: https://github.com/stac-extensions/mgrs/blob/main/examples/item.json

Added pytest-style tests, docs

**PR Checklist:**

- [x] `pre-commit` hooks pass locally
- [x] Tests pass (run `scripts/test`)
- [x] Documentation has been updated to reflect changes, if applicable
- [x] This PR maintains or improves overall codebase code coverage.
- [x] Changes are added to the [CHANGELOG](https://github.com/stac-utils/pystac/blob/main/CHANGELOG.md). See [the docs](https://pystac.readthedocs.io/en/latest/contributing.html#changelog) for information about adding to the changelog.

Issue #993, opened 2023-02-15
Where reasonable, convert tests to free functions


gadomski:
We switched to **pytest** in #939, which means we don't have to use `unittest.TestCase` anymore; we can switch to free functions. In most cases, converting tests should be as simple as changing:

```python
class MyTestCase(unittest.TestCase):
    def test_foo(self) -> None:
        self.assertTrue(True)
```

to

```python
def test_foo() -> None:
    assert True
```

While this is a little bike-sheddy, this does have some benefits:
- Enable [fixtures, parameterization, and custom hooks](https://docs.pytest.org/en/7.1.x/how-to/unittest.html#pytest-features-in-unittest-testcase-subclasses)
- Reduces confusion for new contributors, who might be unsure of which test format to use
- Reduces the indentation level (a little thing, I know, but IMO it reads cleaner to have less leading whitespace)

teks:
This seems straightforward so I'll make an attempt. 28 files have `import unittest` so I may split it into a few PRs so they don't get too long.

However, please let me know if there's a body of tests I should leave alone until some other work is done.

gadomski:
Nope, go to town 😄

PR #1091, opened 2023-04-10
 Expand support for multiple extension uris

jsignell:
**Related Issue(s):**

- #1084

**Description:**

From issue: 
> In https://github.com/stac-utils/pystac/pull/1081 we added support for older schema URIs to the `ProjectionExtension`. This behavior should be generalized so other extensions can use it easily. Ideas:
> 
> - Add `get_schema_uris()` to the base mixin
> - Refactor the base `has_extension()` to check against `get_schema_uris()` instead of `get_schema_uri()`
> - Backport any extensions that have older-but-compatible schema uris (thinking maybe `label` at least?)
> 
> This doesn't solve the problem, it's more of a band-aid -- if a newer version of an extension is incompatible with the Python structure, we don't have a mechanism to "migrate".
> 
> xref #448 which is our tracking issue for the general problem of extension versions.

I looked over the merged PRs from the last year or so to try to get a sense of which extensions had gotten updates recently. 

I also added the previous versions to the `prev_extension_ids` in the `ExtensionHooks`. This means that when the item is migrated, the extension schema uri gets updated as well. Note: It looks like there _is_ a mechanism for migrating per-extension. You can see it in `LabelExtensionHooks.migrate`

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 - Include a copy of the `fields.json` file (for summaries) with each distribution of PySTAC ([#1045](https://github.com/stac-utils/pystac/pull/1045))
 - Removed documentation references to `to_dict` methods returning JSON ([#1074](https://github.com/stac-utils/pystac/pull/1074))
+- Expand support for previous extension schema URIs ([#1091](https://github.com/stac-utils/pystac/pull/1091))
 
 ### Deprecated
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,6 @@
 # TODO move all test case code to this file
 
+from pathlib import Path
 from datetime import datetime
 
 import pytest
@@ -7,6 +8,9 @@ import pytest
 from pystac import Catalog, Collection, Item
 
 from .utils import ARBITRARY_BBOX, ARBITRARY_EXTENT, ARBITRARY_GEOM, TestCases
+
+
+here = Path(__file__).resolve().parent
 
 
 @pytest.fixture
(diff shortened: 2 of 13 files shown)


tests/conftest.py
```diff
@@ -38,3 +42,7 @@ def test_case_8_collection() -> Collection:
 def projection_landsat8_item() -> Item:
     path = TestCases.get_path("data-files/projection/example-landsat8.json")
     return Item.from_file(path)
+
+
+def get_data_file(rel_path: str) -> str:
+    return str(here / "data-files" / rel_path)
```
jsignell, line 48: This is copied from #1088 I will pull any requested changes on that side into this PR.

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@
 ### Changed
 
 - Include a copy of the `fields.json` file (for summaries) with each distribution of PySTAC ([#1045](https://github.com/stac-utils/pystac/pull/1045))
+-
+Expand support for previous extension schema URIs ([#1091](https://github.com/stac-utils/pystac/pull/1091))
```
gadomski, line 15:
```suggestion
- Expand support for previous extension schema URIs ([#1091](https://github.com/stac-utils/pystac/pull/1091))
```

tests/conftest.py
gadomski: We should probably migrate any `TestCases.get_path` to `get_data_file` as well, for consistency's sake. Could happen in this or another PR.

gadomski (changes requested):
LGTM, just a couple light touchups.

tests/conftest.py
jsignell: I think a follow up PR. The intention is to move to fully move to pytest right?

gadomski: Yup (https://github.com/stac-utils/pystac/issues/993, https://github.com/stac-utils/pystac/issues/948), makes sense.

gadomski (approved):

PR #1103, opened 2023-04-24
Remove duplicate self links on items

### LINKED ISSUE #1102: Duplicate `self` link in pystac 1.7.2
```python
import pystac

item = pystac.read_file("https://planetarycomputer.microsoft.com/api/stac/v1/collections/sentinel-2-l2a/items/S2A_MSIL2A_20150704T101006_R022_T35XQA_20210411T133707")
links = item.get_links(rel="self")
print(len(links))

for link in links:
    print(link)
```

That will output 

```
2
<Link rel=self target=https://planetarycomputer.microsoft.com/api/stac/v1/collections/sentinel-2-l2a/items/S2A_MSIL2A_20150704T101006_R022_T35XQA_20210411T133707>
<Link rel=self target=https://planetarycomputer.microsoft.com/api/stac/v1/collections/sentinel-2-l2a/items/S2A_MSIL2A_20150704T101006_R022_T35XQA_20210411T133707>
```

for pystac>=1.7.0. On 1.6.1 it outputs just the single `self` link. None of the other links are duplicated, just `self`.

gadomski:
**Related Issue(s):**

- Fixes #1102

**Description:**

The `self` link is set in the constructor, so `from_dict` doesn't need to add the link again. This PR is against the v1.7 branch so we can use it to release a bugfix v1.7.3.

**PR Checklist:**

### DIFF
--- a/pystac/item.py
+++ b/pystac/item.py
@@ -6,7 +6,7 @@ from html import escape
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Type, TypeVar, Union, cast
 
 import pystac
-from pystac import STACError, STACObjectType
+from pystac import RelType, STACError, STACObjectType
 from pystac.asset import Asset
 from pystac.catalog import Catalog
 from pystac.collection import Collection
@@ -476,13 +476,9 @@ class Item(STACObject):
             assets={k: Asset.from_dict(v) for k, v in assets.items()},
         )
 
-        has_self_link = False
         for link in links:
-            has_self_link |= link["rel"] == pystac.RelType.SELF
-            item.add_link(Link.from_dict(link))
-
-        if not has_self_link and href is not None:
-            item.add_link(Link.self_href(href))
+            if href is None or link.get("rel", None) != RelType.SELF:
+                item.add_link(Link.from_dict(link))
 
         if root:
             item.set_root(root)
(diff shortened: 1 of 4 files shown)


pystac/item.py
```diff
@@ -476,14 +476,9 @@ def from_dict(
...
         )
 
-        has_self_link = False
         for link in links:
-            has_self_link |= link["rel"] == pystac.RelType.SELF
             item.add_link(Link.from_dict(link))
 
-        if not has_self_link and href is not None:
-            item.add_link(Link.self_href(href))
-
```
pjhartzell, line 486: I don't disagree with your changes here in the sense that it remove duplicate logic contained in the constructor (use the `href` to create a `self` link). But this now deleted logic did not add a duplicate `self` link. If you test this revised code with the same example given in the issue, you still get two `self` links. 

I think the problem exists when you create an Item from an href where the Item contains a `self` link. The href is being passed to the constructor, which adds a new `self` link from that href. After using the constructor (which added a new `self` link), the `from_dict` method then adds all the existing links, which also includes the `self` link. So you end up with two `self` links. There needs to be a check for an existing self link in this for loop. But not sure what to do if you end up with two different values.

pjhartzell (changes requested):

pystac/item.py
gadomski: Good catch, thanks. I elected to prefer the `href`-generated `self` link, rather than the one that came from the original item.

pjhartzell (approved):

PR #1093, opened 2023-04-11
Add classification extension

jpolchlo:
**Related Issue(s):**
Closes #788

**Description:**
This PR adds the [Classification Extension](https://github.com/stac-extensions/classification) to pystac.

**PR Checklist:**

- [x] `pre-commit` hooks pass locally
- [x] Tests pass (run `scripts/test`)
- [x] Documentation has been updated to reflect changes, if applicable
- [x] This PR maintains or improves overall codebase code coverage.
- [x] Changes are added to the [CHANGELOG](https://github.com/stac-utils/pystac/blob/main/CHANGELOG.md). See [the docs](https://pystac.readthedocs.io/en/latest/contributing.html#changelog) for information about adding to the changelog.

PR #1110, opened 2023-05-01
Support item_assets extension's AssetDefinition in raster extension

jpolchlo:
**Related Issue(s):**

- Connects #1093 

**Description:**
There is a missing compliance in the raster extension vis-a-vis the item asset extension.  It is permissible by the [raster extension spec](https://github.com/stac-extensions/raster/blob/main/json-schema/schema.json#L51-L55) to tuck `raster_bands` information into the `item_assets` block on a collection.  Raster appears not to support this.  In the interest of creating a relatively uniform strategy to query information in collection assets, we may be interested in elevating the item_assets' `AssetDefinition` to first class status with respect to the raster extension.  This PR proposes a change to make that happen.

This connects to the classification extension work because that extension can use the same mechanism.

Notes:
- `pystac.extensions.item_assets.AssetDefinition` is defined nearly identically to `pystac.Asset`, with the difference that the former disallows the `href` field, while the latter requires it.  If we could find a smart way to define `AssetDefinition` as a subclass of `Asset`, we could possibly entirely eliminate the need for this PR.
- `AssetDefinition` has no notion of an owner, so it's not currently possible to validate that the collection hosting the asset definition has the raster extension registered with it.  We may want to add an owner to this class to improve its expressivity.  Again, this would be obviated by proposal in the above comment.

**PR Checklist:**


jpolchlo:
After in-person discussion, I'll be adding an owner field to the asset definition.  Regularizing `AssetDefinition` with `Asset` didn't seem to pass the sniff test with @gadomski.  Womp womp.

gadomski:
> Regularizing AssetDefinition with Asset didn't seem to pass the sniff test with @gadomski. Womp womp.

I'm probably more of a -0 than -1 on this idea ... I could be on board with trying to share behavior between the two, but I worry about over-abstracting what is really a pretty simple data structure.

jpolchlo:
> I worry about over-abstracting what is really a pretty simple data structure.

Yeah, that tracks.  It's easy to enrich asset definitions, so it's not a huge problem.  Many extension specs are written with `item:assets` in there, but many are not.   Treating `AssetDefinition` too much like `Asset` opens the door to too much "accidental" support for item assets in extensions that didn't mean to have it.  Plus the `href` field is pretty fundamental to Asset-ness, and to not have it is inviting bugs.  Let's just leave it out.

pystac/extensions/raster.py
```diff
@@ -741,6 +746,42 @@ def summaries(
...
+        self.properties = asset.extra_fields
+        if asset.owner and isinstance(asset.owner, pystac.Item):
+            self.additional_read_properties = [asset.owner.properties]
+
+    def __repr__(self) -> str:
+        return "<AssetRasterExtension Asset href={}>".format(self.asset_href)
+
+
+class ItemAssetsRasterExtension(RasterExtension[item_assets.AssetDefinition]):
+    asset_defn: item_assets.AssetDefinition
```
gadomski, line 771: Nit, letters are cheap and abbreviations hide meaning:

```suggestion
    asset_definition: item_assets.AssetDefinition
```

gadomski (approved):

gadomski (changes requested):
Needs CHANGELOG entry, otherwise LGTM.

jpolchlo:
This is ready for merge when tests pass.

gadomski (approved):

Issue #1116, opened 2023-05-09
Arbitrary saving location of child catalogs


constantinius:
It is hard to control the actual storage location of child catalogs when they are child of more than one catalog. Consider the following example:

```python
import pystac

root = pystac.Catalog("root", "root")
left = pystac.Catalog("left", "left")
right = pystac.Catalog("right", "right")

root.add_child(left)
root.add_child(right)

child = pystac.Catalog("child", "child")

left.add_child(child)
right.add_child(child)

root.normalize_and_save("out_test", pystac.CatalogType.ABSOLUTE_PUBLISHED)
```

Whether the "child" catalog is saved under `left/child` or `right/child` depends on the order "left" and "right" are added to "root" (`right/child` in this case, as "right" was added after "left").

---

Additionally, when I change the order of the catalog addition:
```python
root.add_child(right)
root.add_child(left)

# ...

left.add_child(child)
right.add_child(child)
```

Then there is no parent link from "child" to "right" at all, even when this should be the case as the `add_child` on "right" was called last:

```json
{
  "type": "Catalog",
  "id": "child",
  "stac_version": "1.0.0",
  "description": "child",
  "links": [
    {
      "rel": "root",
      "href": "/out_test/catalog.json",
      "type": "application/json"
    },
    {
      "rel": "parent",
      "href": "/out_test/left/catalog.json",
      "type": "application/json"
    },
    {
      "rel": "self",
      "href": "/out_test/left/child/catalog.json",
      "type": "application/json"
    }
  ],
  "stac_extensions": []
}
```

m-mohr:
Related: https://github.com/radiantearth/stac-spec/issues/1230

m-mohr:
PR is at #1117

m-mohr:
Sorry, I'm reopening here because #1117 actually doesn't fully solve the issue. It helps with the parent links, but the storage location is still showing unexpected results.

m-mohr:
I've looked into the code a bit and it looks difficult ;-) Some findings:
- resolve_stac_object overrides the parent that we have carefully set before (from left to right)
- In normalize_href it sets the self links for a single catalog first to left and then to right, which decides the storage location. It doesn't take the parent into account, which I think it should.

gadomski:
```python
left.add_child(child)
right.add_child(child, keep_parent=True)
```

[...]

PR #1118, opened 2023-05-11
Don't override parent link when determining hrefs #1116

m-mohr:
**Related Issue(s):**

- #1116

**Description:**

Don't override parent link when determining hrefs.
Also don't set a self href when the parent set as link is not the current parent iterating over.

**PR Checklist:**

### DIFF
--- a/pystac/catalog.py
+++ b/pystac/catalog.py
@@ -718,9 +718,16 @@ class Catalog(STACObject):
         if not is_absolute_href(root_href):
             root_href = make_absolute_href(root_href, os.getcwd(), start_is_dir=True)
 
-        def process_item(item: Item, _root_href: str) -> Callable[[], None]:
+        def process_item(
+            item: Item, _root_href: str, parent: Optional[Catalog]
+        ) -> Optional[Callable[[], None]]:
             if not skip_unresolved:
                 item.resolve_links()
+
+            # Abort as the intended parent is not the actual parent
+            # https://github.com/stac-utils/pystac/issues/1116
+            if parent is not None and item.get_parent() != parent:
+                return None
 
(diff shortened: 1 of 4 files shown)


m-mohr:
@gadomski I don't think I can fix this, it's too deep in the internals and my changes seem to have unintended side effects. But maybe this PR helps you to nail down the issue(s) more easily?

pystac/link.py
```diff
@@ -334,7 +334,8 @@ def resolve_stac_object(self, root: Optional[Catalog] = None) -> Link:
             and isinstance(self.owner, pystac.Catalog)
         ):
             assert self._target_object
-            self._target_object.set_parent(self.owner)
+            if self._target_object.get_parent() is None:
```
m-mohr, line 337: PySTAC overrides a parent that we had intentionally set, which is not ideal. But it seems some cases need this...

pystac/catalog.py
```diff
@@ -730,13 +737,21 @@ def fn() -> None:
...
+            parent: Optional[Catalog] = None,
         ) -> List[Callable[[], None]]:
             setter_funcs: List[Callable[[], None]] = []
 
             if not skip_unresolved:
                 cat.resolve_links()
 
+            # Abort as the intended parent is not the actual parent
+            # https://github.com/stac-utils/pystac/issues/1116
+            if parent is not None and cat.get_parent() != parent:
```
m-mohr, line 752: What I'm trying here is to figure out whether the parent link and the actual parent in the catalog are not the same and then skip setting the links to avoid that the wrong self link doesn't override the correct self link.

jsignell:
Superseded by #1155

PR #1120, opened 2023-05-12
return all validation issues

### LINKED ISSUE #887: RFC: Allow `validate_all` to return all issues with a STAC catalog
Right now the behavior of `validate_all` and `Catalog.validate_all` is to return only the first problem it encounters in a STAC catalog. 

It would be nice if there was some flag we could set (or alternate method we could use) so that we could get a list of all the issues that were discovered.

`jsonschema` provides a way to do this via [`iter_errors`](https://python-jsonschema.readthedocs.io/en/latest/api/jsonschema/protocols/?highlight=iter_errors#jsonschema.protocols.Validator.iter_errors) and I'm wondering if something like that could work here. 

Also not sure if there is some way to do this already as I'm quite new to `pystac`. Happy to take this on as a first issue if the team thinks it's appropriate.

ircwaves:
**Related Issue(s):**

- #887

**Description:**

Closes #887 by returning all issues found during validation in the source parameter of the `STACValidationError`

**PR Checklist:**

- [X] `pre-commit` hooks pass locally

### DIFF
--- a/pystac/validation/stac_validator.py
+++ b/pystac/validation/stac_validator.py
@@ -159,9 +159,36 @@ class JsonSchemaSTACValidator(STACValidator):
 
         return schema, resolver
 
-    def _validate_from_uri(self, stac_dict: Dict[str, Any], schema_uri: str) -> None:
-        schema, resolver = self.get_schema_from_uri(schema_uri)
-        jsonschema.validate(instance=stac_dict, schema=schema, resolver=resolver)
+    def _validate_from_uri(
+        self,
+        stac_dict: Dict[str, Any],
+        stac_object_type: STACObjectType,
+        schema_uri: str,
+        extension_id: Optional[str] = None,
+        href: Optional[str] = None,
+    ) -> None:
+        try:
+            schema, resolver = self.get_schema_from_uri(schema_uri)
+            # This block is cribbed (w/ change in error handling) from
+            # jsonschema.validate
+            cls = jsonschema.validators.validator_for(schema)
+            cls.check_schema(schema)
+            validator = cls(schema, resolver=resolver)
+            errors = list(validator.iter_errors(stac_dict))
+        except Exception as e:
+            logger.error(f"Exception while validating {stac_object_type} href: {href}")
+            logger.exception(e)
+            raise
+        if errors:
+            msg = self._get_error_message(
+                schema_uri,
+                stac_object_type,
+                extension_id,
+                href,
+                stac_dict.get("id", None),
+            )
+            best = jsonschema.exceptions.best_match(errors)
+            raise pystac.STACValidationError(msg, source=errors) from best
         for uri in resolver.store:
             if uri not in self.schema_cache:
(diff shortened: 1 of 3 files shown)


pystac/validation/stac_validator.py
```diff
@@ -159,9 +159,34 @@ def get_schema_from_uri(self, schema_uri: str) -> Tuple[Dict[str, Any], Any]:
...
+    def _validate_from_uri(
+        self,
+        stac_dict: Dict[str, Any],
+        stac_object_type: STACObjectType,
+        schema_uri: str,
+        extension_id: Optional[str] = None,
+        href: Optional[str] = None,
+    ) -> None:
+        try:
+            schema, resolver = self.get_schema_from_uri(schema_uri)
```
gadomski, line 171: Can you add a comment explaining where this code comes from? (cribbed from the **jsonschema** source)

pystac/validation/stac_validator.py
```diff
@@ -159,9 +159,34 @@ def get_schema_from_uri(self, schema_uri: str) -> Tuple[Dict[str, Any], Any]:
...
+        if errors:
+            msg = self._get_error_message(
+                schema_uri,
+                stac_object_type,
+                extension_id,
+                href,
+                stac_dict.get("id", None),
+            )
+            best = jsonschema.exceptions.best_match(errors)
+            raise pystac.STACValidationError(msg, source=errors) from best
```
gadomski, line 189: I like this -- keeps the original behavior, but augments the exception w/ the full suite of errors. 💡

pystac/validation/stac_validator.py
```diff
@@ -246,22 +269,18 @@ def validate_extension(
         Returns:
            str: URI for the JSON schema that was validated against, or None if
                no validation occurred.
+
+        Raises:
+           STACValidationError if stac_dict is not valid.
```
gadomski, line 274:
```suggestion
           STACValidationError if stac_dict is not valid. The exception is raised from the "best" error, as determined by the jsonschema library. To access all jsonschema validation errors, use ``STACValidationError.source``.
```

Formatting isn't quite right here, and the language could use some work, but I think we need to tell people about this new capability.

gadomski (changes requested):

gadomski:
Also, needs a CHANGELOG entry.

pystac/validation/stac_validator.py
ircwaves: 🐢 small improvements

pystac/validation/stac_validator.py
ircwaves: I'm no Tom Waits, but that wording sounds pretty good to me.

pystac/validation/stac_validator.py
ircwaves: attrib'ed

ircwaves:
> Also, needs a CHANGELOG entry.

Yup. Intentional delay of description. 10 ⭐ penalty.  Described now.

gadomski (approved):

PR #1125, opened 2023-05-16
Always write `/`-delimited paths

### LINKED ISSUE #215: Don't use _pathlib in utils.py

### LINKED ISSUE #619: Problem with relative, unix-like hrefs on Windows
Hi everyone!
I'm not sure what exactly I'm doing wrong or if this is a possible bug. So any help or advice is appreciated!
I have a bunch of STAC items with relative, unix-like asset hrefs. From these items I simply want to create a self-contained   
catalog. That's about it. I've reduced my code to a simple example using only one collection (of an overall larger, tile-based catalog structure) and one item:

```python
import pystac
import os

data_dir = 'C:\\my\\data\\dir'
tile = 'tile_id'
scene = 'scene_name'

sp_extent = pystac.SpatialExtent([None, None, None, None])
tmp_extent = pystac.TemporalExtent([None, None])

test_collection = pystac.Collection(id=tile, description='test collection', 
                                    extent=pystac.Extent(sp_extent, tmp_extent))

item_path = os.path.join(data_dir, tile, scene, scene + '.json')
item = pystac.read_file(href=item_path)

test_collection.add_item(item=item)
extent = test_collection.extent.from_items(items=[item])
test_collection.extent = extent
```
Now, when I normalize all hrefs and save my collection, as shown [here](https://pystac.readthedocs.io/en/1.0/concepts.html#normalizing-hrefs) in the docs, I kind of get what I want. All hrefs, including those of item assets, are relative 👍🏻 but they use windows-style double backslashes 👎🏻. This causes problems with STAC Browser for example.

```python
test_collection.normalize_hrefs(root_href=os.path.join(data_dir, tile)
test_collection.save(catalog_type=pystac.CatalogType.SELF_CONTAINED)
```
case_A

If I replace the backward slashes from the root_href to normalize against, it results in a collection with relative, unix-style hrefs 👍🏻 but, for whatever reason, also in absolute hrefs for all item assets 👎🏻 

```python
test_collection.normalize_hrefs(root_href=os.path.join(data_dir, tile).replace('\\', '/'))
test_collection.save(catalog_type=pystac.CatalogType.SELF_CONTAINED)
```

image10

I've also played around with the `make_all_asset_hrefs_relative` method, but it didn't help unfortunately.

pjhartzell:
**Related Issue(s):**

- Closes #619 
- Closes #215 

**Description:**

- Link and asset HREFs are now always written with POSIX-style (`/`) delimiters whether on a POSIX or Windows system. This better aligns with the STAC specification (see [Matthias's comment](https://github.com/stac-utils/pystac/issues/619#issuecomment-1017926035)). 
- Addresses #215 by removing `_pathlib = os.path` from `utils.py`. Monkey patching `os.path` to behave as if on a Windows machine for testing is no longer needed since the full suite of tests now runs on Windows in CI.

Notes:
- Even though it is no longer used, I left the `JoinType` class and the associated `join_path_or_url` function in `utils.py` since users may be relying on them for other purposes. However, I also marked them as deprecated for eventual removal.
- I added a small utility function `make_posix_style` that converts windows path `\\` and `\` delimiters to `/` delimiters using string manipulation. I endeavored to use it sparingly.
    - Note that when on a Windows machine, both `ntpath.abspath` and `posixpath.abspath` will return a path containing `\\` delimiters. All calls to `os.path.abspath` are therefore wrapped in `make_posix_style`.
- `posixpath.relpath` doesn't play well with Windows drive letters, so we continue to use `os.path.relpath` (which will use `ntpath.relpath` on Windows machines) and accept the fact that existing STAC containing Windows paths should only be manipulated on Windows machines.

**Performance:**

- **POSIX:** Running `./scripts/bench` is flaky on WSL on a Windows machine. Sometimes I get `PERFORMANCE INCREASED` for certain benchmarks, other times decreased. On a Mac, the result is consistently `BENCHMARKS NOT SIGNIFICANTLY CHANGED`.
- **Windows:** Looks good, benchmark script command reports `BENCHMARKS NOT SIGNIFICANTLY CHANGED`.

**PR Checklist:**

### DIFF
--- a/pystac/utils.py
+++ b/pystac/utils.py
@@ -1,8 +1,19 @@
 import os
 import posixpath
+import warnings
 from datetime import datetime, timezone
 from enum import Enum
-from typing import Any, Callable, Dict, List, Optional, TypeVar, Union, cast
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    List,
+    Optional,
+    TypeVar,
+    Union,
+    cast,
+)
 from urllib.parse import ParseResult as URLParseResult
 from urllib.parse import urljoin, urlparse, urlunparse
 
@@ -10,9 +21,26 @@ import dateutil.parser
 
 from pystac.errors import RequiredPropertyMissing
 
-# Allow for modifying the path library for testability
-# (i.e. testing Windows path manipulation on non-Windows systems)
-_pathlib = os.path
+if TYPE_CHECKING:
+    PathLike = os.PathLike[str]
+else:
+    PathLike = os.PathLike
+
+HREF = Union[str, os.PathLike]
+
+
+def make_posix_style(href: HREF) -> str:
+    """Converts double back slashes and single back slashes to single forward
+    slashes for converting Windows paths to Posix style.
+
+    Args:
+        href (Union[str, os.PathLike]) : The href string or path-like object.
+
+    Returns:
+        str : The converted href in string form.
+    """
+    _href = str(os.fspath(href))
+    return _href.replace("\\\\", "/").replace("\\", "/")
 
 
 def safe_urlparse(href: str) -> URLParseResult:
@@ -30,7 +58,13 @@ def safe_urlparse(href: str) -> URLParseResult:
         urllib.parse.ParseResult : The named tuple representing the parsed HREF.
     """
     parsed = urlparse(href)
-    if parsed.scheme != "" and href.lower().startswith("{}:\\".format(parsed.scheme)):
+    if parsed.scheme != "" and (
+        href.lower().startswith("{}:\\".format(parsed.scheme))
+        or (
+            href.lower().startswith("{}:/".format(parsed.scheme))
+            and not href.lower().startswith("{}://".format(parsed.scheme))
+        )
+    ):
         return URLParseResult(
             scheme="",
             netloc="",
@@ -57,11 +91,22 @@ class StringEnum(str, Enum):
 
 
 class JoinType(StringEnum):
-    """Allowed join types for :func:`~pystac.utils.join_path_or_url`."""
+    """DEPRECATED.
+
+    .. deprecated:: 1.8.0
+        No longer used internally by pystac.
+
+    Allowed join types for :func:`~pystac.utils.join_path_or_url`.
+    """
 
     @staticmethod
     def from_parsed_uri(parsed_uri: URLParseResult) -> "JoinType":
-        """Determines the appropriate join type based on the scheme of the parsed
+        """DEPRECATED.
+
+        .. deprecated:: 1.8.0
+            No longer used internally by pystac.
+
+        Determines the appropriate join type based on the scheme of the parsed
         result.
 
         Args:
@@ -71,6 +116,13 @@ class JoinType(StringEnum):
         Returns:
             JoinType : The join type for the URI.
         """
+        warnings.warn(
+            message=(
+                "from_parsed_uri is deprecated and will be removed in pystac "
+                "version 2.0.0. It is no longer used internally by pystac."
+            ),
+            category=DeprecationWarning,
+        )
         if parsed_uri.scheme == "":
             return JoinType.PATH
         else:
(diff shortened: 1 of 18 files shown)


pystac/utils.py
```diff
@@ -30,7 +46,14 @@ def safe_urlparse(href: str) -> URLParseResult:
         urllib.parse.ParseResult : The named tuple representing the parsed HREF.
     """
     parsed = urlparse(href)
-    if parsed.scheme != "" and href.lower().startswith("{}:\\".format(parsed.scheme)):
+    # https://pypi.org/project/urllib3/#history
```
gadomski, line 49: What's this comment all about?

tests/posix_paths/test_posix_paths.py
```diff
@@ -0,0 +1,205 @@
...
+from tests.utils import ARBITRARY_BBOX, ARBITRARY_EXTENT, ARBITRARY_GEOM
+
+
+def check_link(link: Optional[pystac.Link]) -> None:
+    assert link is not None
+    href = link.get_target_str()
+    assert href is not None
+    assert "\\" not in href
+    print()
+    print(href)
```
gadomski, line 20: Shouldn't have leftover `print`s in tests.

tests/posix_paths/test_posix_paths.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+def test_create_item_from_absolute_posix_href() -> None:
+    # Should pass on both Windows and POSIX machines
```
gadomski, line 24: Not a useful comment, as we expect all tests to pass unless they have a platform-specific skip, right?

gadomski (changes requested):
Overall: awesome 🎉 

> Even though it is no longer used, I left the JoinType class and the associated join_path_or_url function in utils.py. Users may be relying on them for other purposes.

Worth deprecating these since they're not used in the library?

> Windows: I'm unable to run the benchmark suite. I'm getting errors from a git clone operation, where it is unable to create some of the test data files because the file paths are too long.

Can you open an issue to track this?

pystac/utils.py
pjhartzell: Huh. No idea. Deleting it.

tests/posix_paths/test_posix_paths.py
pjhartzell: Ugh. Such slop. Nuked. Thanks for catching that.

pjhartzell:
>Worth deprecating these since they're not used in the library?

Done.

> Can you open an issue to track this?

I think I have a workaround for this: `git config --system core.longpaths true` allows for longer paths on a Windows machine.

pystac/utils.py
```diff
@@ -10,11 +11,27 @@
 
 from pystac.errors import RequiredPropertyMissing
 
+HREF = Union[str, os.PathLike]
+
 # Allow for modifying the path library for testability
 # (i.e. testing Windows path manipulation on non-Windows systems)
 _pathlib = os.path
```
gadomski, line 18: Since we're working with windows stuff in this PR, can we remove this alias and close https://github.com/stac-utils/pystac/issues/215 as well?

pystac/utils.py
```diff
@@ -10,11 +11,27 @@
 
 from pystac.errors import RequiredPropertyMissing
 
+HREF = Union[str, os.PathLike]
```
gadomski, line 14:
```suggestion
Href = Union[str, os.PathLike]
```

Since it behaves like a type/class, would `Href` be a more appropriate name?

pystac/utils.py
pjhartzell: We can. I wasn't keen on wrapping another issue into this PR, but it's feeling pretty solid now, so I'll update to close #215.

pystac/utils.py
pjhartzell: Maybe? That's actually there to prevent a circular import from link.py I'll look into a more correct way to get this done.

pjhartzell: @gadomski I eliminated the double definition of `HREF`, which is used in a number of files. Let me know if you'd still like to rename `HREF` to `Href`.

gadomski: Eh, if it was already `HREF` then let's keep it. Thanks!

pystac/utils.py
pjhartzell: Done.

gadomski (approved):

PR #1126, opened 2023-05-18
change STACTypeError to create short informative message

ircwaves:
**Related Issue(s):**

- [pystac-client#130](https://github.com/stac-utils/pystac-client/issues/130) 

**Description:**

This creates a standardized message (which doesn't blow out the console) for `STACTypeError`s.

**PR Checklist:**

- [X] `pre-commit` hooks pass locally

### DIFF
--- a/pystac/errors.py
+++ b/pystac/errors.py
@@ -1,4 +1,4 @@
-from typing import Any, Optional, Union
+from typing import Any, Dict, Optional, Union
 
 
 class TemplateError(Exception):
@@ -24,7 +24,28 @@ class STACTypeError(Exception):
     a Catalog JSON was read in as an Item.
     """
 
-    pass
+    def __init__(
+        self,
+        bad_dict: Dict[str, Any],
+        expected: type,
+        extra_message: Optional[str] = "",
+    ):
+        """
+        Construct an exception with an appropriate error message from bad_dict and the
+        expected that it didn't align with.
+
+        Args:
+            bad_dict: Dictionary that did not match the expected type
+            expected: The expected type.
+            extra_message: message that will be appended to the exception message.
+        """
+        message = (
+            f"JSON (id = {bad_dict.get('id', 'unknown')}) does not represent"
+            f" a {expected.__name__} instance."
+        )
+        if extra_message:
+            message += " " + extra_message
+        super().__init__(message)
 
 
 class DuplicateObjectKeyError(Exception):
--- a/pystac/serialization/identify.py
+++ b/pystac/serialization/identify.py
@@ -248,7 +248,8 @@ def identify_stac_object(json_dict: Dict[str, Any]) -> STACJSONDescription:
     object_type = identify_stac_object_type(json_dict)
 
     if object_type is None:
-        raise pystac.STACTypeError("JSON does not represent a STAC object.")
+        extra_message = f"'type' attribute is {json_dict.get('type', 'not set')}"
+        raise pystac.STACTypeError(json_dict, pystac.STACObject, extra_message)
 
     version_range = STACVersionRange()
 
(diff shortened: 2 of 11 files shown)


pystac/errors.py
```diff
@@ -24,7 +24,26 @@ class STACTypeError(Exception):
...
+        extra_message: Optional[str] = "",
+    ):
+        """
+        Construct an exception with an appropriate error message from bad_dict and the
+        expected that it didn't align with.
+
+        Args:
+            bad_dict: Dictionary that did not match the expected type
+            expected: The expected type.  This should a :class:`~pystac.STACObject`, but
+                is left as `type` for the avoidance of circular imports.
```
gadomski, line 40: Could you use:

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from pystac import STACObject
```

tests/data-files/catalogs/test-case-9/catalog.json
gadomski, line 1: If we're adding a new fixture file, I'd prefer to move away from the "arbitrarily numbered folder" pattern. Maybe `tests/data-files/catalogs/invalid-catalog` or something?

pystac/errors.py
ircwaves: That would work save for some long standing [mypy hates AbstractBaseClasses](https://github.com/python/mypy/issues/4717) issues, and that ItemCollection isn't a STACObject.  So, I motion to proceed with comment brevity and leaving the annotation as `type`.

tests/data-files/catalogs/test-case-9/catalog.json
ircwaves: done.

ircwaves (commented):
Adjustments made.

gadomski (commented):
So this is a little annoying, but the `Catalog.from_file` error message isn't quite what I'd expect. Code (run from root):

```python
from pathlib import Path

from pystac import Catalog

path = (
    Path(__file__).parent
    / "tests"
    / "data-files"
    / "catalogs"
    / "invalid-catalog"
    / "catalog.json"
)
Catalog.from_file(str(path))
```

results in

```
pystac.errors.STACTypeError: JSON (id = broken_cat) does not represent a STACObject instance.
```

It's way better than it was, but the best text would end with "...a Catalog instance". Is there any way to get the "final" class type into the message in those `STACObject` alternate-constructor methods?

ircwaves:
> It's way better than it was, but the best text would end with "...a Catalog instance". Is there any way to get the "final" class type into the message in those `STACObject` alternate-constructor methods?

I thought that there was only one case where that should happen.  Let me take a look.

pystac/errors.py
```diff
@@ -24,7 +24,25 @@ class STACTypeError(Exception):
...
+        expected that it didn't align with.
+
+        Args:
+            bad_dict: Dictionary that did not match the expected type
+            expected: The expected type.
+            extra_message: message that will be appended to the exception message.
+        """
+        super().__init__(
+            f"JSON (id = {bad_dict.get('id', 'unknown')}) does not represent"
+            f" a {expected.__name__} instance.  {extra_message}"
```
gadomski, line 44: nit: remove whitespace if there's no extra message

pystac/serialization/identify.py
```diff
@@ -248,7 +248,7 @@ def identify_stac_object(json_dict: Dict[str, Any]) -> STACJSONDescription:
     object_type = identify_stac_object_type(json_dict)
 
     if object_type is None:
-        raise pystac.STACTypeError("JSON does not represent a STAC object.")
+        raise pystac.STACTypeError(json_dict, pystac.STACObject)
```
gadomski, line 251:
```suggestion
        if "type" in json_dict:
            extra_message = "'type' attribute is {json_dict['type']}"
        else:
            extra_message = None
        raise pystac.STACTypeError(json_dict, pystac.STACObject, extra_message)
```

gadomski (approved):

Issue #283, opened 2021-03-12
Add functionality to create STAC Items derived from other STAC Items


matthewhanson:
If a user wants to create a downstream product from existing STAC Items they should be able to create a new STAC Item for it.

A derived_from function should create a new STAC Item based on 1 or more existing items, and link to the existing ones using a `derived_from` link.

The case where a single Item is produced from a single the geometry and datetime would carry over, but do `platform`, or `instruments`? These are derived products, should they reference the original sensor? And what about extensions that may be in use? Do the derived products also use spectral bands (eo extension) or polarizations (sar extension)? I think we can come up with a reasonable list of fields that make sense to carry over.

In the case where you want to derive a single Item from multiple other Items it's much different, and probably deserves a separate function.  What is the new datetime if based on Items with different datetimes? If the geometries don't match is the new geometry an intersection of all the input geometries (**WARNING** additional dependency may be required if this is the case)

matthewhanson:
Put some more thought into what a `create_derived_item()` function should do.

Parameters:
- `items`: an array of items
- `id`: an new ID must be provided
- `collection`: a collection, if not provided do not set
- `geometry`: if not provided geometry would be a Multi-geometry of whatever type the source items are. If mixed, throw error
- `datetime`: an optional parameter specifying how to calculate the new datetime, e.g., first, last, midpoint

The function would:
- create new STAC Item with provided ID, geometry
- Calc geometry if not provided
- set `start_datetime` and `end_datetime` based on the input STAC Items
- set `datetime` based on input option, or *null*
- add links with `rel=derived_from` equal to all the items source `self` link (if available)
- carry over all properties that are the same across all items
- resulting item would have 0 assets and only `derived_from` links

jsignell:
I am wondering if it would be more ergonomic to have a method like `set_derived_from`. So you would create the new item using `Item(...` and then do `item.set_derived_from(original_item)`. It could just set the `derived_from` link. I know that gets rid of a lot of the magic, but it does seem to match existing patterns and it would put the burden (of calculating or inheriting the geometry for instance) on the user.

gadomski:
I also don't know the use-case for the many->one conversion. I haven't seen that pattern much in the wild. @matthewhanson can you provide more information about that use-case, since it adds so much complexity to this ticket (one->one is pretty chill I think).

> I am wondering if it would be more ergonomic to have a method like set_derived_from.

[...]

PR #1136, opened 2023-05-30
Add `get/add/remove_derived_from` to  pystac.Item

jsignell:
**Related Issue(s):**

- #283

**Description:**

Add `get_derived_from` and `set_derived_from` to  pystac.Item

**PR Checklist:**

### DIFF
--- a/pystac/item.py
+++ b/pystac/item.py
@@ -352,13 +352,13 @@ class Item(STACObject):
         else:
             return cast(Collection, collection_link.resolve_stac_object().target)
 
-    def add_derived_from(self, *items: Item) -> Item:
+    def add_derived_from(self, *items: Union[Item, str]) -> Item:
         """Add one or more items that this is derived from.
 
         This method will add to any existing "derived_from" links.
 
         Args:
-            items : Items to add to this item's derived_from links.
+            items : Items (or href of items) to add to derived_from links.
 
         Returns:
             Item: self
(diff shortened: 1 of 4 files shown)


pystac/item.py
```diff
@@ -352,6 +352,36 @@ def get_collection(self) -> Optional[Collection]:
         else:
             return cast(Collection, collection_link.resolve_stac_object().target)
 
+    def set_derived_from(self, item: Optional[Item]) -> Item:
```
ircwaves, line 355: I feel like this needs to take an iterable for the case where `itemA` is derived from `itemB` and `itemC`

jsignell: Ok cool cool, I'll update this with `add_derived_from(self, *items)`. I'm on the fence about how `remove_derived_from` should work or if it should exist at all. As I see it it could either remove all the `derived_from` links which can also be done with `item.remove_links("derived_from")` or it could remove one or more items from the `derived_from` links. I guess that is the more useful one so that it can be like the opposite of `add_`

ircwaves: I think you are correct that `remove_derived_from` is `self.remove_links("derived_from")`.  Doesn't seem needed, iff you stick with `set_derived_from(self, *items)`, since that will wipe the `derived_from`s out for you.  

If you move to `add_derived_from` (which linguistically implies it won't remove anything), then we might want to doc that `item.remove_links("derived_from")` can be used for removal in the docstring, if not adding the `remove_derived_from` convenience function.  Not sure about the case of removing a subset of the `derived_from` links.  Feels like that moves into the territory of `item.set_derived_from(link for link in item.get_derived_from() if conditional(link))`.

jsignell: Ok I have remove_derived_from accepting a single item_id. See what you think

ircwaves (approved):

PR #1140, opened 2023-06-01
Get assets by eo:bands common_name or name

### LINKED ISSUE #275: Referencing assets by eo:bands common_name
While asset keys can be anything, the EO extension provides a common set of band "colors".  If assets are 1 band per file and the band used has a common band name, it's really useful to be able to reference the asset by that (e.g., item.assets['ref'])

In sat-stac I did this by having an asset lookup function that would look for assets with that key as the common_name if a regular key did not exist:
https://github.com/sat-utils/sat-stac/blob/master/satstac/item.py#L93

This is probably not a good approach for PySTAC, eo is an extension (albeit a core extension) and maybe a more explicit function is better. 

Also some questions remain. If you want to find an asset for a given band are only 1-band files ok?  Or would you want to return any asset that has that band in it?  What if there are multiple assets?

jsignell:
**Related Issue(s):**

- Closes #275

**Description:**

This PR adds a new `get_assets` method on the `ItemEOExtension` class:

```python
import pystac_client
import planetary_computer
from pystac.extensions.eo import EOExtension

catalog = pystac_client.Client.open(
    "https://planetarycomputer.microsoft.com/api/stac/v1",
    modifier=planetary_computer.sign_inplace,
)

landsat = catalog.get_collection("landsat-c2-l2")
item = next(landsat.get_items())

eo_ext = EOExtension.ext(item)
eo_ext.get_assets()
```

image

**PR Checklist:**

### DIFF
--- a/pystac/extensions/eo.py
+++ b/pystac/extensions/eo.py
@@ -483,7 +483,7 @@ class ItemEOExtension(EOExtension[pystac.Item]):
             for key, asset in self.item.get_assets().items()
             if BANDS_PROP in asset.extra_fields
             and all(
-                v is None or v in [b.get(k) for b in asset.extra_fields[BANDS_PROP]]
+                v is None or any(v == b.get(k) for b in asset.extra_fields[BANDS_PROP])
                 for k, v in kwargs.items()
             )
         }
--- a/pystac/extensions/eo.py
+++ b/pystac/extensions/eo.py
@@ -470,7 +470,7 @@ class ItemEOExtension(EOExtension[pystac.Item]):
             name: If set, filter the assets such that only those with a
                 matching ``eo:band.name`` are returned.
             common_name: If set, filter the assets such that only those with a matching
-                ``eo:band.name`` are returned.
+                ``eo:band.common_name`` are returned.
 
         Returns:
             Dict[str, Asset]: A dictionary of assets that match ``name``
--- a/pystac/extensions/eo.py
+++ b/pystac/extensions/eo.py
@@ -459,6 +459,35 @@ class ItemEOExtension(EOExtension[pystac.Item]):
             return [Band(b) for b in bands]
         return None
 
+    def get_assets(
+        self,
+        name: Optional[str] = None,
+        common_name: Optional[str] = None,
+    ) -> Dict[str, pystac.Asset]:
+        """Get the item's assets where eo:bands are defined.
+
+        Args:
+            name: If set, filter the assets such that only those with a
+                matching ``eo:band.name`` are returned.
+            common_name: If set, filter the assets such that only those with a matching
+                ``eo:band.name`` are returned.
+
+        Returns:
+            Dict[str, Asset]: A dictionary of assets that match ``name``
+                and/or ``common_name`` if set or else all of this item's assets were
+                eo:bands are defined.
+        """
+        kwargs = {"name": name, "common_name": common_name}
+        return {
+            key: asset
+            for key, asset in self.item.get_assets().items()
+            if BANDS_PROP in asset.extra_fields
+            and all(
+                v is None or v in [b[k] for b in asset.extra_fields[BANDS_PROP]]
+                for k, v in kwargs.items()
+            )
+        }
+
     def __repr__(self) -> str:
         return "<ItemEOExtension Item id={}>".format(self.item.id)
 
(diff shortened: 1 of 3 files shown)


pystac/extensions/eo.py
```diff
@@ -459,6 +459,35 @@ def _get_bands(self) -> Optional[List[Band]]:
...
+        name: Optional[str] = None,
+        common_name: Optional[str] = None,
+    ) -> Dict[str, pystac.Asset]:
+        """Get the item's assets where eo:bands are defined.
+
+        Args:
+            name: If set, filter the assets such that only those with a
+                matching ``eo:band.name`` are returned.
+            common_name: If set, filter the assets such that only those with a matching
+                ``eo:band.name`` are returned.
```
gadomski, line 473:
```suggestion
                ``eo:band.common_name`` are returned.
```

pystac/extensions/eo.py
```diff
@@ -459,6 +459,35 @@ def _get_bands(self) -> Optional[List[Band]]:
...
+                and/or ``common_name`` if set or else all of this item's assets were
+                eo:bands are defined.
+        """
+        kwargs = {"name": name, "common_name": common_name}
+        return {
+            key: asset
+            for key, asset in self.item.get_assets().items()
+            if BANDS_PROP in asset.extra_fields
+            and all(
+                v is None or v in [b[k] for b in asset.extra_fields[BANDS_PROP]]
```
gadomski, line 486: What happens if the band doesn't have `name` or `common_name`? It _looks_ like it would `KeyError` but maybe I'm missing something.

gadomski (changes requested):

pystac/extensions/eo.py
jsignell: yeah it would KeyError, good call.

pystac/extensions/eo.py
```diff
@@ -459,6 +459,35 @@ def _get_bands(self) -> Optional[List[Band]]:
...
+                and/or ``common_name`` if set or else all of this item's assets were
+                eo:bands are defined.
+        """
+        kwargs = {"name": name, "common_name": common_name}
+        return {
+            key: asset
+            for key, asset in self.item.get_assets().items()
+            if BANDS_PROP in asset.extra_fields
+            and all(
+                v is None or v in [b.get(k) for b in asset.extra_fields[BANDS_PROP]]
```
gadomski, line 486:
```suggestion
                v is None or any(v == b.get(k) for b in asset.extra_fields[BANDS_PROP])
```

Sorry, should have caught this on first review, but this is _technically_ more performant b/c it short circuits when it finds a match.

tests/extensions/test_eo.py
```diff
@@ -441,3 +442,33 @@ def test_older_extension_version(ext_item: Item) -> None:
...
+    [
+        (dict(), 11),
+        ({"name": "B4"}, 1),
+        ({"common_name": "blue"}, 1),
+        ({"name": "B4", "common_name": "red"}, 1),
+        ({"name": "B4", "common_name": "green"}, 0),
+    ],
+)
+def test_get_assets(ext_item: pystac.Item, filter: Dict[str, str], count: int) -> None:
+    assets = EOExtension.ext(ext_item).get_assets(**filter)  # type:ignore
```
gadomski, line 458: The need to use `# type: ignore` here is a design smell, not with this PR, but with how extensions are implemented. Just noting that here so that we might remember later to design an extension interface that doesn't need this :-). No action required on your part.

gadomski (changes requested):

tests/extensions/test_eo.py
jsignell: Yeah totally. I messed around a bit with changing the output type of `EOExtension.ext()` to make it a `Union[ItemEOExtension, AssetEOExtension]` but then I didn't feel comfortable making that change for only this extension :(

pystac/extensions/eo.py
jsignell: ah yes! I meant to go back and see what I could check whether there was a better way, but I forgot. Yours looks good!

tests/extensions/test_eo.py
gadomski: Yeah, returning unions generally feels bad to me (unless it's `Union[None, Thing]`) so I'm ok with how things are, thanks.

gadomski (approved):

Issue #1097, opened 2023-04-17
`_repr_html_` causes GETs


jsignell:
Kind of hard to figure out where this issue should go since the code in question is in pystac itself. But the repercussions are more obvious on the pystac-client side. 

The issue is that `_repr_html_` implicitly triggers 5 GETs:

![image](https://user-images.githubusercontent.com/4806877/232535430-ebebcd9d-84fa-4526-9809-5f6f58b2c786.png)

Just using pystac it looks like there are only two requests:

![image](https://user-images.githubusercontent.com/4806877/232539929-61d22c55-6ffe-4b60-b4eb-26120bb2dcc0.png)

but I added the logging myself (https://github.com/stac-utils/pystac/pull/1096) so I might have missed some calls.

jsignell:
I have an extreme proposal. Opening a PR now

Issue #1124, opened 2023-05-15
HTML rendering should show asset keys


gadomski:
I can't see an obvious way to get the asset key from the HTML rendering (in a Jupyter notebook). This is one of the most important bits of info and should be front and center. (href blacked out because it contains a SAS, even though its not a special one).

<img width="653" alt="Screenshot 2023-05-15 at 4 10 40 PM" src="https://github.com/stac-utils/pystac/assets/58314/33857b39-c63c-41c2-b1bb-c936b88574ce">

PR #1142, opened 2023-06-05
Make `_repr_html_` look like `_repr_json_` output

jsignell:
**Related Issue(s):**

- Closes: #1097, #1124

**Description:**

This is similar to #1130 but it adds an html template that looks like the one that jlab uses for `_repr_json_`. There seemed to be agreement on #1130 that the json repr was a nice repr for stac objects but the lack of support in vscode (and in classic notebook) was a big deterrent.

I mimicked the style of the _repr_json_ output in jupyter lab, but am not depending on any classes since those don't seem to exist in vscode. I think the colors look  pretty ok in light and dark mode, but if people disagree there is probably some tweaks we can do.

vscode:
image

jupyter notebook:
image

jupyterlab light:

image

juptyerlab dark:

image

**PR Checklist:**


jsignell:
There is a minor issue with indentation and then I need to add a few tests, so I'm switching this to WIP

jsignell:
Ok here is some deeper nesting after a0eaf55
Screenshot from 2023-06-05 10-20-31
Screenshot from 2023-06-05 10-19-59

gadomski (approved):
I like this _a lot_, and added bonus that it's a code-removing PR 🎉!

pystac/stac_object.py
```diff
@@ -536,6 +538,14 @@ def to_dict(
         """
         raise NotImplementedError
 
+    def _repr_html_(self) -> str:
+        jinja_env = get_jinja_env()
+        if jinja_env:
+            template = jinja_env.get_template("JSON.jinja2")
+            return str(template.render(dict=self.to_dict()))
```
gadomski, line 545:
```suggestion
            return str(template.render(dict=self.to_dict(transform_hrefs=False)))
```

This will prevent resolving the whole tree, which I'm pretty sure is what we want.

gadomski (changes requested):

pystac/stac_object.py
jsignell: ah yes!

gadomski (approved):

jsignell:
I just noticed that the notebook size got kind of big with this change. Taking a look now to see if I can improve that experience

jsignell:
Ok I did some mild simplification. The sections are still expanded by default when rendered in github, but they are in the old repr as well:

MAIN: 
image

THIS PR:
image

PR #1155, opened 2023-06-08
Change `keep_parent` to `set_parent` and add flag on obj indicating parent override of href

jsignell:
**Related Issue(s):**

- #1116

**Description:**

Extends https://github.com/stac-utils/pystac/pull/1118 but in a potentially strange way.

**PR Checklist:**


m-mohr:
We are currently experimenting with this. Here's an example that we run:

```py
import pystac


def initial():
    root = pystac.Catalog("root", "root")
    projects = pystac.Catalog("projects", "projects")
    themes = pystac.Catalog("themes", "themes")
    variables = pystac.Catalog("variables", "variables")
    products = pystac.Catalog("products", "products")

    root.add_child(projects)
    root.add_child(themes)
    root.add_child(variables)
    root.add_child(products)

    extent = pystac.Extent(
        pystac.SpatialExtent([[0, 0, 1, 1]]),
        pystac.TemporalExtent([[None, None]])
    )

    theme_a = pystac.Catalog("theme_a", "theme_a")
    variable_a = pystac.Catalog("variable_a", "variable_a")
    project_a = pystac.Collection("project_a", "project_a", extent=extent.clone())
    product_a = pystac.Collection("product_a", "product_a", extent=extent.clone())

    themes.add_child(theme_a)
    variables.add_child(variable_a)
    projects.add_child(project_a)
    products.add_child(product_a)

    root.normalize_and_save("test_out", pystac.CatalogType.SELF_CONTAINED)


def adjust():
    root: pystac.Catalog = pystac.read_file("test_out/catalog.json")
    product_a = next(root.get_child("products").get_children())
    project_a = next(root.get_child("projects").get_children())
    theme_a = next(root.get_child("themes").get_children())
    variable_a = next(root.get_child("variables").get_children())

    project_a.add_child(product_a, keep_parent = True)
    theme_a.add_child(product_a, keep_parent = True)
    variable_a.add_child(product_a, keep_parent = True)

    root.save(pystac.CatalogType.SELF_CONTAINED, dest_href="test_out")


initial()
adjust()
```

This results in:
```
test_out/
├── catalog.json
├── products
│   ├── catalog.json
│   └── product_a
│       └── collection.json
├── projects
│   ├── catalog.json
│   └── project_a
│       └── collection.json
├── themes
│   ├── catalog.json
│   └── theme_a
│       └── catalog.json
└── variables
    ├── catalog.json
    └── variable_a
        ├── catalog.json
        └── product_a
            └── collection.json
```

The behavior seems very inconsistent. Why is there a product_a in variable_a? It either should be consistently also in themes and projects or only in products. I'd assume having it only in products, I think.

m-mohr:
@jsignell Yeah, we thought about using "related" as relation type, but the tooling support is very limited.

Debugging through the example, it looks like the self href is set incorrectly when reading in the catalog again. If I change in catalog.py, method add_child, line ~263 from `if self_href:` to `if self_href and not child.keep_parent:` it works for me, but I still need to figure out whether it has any side effects. Any thoughts?

jsignell:
>  If I change in catalog.py, method add_child, line ~263 from `if self_href:` to `if self_href and not child.keep_parent:` it works for me, but I did not figure out yet whether it has any side effects (tests not running for me locally). Any thoughts?

Ooh. Yeah that passes tests for me. I'll add that change to this branch. 

Do you have the capacity to add a test that exercises the same path as the example you provided? If not I can give it a shot. Just let me know.

m-mohr:
As I can't run the tests locally, it is always difficult to figure out whether the test is actually correct. So if you have some minutes, I'd appreciate if you could add it. Thanks :-)

jsignell:
> As I can't run the tests locally

Ah sorry I misunderstood and thought you just had some failing. Is there anything that I can help with in fixing your local dev setup?

> So if you have some minutes, I'd appreciate if you could add it. 

:+1: will do

gadomski:
@m-mohr windows support _should_ be good as of https://github.com/stac-utils/pystac/pull/1125 so if you can't run tests, that's a bug on **pystac**.

m-mohr:
Ah, thanks for the heads up, I'll try it again in the next days. I didn't notice that the tests have been updated recently to work on Windows. 👍

jsignell:
I added a test based on the example above and can confirm that it fails without 1d5d264 and passes with it :clap:

pystac/catalog.py
```diff
@@ -255,10 +255,12 @@ def add_child(
         child.set_root(self.get_root())
         if child.get_parent() is None or not keep_parent:
             child.set_parent(self)
+        elif keep_parent:
+            setattr(child, "keep_parent", True)
 
         # set self link
         self_href = self.get_self_href()
-        if self_href:
+        if self_href and not keep_parent:
```
m-mohr, line 263: Shouldn't we update an self href if the child had no parent before?

```suggestion
        if self_href and not child.keep_parent:
```

jsignell: ah yes sorry I missed the difference there

jsignell:
```suggestion
        if self_href and not getattr(child, "keep_parent", False):
```

pystac/catalog.py
```diff
@@ -255,10 +255,12 @@ def add_child(
         child.set_root(self.get_root())
         if child.get_parent() is None or not keep_parent:
             child.set_parent(self)
+        elif keep_parent:
+            setattr(child, "keep_parent", True)
```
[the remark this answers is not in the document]
jsignell: sure yeah if you prefer that it's all the  to me. Do you think we should flip the `keep_parent` kwarg name to match? Maybe `override_parent` so the default is True and then you have to explicitly set it to false?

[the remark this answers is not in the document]
gadomski: `override_parent` feels strong for what should be a reasonable default thing to do. `set_parent: bool = True` (since that's the method we end up calling)?

CHANGELOG.md
```diff
@@ -43,6 +43,11 @@
 - Catalog `get_item()`. Use `get_items(id)` instead ([#1075](https://github.com/stac-utils/pystac/pull/1075))
 - Catalog and Collection `get_all_items`. Use `get_items(recursive=True)` instead ([#1075](https://github.com/stac-utils/pystac/pull/1075))
 
+### Fixed
+
+- If a child catalog is linked to from two parents, the creation folder depends on the parent link instead of relying on the order of add_child calls ([#1118](https://github.com/stac-utils/pystac/pull/1118))
+- Don't override an existing parent link when determining the href's ([#1118](https://github.com/stac-utils/pystac/pull/1118))
```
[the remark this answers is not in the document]
jsignell: Actually I think the changelog entry could be consolidated a bit.

jsignell:
@m-mohr I just wanted to check if this is good with you, I changed the logic slightly so instead of `keep_parent=True` you would now use `set_parent=False`. I think this makes the behavior clearer. It's not an exact flip though. I made it so whether or not the parent is set in `add_item`/`add_child` depends solely on the `set_parent` karg rather than on that _and_ the presence of a parent. I think this is a nice logical tidy up, but just wanted to raise it to your attention.

gadomski (approved):
LGTM, but if possible let's get @m-mohr's 🆗 before merging.

m-mohr:
I'll try to review this in the next days, thanks.

pystac/catalog.py
```diff
@@ -230,11 +230,11 @@ def add_child(
         child: Union["Catalog", Collection],
         title: Optional[str] = None,
         strategy: Optional[HrefLayoutStrategy] = None,
-        keep_parent: bool = False,
+        set_parent: bool = True,
     ) -> None:
         """Adds a link to a child :class:`~pystac.Catalog` or
         :class:`~pystac.Collection`. This method will set the child's parent to this
-        object (except if a parent is set and keep_parent is true).
+        object (unless ``set_parent`` is False).
```
m-mohr, line 237: Should we mention here that the self link may also get updated if the parent it set?

pystac/catalog.py
```diff
@@ -280,11 +284,11 @@ def add_item(
         item: Item,
         title: Optional[str] = None,
         strategy: Optional[HrefLayoutStrategy] = None,
-        keep_parent: bool = False,
+        set_parent: bool = True,
     ) -> None:
         """Adds a link to an :class:`~pystac.Item`.
-        This method will set the item's parent to this object (except if a parent
-        is set and keep_parent is true).
+        This method will set the item's parent to this object (unless ``set_parent`` is
```
m-mohr, line 290: Should we mention here that the self link may also get updated if the parent it set?

m-mohr (approved):
Looks good to me. Our larger production catalog is generated exactly the same after the change, too. Added two comments to potentially improve the documentation a bit. Thank you!

Issue #846, opened 2022-07-14
Convenience methods for converting STAC objects / linked assets to data containers


TomAugspurger:
Hi all,

I'm tentatively making a pitch to add convenience methods for converting pystac objects (Asset, Item, ItemCollection, ...) and their linked assets to commonly used data containers (`xarray.Dataset`, `geopandas.GeoDataFrame`, `pandas.DataFrame`, etc.).

I'm opening this in `pystac` since **this is primarily for convenience, so that users can method-chain their way from STAC Catalog to data container**, and `pystac` owns the namespaces I care about. You can already do everything I'm showing today without any changes to `pystac` but it *feels less nice*. I really think that `pd.read_csv` is part of why Python is where it is today for data analytics; I want using STAC from Python to be as easy to use as `pd.read_csv`.

Secondarily, it *can* elevate the best-practice way to go from STAC to data containers, by providing a top-level method similar to `to_dict()`.

As a couple hypothetical examples, to give an idea:

```python
ds = (
    catalog
        .get_collection("sentinel-2-l2a")
        .get_item("S2B_MSIL2A_20220612T182919_R027_T24XWR_20220613T123251")
        .assets["B03"]
        .to_xarray()
)
ds
```

Or building a datacube from a pystac-client search (which subclasses pystac).

```python
ds = (
    catalog
        .search(collections="sentinel-2-l2a", bbox=bbox)
        .get_all_items()  # ItemCollection
        .to_xarray()
)
ds
```

## Implementation details

This would be optional. `pystac` would not add required dependencies on `pandas`, `xarray`, etc. It would merely provide the methods `Item.to_xarray`, `Asset.to_xarray`, ... Internally those methods would try to import the implementation and raise an `ImportError` if the optional dependencies aren't met at runtime.

Speaking of the implementations, there's a few things to figure out. Some relatively complicated conversions (like ItemCollection -> xarray) are implemented multiple times (https://stackstac.readthedocs.io/, https://odc-stac.readthedocs.io/en/latest/examples.html). `pystac` certainly wouldn't want to re-implement that conversion and would dispatch to one or either of those libraries (perhaps letting users decide with an `engine` argument).

Others conversions, like Asset -> Zarr, are so straightforward they haven't really been codified in a library yet (though I have a prototype at https://github.com/TomAugspurger/staccontainers/blob/086c2a7d46520ca5213d70716726b28ba6f36ba5/staccontainers/_xarray.py#L61-L63).
[...]

Issue #1051, opened 2023-03-17
Simpler extension interface


jsignell:
I just started interacting with the extensions in pystac and I was surprised by the API. As an xarray user I was expecting to just do `item.proj`. I think we can implement that by copying the [`_CachedAccessor` code from xarray](https://github.com/pydata/xarray/blob/main/xarray/core/extensions.py) (this seems to be what pandas has done as well)

### Existing API
```python
from pystac.extensions.proj import ProjectionExtension
import pystac 

PROJ_EXT_EXAMPLE_URL = "https://raw.githubusercontent.com/stac-extensions/projection/v1.1.0/examples/item.json"
item = pystac.Item.from_file(PROJ_EXT_EXAMPLE_URL)

proj_ext = ProjectExtension.ext(item)
proj_ext.proj
```

### Proposed API
```python
import pystac 

PROJ_EXT_EXAMPLE_URL = "https://raw.githubusercontent.com/stac-extensions/projection/v1.1.0/examples/item.json"
item = pystac.Item.from_file(PROJ_EXT_EXAMPLE_URL)

item.proj.epsg
```

## Adding extension to item

I'm not sure what adding new extensions to existing items should look like. 

### Existing API
```python
from pystac.extensions.eo import EOExtension

eo_ext = EOExtension.add_to(item)
eo_ext.cloud_cover = 0.7
```

### Options

In the most magical world (probably too magical) just setting an extension field on the item could add the extension to the item:
```python
item.eo.cloud_cover = 0.7
```

Another option would be to preserver the existing explicit `add_to` and `remove_from` functionality but maybe with the functions defined on the STACObject class so you could do:

```python
item.add_extension("eo")  
item.eo.cloud_cover = 0.7
```

Either way, I think trying to set the field directly: `item.proj = ProjectionExtension()` should fail.

xref: https://github.com/stac-utils/pystac/issues/448 where extensions are also being discussed.

jsignell:
I did a little implementation (no types yet) just to get a sense of what this feels like and I am pretty sure I like it in the `get` case.

jsignell:
I was definitely approaching this more from the perspective of someone reading rather than writing STAC. I just spent some time looking over [stactools/sentinel2](https://github.com/stactools-packages/sentinel2/blob/main/src/stactools/sentinel2/stac.py) to get a sense of what the write patterns look like. So with that in mind I am going to try to answer Pete's questions. But if there is just no appetite for this type of change, feel free to close this ticket.

[...]

PR #1161, opened 2023-06-14
Xarray Assets Extension class

jsignell:
**Related Issue(s):**

- #846 

**Description:**

EDIT: I removed the `open` method, so this is just a regular extension class now.

This PR implements an extension class for xarray-assets extension with a special `open` method for extended pystac.Assets as described in https://github.com/stac-utils/pystac/issues/846#issuecomment-1587441466

```python
import pystac
from pystac.extensions.xarray_assets import XarrayAssetsExtension

daymet_collection = pystac.Collection.from_file(
    "https://planetarycomputer.microsoft.com/api/stac/v1/collections/daymet-daily-hi"
)
daymet_asset = daymet_collection.assets["zarr-abfs"]
xr_ext = XarrayAssetsExtension.ext(daymet_asset, add_if_missing=True)
xr_ext.open()
```
**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - `KML` as a built in media type ([#1127](https://github.com/stac-utils/pystac/issues/1127))
 - `move/copy/delete` operations for local Assets ([#1158](https://github.com/stac-utils/pystac/issues/1158))
 - Latest core STAC spec jsonshemas are included in pytstac and used for validation ([#1165](https://github.com/stac-utils/pystac/pull/1165))
+- Xarray Assets Extension class ([#1161](https://github.com/stac-utils/pystac/pull/1161))
 
 ### Changed
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -107,6 +107,7 @@ PySTAC provides support for the following STAC Extensions:
 * :mod:`Timestamps <pystac.extensions.timestamps>`
 * :mod:`Versioning Indicators <pystac.extensions.version>`
 * :mod:`View Geometry <pystac.extensions.view>`
+* :mod:`Xarray Assets <pystac.extensions.xarray_assets>`
 
 The following classes are used internally to implement these extensions and may be used
 to create custom implementations of STAC Extensions not supported by the library (see
--- a/docs/api/extensions.rst
+++ b/docs/api/extensions.rst
@@ -32,3 +32,4 @@ pystac.extensions
     timestamps.TimestampsExtension
     version.VersionExtension
     view.ViewExtension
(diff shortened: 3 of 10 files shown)


jsignell:
Not super ergonomic with the current extension access system, but I'm optimistic that the extension API can be revamped a la #1051 without a substantial rewrite of the classes.

gadomski (changes requested):
Couple high level questions:

- Did you consider adding **xpystac** as a optional dependency (ala **orjson**)?
- How do you see this model working for "containers" that aren't tied to a specific extension? E.g. @TomAugspurger's proposed `to_geopandas` for an `ItemCollection`?
- I think this deserves more than just API docs, e.g. updating how-to-read-data-from-stac.ipynb to use the new extension interface?

jsignell:
>     * Did you consider adding **xpystac** as a optional dependency (ala **orjson**)?

I guess that's a good idea in case we need to pin it later on?

>     * How do you see this model working for "containers" that aren't tied to a specific extension? E.g. @TomAugspurger's proposed `to_geopandas` for an `ItemCollection`?

Because ItemCollections can't have extensions or because there would be no need for a geopandas extension? I was kind of thinking the table extension could be the geopandas extension. I guess alternatively there could be a generic reader extension but that might be too far of a stretch from what extensions are meant to be.

>     * I think this deserves more than just API docs, e.g. updating how-to-read-data-from-stac.ipynb to use the new extension interface?

Yeah definitely. I wasn't sure if people would like this idea, so I didn't bother with that yet.

gadomski:
> I guess that's a good idea in case we need to pin it later on?

Yeah, and it makes it clearer that there's **pystac** functionality that uses **xpystac**, instead of just tests.

> ItemCollections can't have extensions

Yup, this.

> I was kind of thinking the table extension could be the geopandas extension.

But theoretically you could want to make a dataframe out of something that doesn't have the table extension in its `stac_extensions` list, correct?

> Yeah definitely. I wasn't sure if people would like this idea, so I didn't bother with that yet.

👍🏼

jsignell:
> But theoretically you could want to make a dataframe out of something that doesn't have the table extension in its `stac_extensions` list, correct?

Yeah but couldn't you just add the table extension in that case?

gadomski:
> Yeah but couldn't you just add the table extension in that case?

If you don't _need_ the table extension to put the STAC stuff into a data frame, IMO you shouldn't have to add the table extension. Otherwise, I could see a scenario where someone loads some STAC, adds the table extension to do some data frame work, then has to strip the table extension back out again before saving the STAC somewhere else -- feels hackey.

gadomski (approved):

PR #1162, opened 2023-06-15
Tests don't depend on external files

### LINKED ISSUE #494: Make sure all test data is local
Some tests seem to be [pulling files from for example https://cmr.earthdata.nasa.gov/stac/](https://cmr.earthdata.nasa.gov/stac/), which is causing false positives and slowing down test runs. We should probably store a copy of such data locally or create minimal equivalent test data explicitly for the relevant purpose, to avoid these issues.

### LINKED ISSUE #949: Avoid network calls in tests suite

jsignell:
**Related Issue(s):**

- closes #949, closes #494 

**Description:**

The vast majority of these changes have to do with validation so they might not be necessary if 

**PR Checklist:**


jsignell:
Ok so apparently the tests depend on whether urllib3 is installed or not :facepalm:

pystac/validation/local_validator.py
```diff
@@ -66,6 +57,15 @@ def item_validator(version: str = VERSION) -> Draft7Validator:
             ] = _read_schema(f"stac-spec/v{version}/{name}.json")
         return Draft7Validator(schema, resolver=resolver)
 
+    def catalog_validator(self, version: str = VERSION) -> Draft7Validator:
```
jsignell, line 60: I noticed that these were not figuring out the items sub-schemas properly, so I switched these to just all have everything stored in their resolvers.

gadomski (approved):

PR #1168, opened 2023-06-21
Include collection assets in `make_all_asset_hrefs_relative/absolute` 

jsignell:
**Related Issue(s):**

- https://github.com/stac-utils/stactools/issues/423

**Description:**

Broadly "all_assets" for a catalog should include collection assets. Specifically I needed this functionality to solve the stactools issue

**PR Checklist:**

### DIFF
--- a/pystac/collection.py
+++ b/pystac/collection.py
@@ -20,7 +20,7 @@ from typing import (
 from dateutil import tz
 
 import pystac
-from pystac import CatalogType, STACObjectType
+from pystac import CatalogType, STACError, STACObjectType
 from pystac.asset import Asset
 from pystac.catalog import Catalog
 from pystac.errors import DeprecatedWarning, ExtensionNotImplemented, STACTypeError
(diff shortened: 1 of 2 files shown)


pystac/collection.py
```diff
@@ -760,6 +766,50 @@ def delete_asset(self, key: str) -> None:
...
+            href = asset.href
+            if not is_absolute_href(href):
+                if self_href is None:
+                    self_href = self.get_self_href()
+                    if self_href is None:
+                        raise STACError(
+                            "Cannot make relative asset HREFs absolute "
+                            "if no self_href is set."
+                        )
+                asset.href = make_absolute_href(asset.href, self_href)
```
ircwaves, line 809: There could be something I'm missing here, but I think this can be de-nested like this:
```suggestion
        self_href = self.get_self_href()
        if self_href is None:
            raise STACError(
                "Cannot make relative asset HREFs absolute "
                "if no self_href is set."
            )
        for asset in self.assets.values():
            href = asset.href
            if not is_absolute_href(href):
                asset.href = make_absolute_href(asset.href, self_href)
```

ircwaves (approved):
I don't think my suggestion is required, but it reads cleaner to my brain.  As a bonus, a commit will bonk the CI when landsat's stac-api is back up and running.

pystac/collection.py
jsignell: oh yeah I should have mentioned that I just copied this over from the Item class. I like yours though. I'll change it in both places.

jsignell: I kept the error raising within the assets loop since we probably don't need to raise an error if there are no assets that need to be changed.

PR #1169, opened 2023-06-21
Let non-hierarchical links to hierachical elements be relative paths for self-contained catalogs

### LINKED ISSUE #1112: Adding custom links between sibling catalogs creates absolute HREFs
In the following example, we have a `root` catalog which includes two sub-catalogs `a` and `b`. I now want to highlight a reference between the two siblings, which cannot be replaced by a parent-child link:

```python
import pystac

root = pystac.Catalog("root", "root")
a = pystac.Catalog("a", "a")
b = pystac.Catalog("b", "b")

root.add_child(a)
root.add_child(b)

a.add_link(pystac.Link("related", b))

root.normalize_and_save("test_out", pystac.CatalogType.SELF_CONTAINED)
```

Even when I now save it as "self-contained", I now have an absolute link from `a` to `b`:

```json
{
  "type": "Catalog",
  "id": "a",
  "stac_version": "1.0.0",
  "description": "a",
  "links": [
    {
      "rel": "root",
      "href": "../catalog.json",
      "type": "application/json"
    },
    {
      "rel": "related",
      "href": "/home/<user>/test_out/b/catalog.json"
    },
    {
      "rel": "parent",
      "href": "../catalog.json",
      "type": "application/json"
    }
  ],
  "stac_extensions": []
}
```

I would have expected something like that:
```json
...
   {
      "rel": "related",
      "href": "../b/catalog.json"
    },
...
```
I can't seem to get rid of the absolute links, even when using `stac copy` or similar commands.

jpolchlo:
**Related Issue(s):**
Closes #1112

**Description:**
This is an attempt to address the unexpected and maybe undesirable behavior identified in #1112.  For this, we consider the case of a catalog tree which is defined using hierarchical links.  When serialized out as a self-contained catalog, using `to_dict` yields relative links for all elements.  Now, when we add additional links that have non-hierarchical `rel` types, but refer to entities which are in the hierarchical tree, the non-standard links are serialized with absolute hrefs.

For the following tree
```mermaid
flowchart TD
    root -- child --> A
    root --child--> B
    A -. sibling .-> B
```

When we serialize `A`, we want to see
```json
{
    "type": "Catalog",
    "id": "a",
    "stac_version": "1.0.0",
    "description": "a",
    "links": [
        {
            "rel": "root",
            "href": "../catalog.json",
            "type": "application/json"
        },
        {
            "rel": "sibling",
            "href": "../b/catalog.json"
        },
        {
            "rel": "parent",
            "href": "../catalog.json",
            "type": "application/json"
        },
        {
            "rel": "self",
            "href": "/opt/data/stac/a/catalog.json",
            "type": "application/json"
        }
    ]
}
```
because the referent of the `sibling` link has been serialized elsewhere in the catalog tree using a relative link.  This is the result that is delivered by this PR.  Whether this should be the default behavior is a separate concern, and we can have that discussion below.

**PR Checklist:**

### DIFF
--- a/pystac/link.py
+++ b/pystac/link.py
@@ -183,10 +183,15 @@ class Link(PathLike):
                 *pystac.EXTENSION_HOOKS.get_extended_object_links(self.owner),
             ]
             # if a hierarchical link with an owner and root, and relative catalog
-            if root and root.is_relative() and self.rel in rel_links:
-                owner_href = self.owner.get_self_href()
-                if owner_href is not None:
-                    href = make_relative_href(href, owner_href)
+            if root and root.is_relative():
+                if self.rel in rel_links:
+                    owner_href = self.owner.get_self_href()
+                    if owner_href is not None:
+                        href = make_relative_href(href, owner_href)
+                elif self.target in root.get_target_hierarchy():
+                    owner_href = self.owner.get_self_href()
+                    if owner_href is not None:
+                        href = make_relative_href(href, owner_href)
 
         return href
 
--- a/pystac/link.py
+++ b/pystac/link.py
@@ -184,11 +184,7 @@ class Link(PathLike):
             ]
             # if a hierarchical link with an owner and root, and relative catalog
             if root and root.is_relative():
-                if self.rel in rel_links:
-                    owner_href = self.owner.get_self_href()
-                    if owner_href is not None:
-                        href = make_relative_href(href, owner_href)
-                elif self.target in root.get_target_hierarchy():
+                if self.rel in rel_links or root.target_in_hierarchy(self.target):
                     owner_href = self.owner.get_self_href()
                     if owner_href is not None:
                         href = make_relative_href(href, owner_href)
--- a/pystac/stac_object.py
+++ b/pystac/stac_object.py
@@ -138,26 +138,41 @@ class STACObject(ABC):
         return remove
 
     def target_in_hierarchy(self, target: Union[str, STACObject]) -> bool:
-        """Recursively collects all the targets referred to by the hierarchical
-        links of the current STACObject.
+        """Determine if target lin is somewhere in the hierarchical link tree of
+        a STACObject.
+
+        Args:
+            target: A string or STACObject describing the target to search for
 
         Returns:
-            Set[Union[str, STACObject]]: All encountered targets
+            bool: Returns True if the target was found in the hierarchical link tree
+                for the current STACObject
         """
 
         def traverse(
             obj: Union[str, STACObject], visited: Set[Union[str, STACObject]]
-        ) -> Set[Union[str, STACObject]]:
+        ) -> bool:
+            if obj == target:
+                return True
             if isinstance(obj, str):
-                return visited
+                return False
 
-            hierarchical_links = [link for link in obj.links if link.is_hierarchical()]
-            new_targets = set([link.target for link in hierarchical_links]) - visited
-            for target in new_targets:
-                visited = traverse(target, visited.union(set([target])))
-            return visited
+            new_targets = [
+                link.target
+                for link in obj.links
+                if link.is_hierarchical() and link.target not in visited
+            ]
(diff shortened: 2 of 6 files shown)


pystac/link.py
```diff
@@ -183,10 +183,15 @@ def get_href(self, transform_href: bool = True) -> Optional[str]:
...
-                    href = make_relative_href(href, owner_href)
+            if root and root.is_relative():
+                if self.rel in rel_links:
+                    owner_href = self.owner.get_self_href()
+                    if owner_href is not None:
+                        href = make_relative_href(href, owner_href)
+                elif self.target in root.get_target_hierarchy():
+                    owner_href = self.owner.get_self_href()
+                    if owner_href is not None:
+                        href = make_relative_href(href, owner_href)
```
gadomski, line 194: What do you think about this refactor?

```suggestion
                if self.rel in rel_links or root.has_descendant(self.target):
                    owner_href = self.owner.get_self_href()
                    if owner_href is not None:
                        href = make_relative_href(href, owner_href)
```

w/ the necessary changes to add a `has_descendant` method to `Catalog` in place of `get_target_hierarchy`? This would read a bit cleaner to me at least.

gadomski (changes requested):
Can you add a test case that mirrors the example in the PR description? Thanks!

pystac/link.py
jpolchlo: Oops.  Yeah.  Overthought this.  `or` short circuits.  Duh.

pystac/stac_object.py
```diff
@@ -136,6 +137,28 @@ def remove_hierarchical_links(self, add_canonical: bool = False) -> List[Link]:
         self.links = keep
         return remove
 
+    def target_in_hierarchy(self, target: Union[str, STACObject]) -> bool:
+        """Recursively collects all the targets referred to by the hierarchical
+        links of the current STACObject.
+
+        Returns:
+            Set[Union[str, STACObject]]: All encountered targets
```
gadomski, line 145: Docs need updating for the new method signature.

pystac/stac_object.py
```diff
@@ -136,6 +137,28 @@ def remove_hierarchical_links(self, add_canonical: bool = False) -> List[Link]:
...
+            if isinstance(obj, str):
+                return visited
+
+            hierarchical_links = [link for link in obj.links if link.is_hierarchical()]
+            new_targets = set([link.target for link in hierarchical_links]) - visited
+            for target in new_targets:
+                visited = traverse(target, visited.union(set([target])))
+            return visited
+
+        return target in traverse(self, set([self]))
```
gadomski, line 160: This should be refactored to short-circuit when the target is encountered, so it doesn't walk the whole tree if it doesn't need to.

gadomski (changes requested):

pystac/stac_object.py
jpolchlo: 🤦

pystac/stac_object.py
```diff
@@ -136,6 +137,43 @@ def remove_hierarchical_links(self, add_canonical: bool = False) -> List[Link]:
         self.links = keep
         return remove
 
+    def target_in_hierarchy(self, target: Union[str, STACObject]) -> bool:
+        """Determine if target lin is somewhere in the hierarchical link tree of
```
gadomski, line 141:
```suggestion
        """Determine if target is somewhere in the hierarchical link tree of
```

pystac/stac_object.py
```diff
@@ -136,6 +137,43 @@ def remove_hierarchical_links(self, add_canonical: bool = False) -> List[Link]:
...
+        a STACObject.
+
+        Args:
+            target: A string or STACObject describing the target to search for
```
gadomski, line 145:
```suggestion
            target: A string or STACObject to search for
```

gadomski (approved):

PR #1211, opened 2023-08-28
Use three digits for ~= v0.* deps

gadomski:
**Description:**

I realized recently that using `~=` and two digits for dependencies that start with `0.` is more-or-less meaningless, since it expands to `>=0.x,<1.0`. This updates to add a third digit for `0.*` dependencies.

This also fixes our bench check caused by `asv` breaking change in their CLI that popped up due to the dependency pinning problem. Doesn't need a changelog.

**PR Checklist:**


jsignell (approved):
Nice catch!

Issue #1186, opened 2023-07-07
Don't use `RefResolver`


gadomski:
As of **jsonschema** v4.18, `RefResolver` is deprecated:

```
DeprecationWarning: jsonschema.RefResolver is deprecated as of v4.18.0, in favor of the https://github.com/python-jsonschema/referencing library, which provides more compliant referencing behavior as well as more flexible APIs for customization. A future release will remove RefResolver. Please file a feature request (on referencing) if you are missing an API for the kind of customization you need.
```

This issue is to track the need to update to The New Way.

gadomski:
Thanks for the report Vincent! I wonder if we should ceil **jsonschema** below v4.18 for the time being? What's your sense of how broken things are over there?

vincentsarago:
Yep that's what I did in #1189 😅

gadomski:
Yup, agreed. I'll open a PR to fix.

gadomski:
#1189 is released in [v1.8.2](https://github.com/stac-utils/pystac/releases/tag/v1.8.2).

jsignell:
I'm going to look into this a bit today to see if the timing is right to pick it up

2320sharon:
Hi @jsignell, thank you for addressing this issue. It's blocking the resolution of https://github.com/stac-utils/pystac/issues/1214, which is affecting Jupyter Lab in downstream applications. We truly appreciate your efforts!

PR #1215, opened 2023-09-08
Remove references to `RefResolver`

jsignell:
**Related Issue(s):**

- #1186 

**Description:**

I am working though the recommendations on https://python-jsonschema.readthedocs.io/en/latest/referencing/ these are the steps as I understand them:

- [x] Replace `RefResolver` with `Registry`
- [x] Create a callable that loads referenced relative schemas
- [x] Create a callable that loads referenced absolute schemas

**PR Checklist:**


gadomski:
> It was never really intended for public consumption.

Yeah, in retrospect `pystac.validation.local_validator` probably should have been `pystac.validation._local_validator`. Maybe we "privatize" the whole module and drop a deprecation warning on the "public" version? E.g. (crib version):

```python
# in pystac/validation/local_validator.py

warnings.warn("This module is deprecated and will be removed in v2")
from ._local_validator import *
__all__ = ["the", "things", "we", "need", "to", "export"]
```

gadomski (commented):
Makes sense to me ... I think the only outstanding question is https://github.com/stac-utils/pystac/pull/1215#issuecomment-1718369556.

jsignell:
Hmm the functionality in the old vs new are pretty different. Maybe I should just leave the old `local_validator.py` alone and toss a deprecation warning on it and then create a new `_local_validator.py`.

jsignell:
Ok see what you think about this approach to deprecation. Note that it probably won't have quite as good coverage as main because the deprecated class is untested. I can add some new tests if that is too gross.

pystac/validation/local_validator.py
```diff
@@ -13,6 +15,48 @@
...
+            )
+        },
+    }
+
+
+############################### DEPRECATED #################################
+
 ITEM_SCHEMA_URI = (
     f"https://schemas.stacspec.org/v{VERSION}/item-spec/json-schema/item.json"
 )
```
gadomski, line 62: If this constant isn't deprecated, we should move it above the comment. If it is, we should deprecate it with `__getattr__`.

gadomski (commented):
Just one small nit then we should be good, thanks!

pystac/validation/local_validator.py
jsignell: Yeah I am trying to deprecate those constants. I didn't know about the `__getattr__` trick!

gadomski (approved):

Issue #793, opened 2022-04-22
Allow extensions for un-parented assets


gadomski:
Its not uncommon to create `Assets` before `Items`, as many of the item properties (e.g. `geometry`) can be derived from `Assets`. This currently requires a strange dance, where you need to:
1. Create assets
2. Generate item attributes from those assets
3. Create the item
4. Add extensions for those assets

If we could extend assets _before_ they are added to the item, this workflow could be simpler.

## One possible solution

`Assets` also have a `stac_extensions` field that is excluded from its representative dictionary. This field is populated when the `Asset` is created when reading an `Item`, and propogated up to the `Item` on, e.g., `to_dict()`. This would allow something like (pseudo-python):

```python
asset = Asset(href="my-cog.tif")
geometry, proj = read_proj_from_tif(asset.href)  # example utility method that reads proj and geometry information from a tif
projection = ProjectionExtension.ext(asset, add_if_missing=True)
projection.update(**proj)
item = Item(geometry=geometry, ...)  # omitting the rest 
item.add_asset(asset)
```

PR #1243, opened 2023-09-27
Simpler extension interface

### LINKED ISSUE #1051: Simpler extension interface

jsignell:
**Related Issue(s):**

- Closes #1051 

**Description:**

When the extension is implemented:
```python
import pystac

PROJ_EXT_EXAMPLE_URL = "https://raw.githubusercontent.com/stac-extensions/projection/v1.1.0/examples/item.json"

item = pystac.Item.from_file(PROJ_EXT_EXAMPLE_URL)
item.ext.proj.epsg
# 32659
```

When the extension isn't implemented:

```python
item.ext.eo
```
```python-traceback
---------------------------------------------------------------------------
ExtensionNotImplemented                   Traceback (most recent call last)
Cell In[5], line 1
----> 1 item.ext.eo

File ~/pystac/pystac/extensions/ext.py:47, in ItemExt.eo(self)
     45 @property
     46 def eo(self) -> EOExtension[pystac.Item]:
---> 47     return EOExtension.ext(self.stac_object)

File ~/pystac/pystac/extensions/eo.py:408, in EOExtension.ext(cls, obj, add_if_missing)
    397 """Extends the given STAC Object with properties from the
    398 :stac-ext:`Electro-Optical Extension <eo>`.
    399 
   (...)
    405     pystac.ExtensionTypeError : If an invalid object type is passed.
    406 """
    407 if isinstance(obj, pystac.Item):
--> 408     cls.validate_has_extension(obj, add_if_missing)
    409     return cast(EOExtension[T], ItemEOExtension(obj))
    410 elif isinstance(obj, pystac.Asset):

File ~/pystac/pystac/extensions/base.py:201, in ExtensionManagementMixin.validate_has_extension(cls, obj, add_if_missing)
    198     cls.add_to(obj)
    200 if not cls.has_extension(obj):
--> 201     raise pystac.ExtensionNotImplemented(
    202         f"Could not find extension schema URI {cls.get_schema_uri()} in object."
    203     )

ExtensionNotImplemented: Could not find extension schema URI https://stac-extensions.github.io/eo/v1.1.0/schema.json in object.

```

Add an extension

```python
item.ext.add("eo")
```

**PR Checklist:**


gadomski (commented):
Concept seems good to me! Hopefully this will be a pretty low maintenance burden once it gets set up, since most of the work is still in the extensions themselves.

jsignell:
okie dokie! It's greeeeennnn over here

jsignell:
I guess I can do the docs updates on this branch as well.

pystac/asset.py
```diff
@@ -261,6 +262,12 @@ def delete(self) -> None:
         href = _absolute_href(self.href, self.owner, "delete")
         os.remove(href)
 
+    @property
+    def ext(self) -> AssetExt:
```
gadomski, line 266: We should docstring these public properties. I'll just drop this one comment about that.

pystac/extensions/ext.py
gadomski, line 1: I like how this turned out -- it's now very easy to see which extensions apply to which object types. 👍🏼

gadomski (changes requested):
No notes on the impl, I'm 👍🏼 ^ 2. Looks great. I think we just need docstrings on the functions, and maybe a "Making a new extension class" tutorial? The tutorial could be its own ticket too if you want to push this in first.

philvarner:
1. I'd prefer if `item.ext.eo` returned None rather than throwing an Exception. I guess it's more Pythonic to throw an exception for everything, but dislike that convention, since it's not an error condition if the Item doesn't implement the extension.

With that, I could write code as:

```
if eo := item.ext.eo:
  print(eo.cloud_cover)
```

instead of:

```
try:
  print(eo.cloud_cover)
except ExtensionNotImplemented:
  pass
```

2. I'd prefer if `item.ext.add`, instead of accepting a raw string parameter (presumably the Field Name Prefix?) took a parameter of a custom type (e.g., `Extension`) that all extensions had to implement, as a form of dynamic enum. That would give better typechecking than just a string, which in practice is pretty error-prone.

gadomski:
> That would give better typechecking than just a string, which in practice is pretty error-prone.

I think the use of `Literal` here makes the type checking better: https://github.com/stac-utils/pystac/pull/1243/files#diff-3a39d1c18e4343de9e2ad0a08ee71c3e616589cbc3677eeac625e679f195ea16R27-R47.

> took a parameter of a custom type (e.g., Extension)

I believe @jsignell started with an `Enum` implementation, but it felt clunky. I think a soft design goal here might be "access and set extension attributes without additional imports"?

> I'd prefer if item.ext.eo returned None rather than throwing an Exception. 

The only downside here is that for type checking, `item.ext.eo` then becomes `Optional[Extension]`, so you have to do `assert item.ext.eo` before accessing the attributes to make **mypy** happy. The "throw if absent" pattern makes the "I know it's there" use case slightly less verbose. 🤷🏼

jsignell:
> > I'd prefer if item.ext.eo returned None rather than throwing an Exception.
> 
> The only downside here is that for type checking, `item.ext.eo` then becomes `Optional[Extension]`, so you have to do `assert item.ext.eo` before accessing the attributes to make **mypy** happy. The "throw if absent" pattern makes the "I know it's there" use case slightly less verbose. 🤷🏼

I didn't call it out in my example, but there is also a `has` method so the no-error incantation would be:

```python
if item.ext.has("eo"):
    print(item.ext.eo.cloud_cover)
```

Does that feel better?

philvarner:
I'm fine with whatever y'all decide on.

pystac/extensions/ext.py
```diff
@@ -0,0 +1,290 @@
...
+            return cast(
+                bool, _get_class_by_name(name).has_extension(self.stac_object.owner)
+            )
+
+    def add(self, name: EXTENSION_NAMES) -> None:
+        if self.stac_object.owner is None:
+            raise pystac.STACError(
+                f"Attempted to add extension='{name}' for an Asset with no owner. "
+                "Use Asset.set_owner and then try to add the extension again."
+            )
```
constantinius, line 213: In the past, this was a real pain to work with, and ended up with strange workarounds. e.g in some stactools packages it is common to have a metadata mapping class, preparing assets for the stac producer to add to the item. It currently cannot enrich the assets with any metadata from extensions.
I would be very glad if it was somehow possible to add extensions to assets, even if they're not yet added to their owners yet.
A suggestion: maybe storing all extension URIs in a separate list, which gets added to the parents list once added there.

gadomski: This is a slightly different issue than what this PR addresses, and is tracked here: https://github.com/stac-utils/pystac/issues/793. I 100% agree with you for the record.

jsignell:
Finally circled back. I took a pass at the docs and tutorials as well.

gadomski (approved):

gadomski:
🎉 thanks for this @jsignell really stoked about it.

PR #1256, opened 2023-10-08
Assets keep their own `stac_extensions`

constantinius:
so that they can be extended without being added to an Item.
Deprecating validate_owner_has_extension
Switching from validate_owner_has_extension to ensure_has_extension
in extensions dealing with assets
Extracting a final list of assets in the `to_dict`
function for Items and Collections.

**Related Issue(s):**

- #793

**Description:**

**PR Checklist:**

### DIFF
--- a/pystac/asset.py
+++ b/pystac/asset.py
@@ -35,6 +35,8 @@ class Asset:
         extra_fields : Optional, additional fields for this asset. This is used
             by extensions as a way to serialize and deserialize properties on asset
             object JSON.
+        stac_extensions : Optional, a list of schema URIs for STAC Extensions
+            implemented by this STAC Asset.
     """
 
     href: str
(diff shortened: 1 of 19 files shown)


constantinius:
765a05f6b28ea3ae245db94829dfa6845dc17154 does that.

This makes the extension handling more complex now, though, as we have to introduce a couple of new code paths as we cannot simply treat assets as objects having a `stac_extensions: Optional[List[str]]`.

What would you say?

gadomski:
> This makes the extension handling more complex now, though, as we have to introduce a couple of new code paths as we cannot simply treat assets as objects having a stac_extensions: Optional[List[str]].

You're right, that is pretty awkward. Maybe `_stac_extensions` was a bad idea 😬. Let's remove 765a05f for now and keep it at `stac_extensions`. Thanks for trying that out 🙇🏼.

PR #1261, opened 2023-10-12
pyupgrade to 3.9

### LINKED ISSUE #1235: Run pyupgrade to get to 3.9
Once https://github.com/stac-utils/pystac/issues/1148 is closed, we should (separately) run [pyupgrade](https://github.com/asottile/pyupgrade) (from a Python 3.9 environment) to update the code to use new language features.

gadomski:
**Related Issue(s):**

- Closes #1235 

**Description:**

Applies pyupgrade, and installs a pre-commit rachet to make sure we don't backslide. I don't use a `local` pre-commit version (like the other hooks) b/c users shouldn't have to have it installed in their environment IMO.

No functional changes, and it's only formatting so no CHANGELOG needed.

**PR Checklist:**

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,6 +2,12 @@
 # Please run `pre-commit run --all-files` when adding or changing entries.
 
 repos:
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v3.15.0
+    hooks:
+      - id: pyupgrade
+        language_version: "3.9"
+
   - repo: local
     hooks:
       - id: ruff
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,8 @@ repos:
     rev: v3.15.0
     hooks:
       - id: pyupgrade
-        language_version: "3.9"
+        args:
+          - "--py39-plus"
 
   - repo: local
     hooks:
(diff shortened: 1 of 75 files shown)


.pre-commit-config.yaml
```diff
@@ -2,6 +2,12 @@
 # Please run `pre-commit run --all-files` when adding or changing entries.
 
 repos:
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v3.15.0
+    hooks:
+      - id: pyupgrade
+        language_version: "3.9"
```
jsignell, line 9: Sometimes I just run this hook and then don't actually include it in the pre-commit-config. Just  sometime to consider in case the lint issues are hard to fix.

pystac/asset.py
```diff
@@ -41,38 +41,38 @@ class Asset:
     href: str
     """Link to the asset object. Relative and absolute links are both allowed."""
 
-    title: Optional[str]
+    title: str | None
```
jsignell, line 44: :heart_eyes:

.pre-commit-config.yaml
gadomski: Yeah, I'm a little wary of it getting in folks' way, but I'd also like to avoid another "do it sometimes" step a la `pytest --record-mode rewrite`. We can remove it later if we find that folks are running into problems.

jsignell (approved):

.pre-commit-config.yaml
jsignell: oh yeah I would have it be a "just do it when you bump up the min python version"

PR #1264, opened 2023-10-13
DRY asset methods

### LINKED ISSUE #1199: Add asset-modifying methods to Collection
`Collections` can have assets too, but methods such as `add_asset`, `make_asset_hrefs_absolute`, etc are only on `Item`. These should probably be moved to a mixin and included in both `Collection` and `Item`.

gadomski:
**Related Issue(s):**

- Closes #1199
- Refactor of #1168

**Description:**

I started working on #1199 only to discover it had already be fixed by #1168. However, I went an extra step and refactored the asset-modifying methods to an `Assets` protocol, so we only need to implement the logic once. The only hackey bit is the `get_self_href` stub, which is in place instead of trying to refactor all link-related logic to a `Links` protocol (which I tried but felt too disruptive).

**PR Checklist:**

### DIFF
--- a/pystac/asset.py
+++ b/pystac/asset.py
@@ -4,16 +4,15 @@ import os
 import shutil
 from copy import copy, deepcopy
 from html import escape
-from typing import TYPE_CHECKING, Any, TypeVar
+from typing import TYPE_CHECKING, Any, Protocol, TypeVar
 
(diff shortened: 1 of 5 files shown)


jsignell (approved):
It's beautiful :heart_eyes:

Issue #1247, opened 2023-09-28
Update file extension to v2.1.0


gadomski:
It's currently at v2.0.0. Changes:

```markdown
### Added

- `file:local_path` to indicate a relative local path for the asset file. [#7](https://github.com/stac-extensions/file/pull/7)
- Support for STAC Catalogs and link objects. [#9](https://github.com/stac-extensions/file/pull/9)
```

Issue #472, opened 2021-06-23
Add File Info Extension v1.0.0 -> v2.0.0 migration


duckontheweb:
Support for v2.0.0 of the File Info Extension was added in #442, but the `migrate` hook was not updated to include a migration from v1.0.0 to the new version. That hook should be updated to include the migration.

Based on my understanding of the changes between versions, the following fields should be migrated to [Raster Bands Object](https://github.com/stac-extensions/raster#raster-band-object) fields (dropping the `file:` prefix as well):

* `file:bits_per_sample`
* `file:data_type`
* `file:nodata`
* `file:unit`

@emmanuelmathot @matthewhanson @m-mohr Feel free to correct this if I got anything wrong.

emmanuelmathot:
I am not sure we can easily transfer those information that were describing the file to one or more raster bands. There are several cases that should be supported and thus several implementation rules must be defined. A basic rule could be to apply the file fields to all raster bands corresponding fields. If the assets of the original item does not describe any raster band, is a default raster band created?

duckontheweb:
> A basic rule could be to apply the file fields to all raster bands corresponding fields. 

I was thinking that if `raster:bands` is defined we would check if the field is already defined in *any* of the Raster Band Objects. If it is, we would take no action since we would not want to overwrite a value defined by the Raster Extension. If it is not defined, then we would populate that field with the `file:*` value.

> If the assets of the original item does not describe any raster band, is a default raster band created?

Yeah, this is a tricky. I was thinking that we should create a default Raster Band Object so that we don't lose the information. However, that could be misleading if an asset has more than 1 band but we create only a single Raster Band Object. It also doesn't seem possible to discover the number of Raster Band Objects that should be created without opening the file, which is not something we want to do here. What is your opinion on this @emmanuelmathot?

emmanuelmathot:
That is really tricky. We should avoid add raster bands to non raster assets. I would create a default raster band so that we do not loose the information . But I would add some kind of filters to be sure that the asset is a potential raster (e.g. mime-type, file extension)

duckontheweb:
> We should avoid add raster bands to non raster assets. I would create a default raster band so that we do not loose the information . But I would add some kind of filters to be sure that the asset is a potential raster (e.g. mime-type, file extension)

[...]

PR #1265, opened 2023-10-13
Update file extension and add `Link.ext`

jsignell:
**Related Issue(s):**

- #1247 

**Description:**

The update added a new field and let the ext apply to links and catalogs.
This is the first extension in pystac that can operate on a `Link` so I added the `Link.ext` syntax as well.

**PR Checklist:**


pystac/extensions/base.py
```diff
@@ -190,7 +190,7 @@ def validate_owner_has_extension(
     @classmethod
     def ensure_owner_has_extension(
         cls,
-        asset: pystac.Asset | AssetDefinition,
+        asset_or_link: pystac.Asset | AssetDefinition | pystac.Link,
```
gadomski, line 193: This is a breaking change, isn't it? Since folks might be passing-by-keyword?

Ok if it is, we just would need to defer this until v2 then?

jsignell: I can leave it as being called `asset`. No strong sentiment on that.

jsignell: yeah it's note in v1.8.4, so maybe it's ok?

tests/data-files/file/catalog.json
jsignell, line 1: I copied these examples over fresh from the extension repo.

pystac/extensions/file.py
```diff
@@ -220,23 +225,94 @@ def get_schema_uri(cls) -> str:
...
+        return f"<LinkFileExtension Link href={self.link_href}>"
+
+
 class FileExtensionHooks(ExtensionHooks):
     schema_uri: str = SCHEMA_URI
-    prev_extension_ids = {"file"}
-    stac_object_types = {pystac.STACObjectType.ITEM}
+    prev_extension_ids = {
+        "file",
+        "https://stac-extensions.github.io/file/v2.0.0/schema.json",
```
gadomski, line 309: Could we also migrate from v1.0.0 as long as those removed fields aren't present? https://github.com/stac-extensions/file/blob/main/CHANGELOG.md#removed-1

jsignell: sure! We can just drop them on the floor right?

gadomski: I _think_ they can stay in the `additional_fields`, they just won't be accessible via the extension class methods? My instinct is that its better to keep them around rather than silently throwing them away, but I'd be ok w/ either way I think.

Also found a separate issue for that migration, so we could kill that too: https://github.com/stac-utils/pystac/issues/472.

It shouldn't be too many users, v2 came out _right_ after STAC v1.0.0.

jsignell: no no no. Ignore that.

jsignell: yeah I think we should do at least do a warning when updating to the new version if they have those removed fields. Then we can leave them in additional_fields. I guess I'm trying to figure out what migrate should be responsible for.

gadomski: I think warning-and-leave is the best, it's both permissive and loud :-). 👍🏼

gadomski (approved):

Issue #1286, opened 2023-12-04
Support for projection v2.0.0 extension


jlaura:
When [the switch to `proj:code`](https://github.com/stac-extensions/projection/pull/12) lands in the projection extension and a 2.0 is released, it would be awesome to have PySTAC update for the new field. 

Believe that will mean supporting projection v1.1.0 that uses `proj:epsg` and then 2.0.0 that has the new `proj:code` extension.

Best, 
Jay

harryle95:
Hey guys, I am just wondering what's the progress on this. We are using this pystac extension for our project and found the same issue with code and epsg. I am happy to help with the PR if needed.

KeynesYouDigIt:
@gadomski / @jsignell Should I take a crack at this? looking for a nice "medium sized" issue to keep getting my feet wet :)

gadomski:
Yeah! @jsignell has a WIP branch here: https://github.com/stac-utils/pystac/pull/1287

KeynesYouDigIt:
@gadomski can you assign me for now? I'll start breaking it down tomorrow, will raise heck if I get stuck :smile:

gadomski:
You bet! Thanks for jumping in! FYI I'm off until 02 Jan so won't be doing any reviews until then.

PR #1287, opened 2023-12-04
ProjectionExtension v2 (proj:epsg -> proj:code)

jsignell:
**Related Issue(s):**

- #1286 

**Description:**

The intention of this PR is to still support using epsg and just store it in `proj:code`. 

> [!NOTE]
> This PR removes epsg as a positional argument to ProjectionExtension.apply. It also removes all positional arguments to it should raise nice errors for that scenario.
>
> Another option would be to allow positional arguments and automatically deduce whether they should be interpreted as `epsg` or `code` based on type. I suspect that using `epsg` as a positional argument in `apply` is rare, so I strongly prefer the approach in this PR.

TO DO:
 - [x] get summaries working properly (waiting on https://github.com/stac-utils/stac-fields/pull/27)
 - [x] remove local override of projection extension (waiting on https://github.com/stac-extensions/projection/pull/12)
 - [x] add explicit tests for old versions of the extension

**PR Checklist:**


nick-j-roberts:
Just checking in on this, is there a date by which the change to support proj:code instead of proj:epsg is expected to take effect?

KeynesYouDigIt:
@jsignell I noticed that to fix summaries, I am going to have to edit https://github.com/stac-utils/pystac/blob/main/pystac/static/fields-normalized.json - which is easy enough! but it's not formatted at all as a file, so I wanted to check: is it a build artifact of some other `dict` or somethinng? or is it just kinda new / should be edited directly if I need to make changes?

gadomski:
> or is it just kinda new / should be edited directly if I need to make changes?

`fields-normalized.json` is pulled from https://www.npmjs.com/package/@radiantearth/stac-fields via [this script](https://github.com/stac-utils/pystac/blob/main/scripts/pull-static).

KeynesYouDigIt:
Right I saw that, and it does make sense, but if I need to change it, do I
just Change it like I would than any other file in the repo?

gadomski:
> Right I saw that, and it does make sense, but if I need to change it, do I
> just Change it like I would than any other file in the repo?

I don't quite understand the question. Any modifications to `fields-normalized.json` should probably be made via PR on https://github.com/stac-utils/stac-fields, unless it's a PySTAC-specific modification.

KeynesYouDigIt:
Sorry I didn’t get to follow up - After I walked away from my computer, I
was wondering if that was what you meant. Perfect that answers my question,
I’ll probably start with a PR to that repo then! Thanks!

Vincent Buscarello

jsignell:
> Sorry I didn’t get to follow up - After I walked away from my computer, I was wondering if that was what you meant. That answers my question, I’ll probably start with a PR to that repo then! Thanks!

Hi @KeynesYouDigIt! Thanks for trying to pick this back up. I had opened a PR on stac-fields back when I was working on this, and it looks like that morphed into https://github.com/stac-utils/stac-fields/commit/ebd9c99a83915cd3ad2832905f4b7fbee09cae0a is there something more that you need? Or maybe it's just a matter of updating a dependency.

KeynesYouDigIt:
Hi @jsignell !

> Or maybe it's just a matter of updating a dependency.

Whoops, yeah thats it.
[duh](https://i.giphy.com/media/v1.Y2lkPTc5MGI3NjExb2lib3owb3N3a2RkcXV0ZGJmbTl0b2hieWh6NGltdzBuemZsc3FpMyZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/l0ExiYECygKyesazu/giphy.gif)

Shoulda seen that, thanks for the help! Will let you know how it goes, hoping to wrap it up soon.

KeynesYouDigIt:
@jsignell So I have commits on this branch that should cover the step

[ ] get summaries working properly 
https://github.com/KeynesYouDigIt/pystac/tree/refs/heads/proj-v2
(CI workflow) https://github.com/KeynesYouDigIt/pystac/actions/runs/12727145368

Let me know if I should PR to your branch or directly to this one, I have no preference.


As far as the next step, I'm hunting around for exactly what this means...
[ ] remove local override of projection extension

Could you point me to where this is done and what this means? Does this refer to some of the URI constants in `pystac/extensions/projection.py` or something else? (still getting my head around the extensions stuff)

pystac/validation/jsonschemas/projection.json
jsignell, line 1: > remove local override of projection extension

@KeynesYouDigIt I think I meant that we should delete this file once projection extensions v2 had been released (which it now has)

pystac/validation/local_validator.py
```diff
@@ -47,6 +47,11 @@ def get_local_schema_cache() -> dict[str, dict[str, Any]]:
                 "provider",
             )
         },
+        # FIXME: This is temporary, until https://github.com/stac-extensions/projection/pull/12
```
jsignell, line 50: @KeynesYouDigIt this is the spot where we can drop the ref

jsignell:
Ok! @KeynesYouDigIt thanks so much for pushing this forward! I pulled your commits into this branch and did a few tidy ups. It looks like things are passing now, but as you mentioned more unit tests/docs would be super useful!

pystac/extensions/projection.py
```diff
@@ -65,7 +68,9 @@ class ProjectionExtension(
 
     def apply(
         self,
-        epsg: int | None,
+        *,
+        epsg: int | None = None,
+        code: str | None = None,
```
jsignell, line 73: @gadomski if we want strictly no API changes we could take away the no positional args constraint, but I think we'd want to add it in pystac v2 to make the behavior very explicit.

gadomski: Yeah I think that's fine, might break a few folks but it's an easy fix on their part.

pystac/extensions/projection.py
```diff
@@ -96,7 +104,15 @@ def apply(
             transform : The affine transformation coefficients for
                 the default grid
         """
-        self.epsg = epsg
+        if epsg is not None and code is not None:
+            raise KeyError(
+                "Only one of the options ``code`` and ``epsg`` should be specified."
+            )
+        elif epsg:
+            self.epsg = epsg
```
gadomski, line 112: 🤔 down in the [setter](https://github.com/stac-utils/pystac/pull/1287/files#diff-5718f1829c91c51869d65fe57f84b3405efb887f0db416197ca48395576dd022R142) we set `code`, but here we set `epsg`. Is there a reason those are different?

gadomski (commented):
Oof, thanks for doing this — PySTAC extensions, huh 😓? Just one or two comments, otherwise looks good.

pystac/extensions/projection.py
jsignell: This is using the `epsg` setter. It's basically saving having to write `self.code = f"EPSG:{epsg}"` in two places.

gadomski (approved):

KeynesYouDigIt:
Great work everybody! CC: @nick-j-roberts

PR #1292, opened 2023-12-21
Adds user-agent to the request, if it doesn't exist in the headers.

Geosynopsis:
**Related Issue(s):**

- #

**Description:**
Some servers block the request without user-agent. This PR adds user-agent to headers if it doesn't already exist (through headers override).

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [Unreleased]
 
 ### Changed
-
+- Adds user-agent header to the request headers if it doen't exist ([#1292](https://github.com/stac-utils/pystac/pull/1292))
 - Made item pickles smaller by changing how nested links are stored([#1285](https://github.com/stac-utils/pystac/pull/1285))
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [Unreleased]
 
 ### Changed
-- Adds user-agent header to the request headers if it doen't exist ([#1292](https://github.com/stac-utils/pystac/pull/1292))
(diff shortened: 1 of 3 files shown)


gadomski (commented):
@Geosynopsis is this change necessary? Working from current **main**, it appears as though **urllib** is sending a user agent already.

Client code:

```python
from pystac.stac_io import DefaultStacIO

stac_io = DefaultStacIO()
stac_io.read_text("http://localhost:7822/catalog.json")
```

To test:

```shell
$ nc -l 7822 & python test_1292.py && fg
[2] 48657
GET /catalog.json HTTP/1.1
Accept-Encoding: identity
Host: localhost:7822
User-Agent: Python-urllib/3.11
Connection: close

-- >8 --
```

Note the `User-Agent: Python-urllib/3.11` in the request.

Geosynopsis:
@gadomski It seems to be the issue with the URL with `https` behind cloudflare. Could you give it a try for this `https://www.cloudflare.com/en-gb/`. You should get 403 Forbidden without the headers.

gadomski:
> It seems to be the issue with the URL with https behind cloudflare.

That shouldn't be a **pystac** issue then -- sounds like the `User-Agent` header might be getting stripped on the way through? It would be very surprising if **urllib** _didn't_ send `User-Agent` for `https` requests but did for `http` requests.

Geosynopsis:
> > It seems to be the issue with the URL with https behind cloudflare.
> 
> That shouldn't be a **pystac** issue then -- sounds like the `User-Agent` header might be getting stripped on the way through? It would be very surprising if **urllib** _didn't_ send `User-Agent` for `https` requests but did for `http` requests.

Yeah, you're right it shouldn't be **pystac** issue. I'll close the pr. Either the `User-Agent` is stripped or the Cloudflare blocks `Python-urllib` agent.

PR #1294, opened 2023-12-23
add API layout strategy

thomas-maschler:
**Related Issue(s):**

- https://github.com/stac-utils/pystac-client/issues/241

**Description:**
This PR implements the API Layout Strategy following the STAC API Specifications.
Unlike the Best Practice Layout Strategy, this strategy is not based on file paths. Instead, it uses HTTP endpoints. This requires refactoring of Catalog's `add_child`, `add_item`, and  `update_href` methods not to remove the last path element. Instead, I moved this step into the `HrefLayoutStrategy.get_href` method and made it conditional based on whether the path is a file or not.

Together with https://github.com/stac-utils/pystac/pull/1295, this PR is part of setting the foundation for enabling API transactions using `pystac-client`.

**PR Checklist:**

- [x] `pre-commit` hooks pass locally
- [x] Tests pass (run `scripts/test`)
- [x] Documentation has been updated to reflect changes, if applicable
- [x] This PR maintains or improves overall codebase code coverage.
- [x] Changes are added to the [CHANGELOG](https://github.com/stac-utils/pystac/blob/main/CHANGELOG.md). See [the docs](https://pystac.readthedocs.io/en/latest/contributing.html#changelog) for information about adding to the changelog.

PR #1295, opened 2023-12-23
Default layout strategy for catalogs

thomas-maschler:
**Related Issue(s):**

- https://github.com/stac-utils/pystac-client/issues/241

**Description:**

This PR enables setting a default layout strategy for a catalog. Users can either set the default layout strategy
when instantiating the catalog or defining a fall back strategy when subclassing Catalog.

Together with #1294, this PR is part of setting the foundation for enabling API transactions using `pystac-client`.

**PR Checklist:**

### DIFF
--- a/pystac/catalog.py
+++ b/pystac/catalog.py
@@ -128,6 +128,10 @@ class Catalog(STACObject):
             catalog's self link's HREF.
         catalog_type : Optional catalog type for this catalog. Must
             be one of the values in :class:`~pystac.CatalogType`.
+        strategy : The layout strategy to use for setting the
+            HREFs of the catalog child objections and items.
+            If not provided, it will default to the strategy of the root and fallback to
+            :class:`~pystac.layout.BestPracticesLayoutStrategy`.
     """
 
     catalog_type: CatalogType
--- a/pystac/catalog.py
+++ b/pystac/catalog.py
@@ -167,6 +167,9 @@ class Catalog(STACObject):
     a canonical format.
     """
 
+    _layout_strategy: HrefLayoutStrategy = BestPracticesLayoutStrategy()
+    """Fallback layout strategy"""
+
     def __init__(
         self,
         id: str,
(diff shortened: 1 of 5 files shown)


pystac/catalog.py
```diff
@@ -167,6 +171,9 @@ class Catalog(STACObject):
     a canonical format.
     """
 
+    _layout_strategy: HrefLayoutStrategy = BestPracticesLayoutStrategy()
```
jsignell, line 174: So the idea is that pystac will be able to override this in the `Client` class?

pystac/catalog.py
```diff
@@ -128,6 +128,10 @@ class Catalog(STACObject):
             catalog's self link's HREF.
         catalog_type : Optional catalog type for this catalog. Must
             be one of the values in :class:`~pystac.CatalogType`.
+        strategy : The layout strategy to use for setting the
```
jsignell, line 131: See comment above about choosing `strategy` vs `layout_strategy` for the kwarg.

jsignell (commented):
This looks really good too! The only issue is consolidating on an argument name.

pystac/catalog.py
thomas-maschler: Yes, correct. For example, the `pystac_client.Client` could set the fallback strategy to `APILayoutStrategy`

thomas-maschler:
Thanks for the review, I updated the argument name.

pystac/catalog.py
```diff
@@ -221,6 +231,19 @@ def is_relative(self) -> bool:
...
+    def _get_strategy(self, strategy: HrefLayoutStrategy | None) -> HrefLayoutStrategy:
+        if strategy is not None:
+            return strategy
+        elif self.strategy is not None:
+            return self.strategy
+        elif root := self.get_root():
+            if root.strategy is not None:
+                return root.strategy
+            else:
+                return root._fallback_strategy
```
jsignell, line 243: I think this rename is a lot clearer too :+1:

jsignell (approved):

PR #1320, opened 2024-03-18
add custom validator as input

fmigneault:
**Related Issue(s):**

- n/a

**Description:**

Currently, performing validation on STAC objects with extensions that are not directly part of `pystac`, or that the schema is not yet published at the expected URI defined by the extension, is not convenient or obvious. 

In order to call `<STACObject>.validate()`, one has to either duplicate the operations done in `pystac.validation.validate_dict` to use their own validator or set it globally with `RegisteredValidator.set_validator` before calling `<STACObject>.validate()`. This PR adds a `validator` input to all the relevant functions to directly provide the validator to use for validation, with a local reference. This makes it easier to manage the "active" validator at the moment the `validate` method gets called.

**PR Checklist:**

### DIFF
--- a/pystac/stac_object.py
+++ b/pystac/stac_object.py
@@ -71,6 +71,11 @@ class STACObject(ABC):
         implementation. For JSON Schema validation (default validator), this
         will be a list of schema URIs that were used during validation.
 
+        Args:
+            validator : A custom validator to use for validation of the object.
+                If omitted, the default validator from
+                :class:`~pystac.validation.RegisteredValidator`
+                will be used instead.
         Raises:
             STACValidationError
         """
--- a/pystac/stac_object.py
+++ b/pystac/stac_object.py
@@ -61,19 +61,22 @@ class STACObject(ABC):
         self.links = []
         self.stac_extensions = stac_extensions
 
-    def validate(self) -> list[Any]:
+    def validate(
+        self,
+        validator: "pystac.validation.stac_validator.STACValidator" = None,
+    ) -> list[Any]:
         """Validate this STACObject.
 
         Returns a list of validation results, which depends on the validation
-        implementation. For JSON Schema validation, this will be a list
-        of schema URIs that were used during validation.
+        implementation. For JSON Schema validation (default validator), this
+        will be a list of schema URIs that were used during validation.
 
         Raises:
             STACValidationError
         """
         import pystac.validation
 
-        return pystac.validation.validate(self)
+        return pystac.validation.validate(self, validator=validator)
 
     def add_link(self, link: Link) -> None:
         """Add a link to this object's set of links.
--- a/pystac/stac_object.py
+++ b/pystac/stac_object.py
@@ -63,7 +63,7 @@ class STACObject(ABC):
 
     def validate(
         self,
-        validator: "pystac.validation.stac_validator.STACValidator" = None,
+        validator: pystac.validation.stac_validator.STACValidator | None = None,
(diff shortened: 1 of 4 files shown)


pystac/stac_object.py
```diff
@@ -61,19 +61,22 @@ def __init__(self, stac_extensions: list[str]) -> None:
...
+        self,
+        validator: "pystac.validation.stac_validator.STACValidator" = None,
+    ) -> list[Any]:
         """Validate this STACObject.
 
         Returns a list of validation results, which depends on the validation
-        implementation. For JSON Schema validation, this will be a list
-        of schema URIs that were used during validation.
+        implementation. For JSON Schema validation (default validator), this
+        will be a list of schema URIs that were used during validation.
```
jsignell, line 72: Add an `Args:` list to this docstring like we do in the other method definitions.

pystac/validation/__init__.py
```diff
@@ -19,11 +19,15 @@
...
-def validate(stac_object: STACObject) -> list[Any]:
+def validate(
+    stac_object: STACObject,
+    validator: STACValidator | None = None,
+) -> list[Any]:
     """Validates a :class:`~pystac.STACObject`.
 
     Args:
         stac_object : The stac object to validate.
+        validator: The validator to employ.
```
jsignell, line 30: Expand this description. It should be clear that the validator will be used instead of rather than in addition to the default.

jsignell (commented):
This seems like a nice change! It would be great to have more information in the docstrings and some tests.

fmigneault:
@jsignell I will look into adding a test.

pystac/validation/__init__.py
```diff
@@ -67,6 +73,10 @@ def validate_dict(
         extensions : Extension IDs for this stac object. If not supplied,
             PySTAC's identification logic to identify the extensions.
         href : Optional HREF of the STAC object being validated.
+        validator : A custom validator to use for validation of the STAC dictionary.
```
jsignell, line 76: This description is great! Can you use it above in the `def validate` as well?

tests/validation/test_validate.py
```diff
@@ -183,6 +183,55 @@ def test_validates_geojson_with_tuple_coordinates(self) -> None:
         # Should not raise.
         item.validate()
 
+    @pytest.mark.vcr()
```
jsignell, line 186: nice test! If you need vcr then there is probably a cassette file that you need to check in as well. In that case CI will fail. If CI passes then I think you don't need to have `pytest.mark.vcr` here.

jsignell (commented):
Thanks for adding the test and docs!

tests/validation/test_validate.py
fmigneault: I am not too familiar with VCR. I assume I need the same references that are contained in `data-files/item/sample-item.json` ? I simply reused that item from another test case.

fmigneault:
@jsignell 
Can this be validated again?
I would like to integrate it to complete https://github.com/stac-extensions/ml-aoi/pull/8 
Also, would it be possible to get a release tag soon after so I can pin the minimum version?

tests/validation/test_validate.py
jsignell: I took a look and it seems like this decorator is good enough. Don't need a new cassette as well.

jsignell (approved):
This looks good to me! Thanks for your patience.

PR #1372, opened 2024-07-25
Add example of custom StacIO for Azure Blob storage

bmcandr:
**Description:**
Adds an example of a custom `StacIO` for reading/writing STAC objects from/to Azure Blob storage. I organized the existing S3 example and new Blob storage example into tabs to avoid cluttering up the docs:

Screenshot 2024-07-25 at 15 24 55

I implemented this custom `StacIO` and thought it might be a nice to include an example for another cloud provider in the docs, but feel free to reject if this is unwanted.

**PR Checklist:**

### DIFF
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -318,49 +318,114 @@ argument of most object-specific I/O methods. You can also use
 :meth:`pystac.StacIO.set_default` in your client's ``__init__.py`` file to make this
 sub-class the default :class:`pystac.StacIO` implementation throughout the library.
 
-For example, this code will allow
+For example, the following code examples will allow
 for reading from AWS's S3 cloud object storage using `boto3
-<https://boto3.amazonaws.com/v1/documentation/api/latest/index.html>`__:
+<https://boto3.amazonaws.com/v1/documentation/api/latest/index.html>`__
+or Azure Blob Storage using the `Azure SDK for Python
+<https://learn.microsoft.com/en-us/python/api/overview/azure/storage-blob-readme?view=azure-python>`__:
 
-.. code-block:: python
+.. tab-set::
+   .. tab-item:: AWS S3
 
-   from urllib.parse import urlparse
-   import boto3
-   from pystac import Link
-   from pystac.stac_io import DefaultStacIO, StacIO
-   from typing import Union, Any
+      .. code-block:: python
 
-   class CustomStacIO(DefaultStacIO):
-      def __init__(self):
-         self.s3 = boto3.resource("s3")
-         super().__init__()
+         from urllib.parse import urlparse
+         import boto3
+         from pystac import Link
+         from pystac.stac_io import DefaultStacIO, StacIO
+         from typing import Union, Any
 
-      def read_text(
-         self, source: Union[str, Link], *args: Any, **kwargs: Any
-      ) -> str:
-         parsed = urlparse(source)
-         if parsed.scheme == "s3":
-            bucket = parsed.netloc
-            key = parsed.path[1:]
+         class CustomStacIO(DefaultStacIO):
+            def __init__(self):
+               self.s3 = boto3.resource("s3")
+               super().__init__()
 
-            obj = self.s3.Object(bucket, key)
-            return obj.get()["Body"].read().decode("utf-8")
-         else:
-            return super().read_text(source, *args, **kwargs)
+            def read_text(
+               self, source: Union[str, Link], *args: Any, **kwargs: Any
+            ) -> str:
+               parsed = urlparse(source)
+               if parsed.scheme == "s3":
+                  bucket = parsed.netloc
+                  key = parsed.path[1:]
 
-      def write_text(
-         self, dest: Union[str, Link], txt: str, *args: Any, **kwargs: Any
-      ) -> None:
-         parsed = urlparse(dest)
-         if parsed.scheme == "s3":
-            bucket = parsed.netloc
-            key = parsed.path[1:]
-            self.s3.Object(bucket, key).put(Body=txt, ContentEncoding="utf-8")
-         else:
-            super().write_text(dest, txt, *args, **kwargs)
+                  obj = self.s3.Object(bucket, key)
+                  return obj.get()["Body"].read().decode("utf-8")
+               else:
+                  return super().read_text(source, *args, **kwargs)
 
-   StacIO.set_default(CustomStacIO)
+            def write_text(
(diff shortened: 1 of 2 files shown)


bmcandr:
Full example code snippet:

```python
import os
from pathlib import PurePosixPath
from typing import Any, Tuple, Union
from urllib.parse import urlparse

from azure.storage.blob import BlobClient, ContentSettings
from pystac import Link
from pystac.stac_io import DefaultStacIO, StacIO

class BlobStacIO(DefaultStacIO):
   """A custom StacIO class for reading and writing STAC objects
   from/to Azure Blob storage.
   """

   def _parse_blob_url(self, url: str) -> Tuple[str, str]:
      path = PurePosixPath(urlparse(url).path)
      container = path.parts[1]
      blob = "/".join(path.parts[2:])
      return container, blob

   def _get_blob_client(self, container: str, blob: str) -> BlobClient:
      return BlobClient.from_connection_string(
            os.environ["AZURE_STORAGE_CONNECTION_STRING"],
            container_name=container,
            blob_name=blob,
      )

   def read_text(self, source: Union[str, Link], *args: Any, **kwargs: Any) -> str:
      if isinstance(source, Link):
            source = source.href
      if source.startswith("https"):
            container, blob = self._parse_blob_url(source)
            blob_client = self._get_blob_client(container, blob)
            obj = blob_client.download_blob().readall().decode()
            return obj
      else:
            return super().read_text(source, *args, **kwargs)

   def write_text(
      self, dest: Union[str, Link], txt: str, *args: Any, **kwargs: Any
   ) -> None:
      """Write STAC Objects to Blob storage. Note: overwrites by default."""
      if isinstance(dest, Link):
            dest = dest.href
      if dest.startswith("https"):
            container, blob = self._parse_blob_url(dest)
            blob_client = self._get_blob_client(container, blob)
            blob_client.upload_blob(
               txt,
               overwrite=True,
               content_settings=ContentSettings(content_type="application/json"),
            )
      else:
            super().write_text(dest, txt, *args, **kwargs)

StacIO.set_default(BlobStacIO)
```

docs/concepts.rst
```diff
@@ -318,49 +318,114 @@ argument of most object-specific I/O methods. You can also use
...
+
+            def _parse_blob_url(self, url: str) -> Tuple[str, str]:
+               path = PurePosixPath(urlparse(url).path)
+               container = path.parts[1]
+               blob = "/".join(path.parts[2:])
+               return container, blob
+
+            def _get_blob_client(self, container: str, blob: str) -> BlobClient:
+               return BlobClient.from_connection_string(
+                     os.environ["AZURE_STORAGE_CONNECTION_STRING"],
```
gadomski, line 395: I'm not a huge fan of implicit dependencies on env vars in the examples — could we define this connection string as a class-level attribute that falls back to an env var?

docs/concepts.rst
```diff
@@ -318,49 +318,114 @@ argument of most object-specific I/O methods. You can also use
...
+                     container_name=container,
+                     blob_name=blob,
+               )
+
+            def read_text(self, source: Union[str, Link], *args: Any, **kwargs: Any) -> str:
+               if isinstance(source, Link):
+                     source = source.href
+               if source.startswith("https"):
```
gadomski, line 403: Is there any way to be more specific to allow us to use this IO for other non-Azure `https` operations as well? E.g. I remember that Planetary Computer blob storage hostnames all looked like `https://*.blob.core.windows.net` — is that true for all blob storage?

Over in AWS land we have the nice `s3` protocol for detecting which URLs we want to handle, trying to get something similar here.

docs/concepts.rst
```diff
@@ -318,49 +318,114 @@ argument of most object-specific I/O methods. You can also use
...
+            ) -> None:
+               """Write STAC Objects to Blob storage. Note: overwrites by default."""
+               if isinstance(dest, Link):
+                     dest = dest.href
+               if dest.startswith("https"):
+                     container, blob = self._parse_blob_url(dest)
+                     blob_client = self._get_blob_client(container, blob)
+                     blob_client.upload_blob(
+                        txt,
+                        overwrite=True,
```
gadomski, line 422: Same as the connection string comment, can we make this a class-level attribute to show folks how they might disable overwriting?

gadomski (changes requested):
Thanks for the contribution — I love the use of tabs, very clean. Couple code-level nits but in general makes sense to me.

docs/concepts.rst
bmcandr: Good call!

docs/concepts.rst
bmcandr: Done!

docs/concepts.rst
bmcandr: Blob-land [also supports the `abfs://` protocol](https://learn.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-introduction-abfs-uri). At one point I swear `az://` was also in the mix, but can't find any evidence of that now... Either way, I updated the example to support both the HTTP and `abfs://` style URIs.

docs/concepts.rst
bmcandr: There are a number of ways to instantiate the `BlobClient` depending on how you want to authenticate. I updated the example to support either the `conn_str` or `account_url` + `credential` if required. These are now class attributes and the example includes a section showing how to set them.

bmcandr:
Here's the snippet I used for testing that reads an Item I copied from Planetary Computer into our Blob storage:

```python
import pystac
from pystac.stac_io import StacIO
from typing import cast
from azure.core.credentials import AzureSasCredential

from io_utils.stac.blob_stacio import BlobStacIO

BlobStacIO.account_url = "https://impactobservatory.blob.core.windows.net"
BlobStacIO.credential = AzureSasCredential(
    "sp=r&st=2024-07-29T23:22:42Z&se=2024-09-30T07:22:42Z&spr=https&sv=2022-11-02&sr=b&sig=vSnlRevk6ZPl0GjyV%2Fe9KoJ8jh5B60gU%2B%2BwYMRMkQsE%3D"
)
BlobStacIO.overwrite = False

StacIO.set_default(BlobStacIO)


item_href = "https://impactobservatory.blob.core.windows.net/io-brendan-dev/S2B_MSIL2A_20240729T144729_R139_T20QRF_20240729T170459.json"
item: pystac.Item = cast(pystac.Item, pystac.read_file(item_href))

print(item.assets)
```

gadomski (approved):

gadomski:
Thanks for the changes, LGTM! Appreciate it!

Issue #1392, opened 2024-08-21
Consistently return `None` from setter methods


gadomski:
Throughout **pystac**, "setter" methods usually return `None`, e.g.:

- [def set_owner(self, obj: Assets) -> None:](https://github.com/stac-utils/pystac/blob/e326b41a86b308c49dce92f1a525fee05bb21928/pystac/asset.py#L86-L87)
- [def set_root(self, root: Catalog | None) -> None:](https://github.com/stac-utils/pystac/blob/e326b41a86b308c49dce92f1a525fee05bb21928/pystac/catalog.py#L213-L214)
- [def set_self_href(self, href: str | None) -> None:](https://github.com/stac-utils/pystac/blob/e326b41a86b308c49dce92f1a525fee05bb21928/pystac/stac_object.py#L292-L293)

However, there are two instances (that I can find) where setters return something:

- [def set_collection(self, collection: Collection | None) -> Item:](https://github.com/stac-utils/pystac/blob/e326b41a86b308c49dce92f1a525fee05bb21928/pystac/item.py#L260-L261)
- [def set_owner(self, owner: STACObject | None) -> Link:](https://github.com/stac-utils/pystac/blob/e326b41a86b308c49dce92f1a525fee05bb21928/pystac/link.py#L117-L118)

I think we should change those two methods to return `None` to make things consistent across the library. This can be part of a v2.0 since it is a breaking change.

PR #1393, opened 2024-08-21
Update docs to reflect that collection rel type can be for anything

### LINKED ISSUE #1379: Add `collection` as a relation type

gadomski:
**Related Issue(s):**

- Closes #1379 

**Description:**

We already have a [collection rel type](https://github.com/stac-utils/pystac/blob/28352e7c8932ee25f72c1eb74fd8516cd2a7ed73/pystac/rel_type.py#L18-L19), so this PR is just a small docs tweak to reflect that collection links don't have to be only for items.

It's not required, but it would be nice to move `{get|set}_collection` to `STACObject` from `Item`: https://github.com/stac-utils/pystac/blob/28352e7c8932ee25f72c1eb74fd8516cd2a7ed73/pystac/item.py#L260-L295

However, as documented in #1392, `set_collection` returns `self`, which makes it a bit more awkward to return from the `STACObject` base class. My instinct is to hold off for moving `{get|set}_collection`, and just create a tracking issue in the v2.0 milestone so that we can solve #1392 and the method move at the same time.

**PR Checklist:**

- [x] Create tracking issue for moving `{get|set}_collection` to `STACObject`: https://github.com/stac-utils/pystac/issues/1399


jsignell (approved):
lgtm

PR #1394, opened 2024-08-21
Switch to ruff-format from black

gadomski:
**Description:**

Since we already have ruff for linting, this gets one one fewer dep. Doesn't need a changelog.

Also includes a couple of tweaks to ape https://github.com/astral-sh/ruff-pre-commit/blob/main/.pre-commit-hooks.yaml more exactly but still use the system `ruff`.

**PR Checklist:**

### DIFF
--- a/pystac/serialization/common_properties.py
+++ b/pystac/serialization/common_properties.py
@@ -89,8 +89,9 @@ def merge_common_properties(
                 collection_props = collection["properties"]
         else:
             raise ValueError(
-                "{} is expected to be a Collection or "
-                "dict but is neither.".format(collection)
+                "{} is expected to be a Collection or " "dict but is neither.".format(
+                    collection
+                )
             )
 
         if collection_props is not None:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,26 +13,18 @@ repos:
     hooks:
       - id: ruff
         name: ruff
-        entry: ruff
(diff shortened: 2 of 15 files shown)


jsignell (approved):
I am fine with this in theory but it looks like the max line len changed and resulted in most of the file changes. Is there a way to hold it constant so this PR doesn't touch as much of the codebase?

gadomski:
> I am fine with this in theory but it looks like the max line len changed and resulted in most of the file changes.

🤔 did it? [This line](https://github.com/stac-utils/pystac/pull/1394/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R105) stayed the same at `88`, which is the [black default](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length). I _think_ most of the rearrangements are from how parenthesis, commas, and multi-line strings are handled.

pystac/serialization/common_properties.py
```diff
@@ -89,8 +89,9 @@ def merge_common_properties(
                 collection_props = collection["properties"]
         else:
             raise ValueError(
-                "{} is expected to be a Collection or "
-                "dict but is neither.".format(collection)
+                "{} is expected to be a Collection or " "dict but is neither.".format(
+                    collection
```
jsignell, line 93: It was this kind of change that made me think the line length had changed. But I think you are right that it it's just changing how they break.

jsignell (approved):

PR #1406, opened 2024-09-22
Fix recursion error when eo band name is missing

### LINKED ISSUE #1402: Calling `bands` property on a `pystac.extensions.eo.EOExtension` can result in RecursionError
The following code snippet leads to a `RecursionError` on pystac 1.10.1:
```python
collection = pystac.read_file('https://planetarycomputer.microsoft.com/api/stac/v1/collections/landsat-c2-l2')
item_assets = collection.ext.item_assets
item_asset = list(item_assets.values())[2]
item_asset.ext.eo.bands
```

When inspecting the STAC collection, it can be seen that the ´bands´ in the `eo` extension of `item_asset` doesn't have the required `name` field. Instead of a `RequiredPropertyMissing`, however, we get a `RecursionError`, seemingly because of the following:

The error message of `RequiredPropertyMissing` reads
```python
msg = msg or f"{repr(obj)} does not have required property {prop}
```
`Band.__repr__` is defined as
```python
def __repr__(self) -> str:
    return f"<Band name={self.name}>"
```
Which in its turn will again try to raise a `RequiredPropertyMissing` error, and so on.

gadomski:
**Related Issue(s):**

- Closes #1402 

**Description:**

Fixed simply by using `get` on the properties dict instead of the accessor.

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 
 - Make `get_all_collections` properly recursive ([#1361](https://github.com/stac-utils/pystac/pull/1361))
 - Set `Item::collection` to `None` when there is no collection ([#1400](https://github.com/stac-utils/pystac/pull/1400))
+- Recursion error when `name` not set on `eo:bands` ([#1406](https://github.com/stac-utils/pystac/pull/1406))
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


jsignell (approved):
LGTM :clap: :clap:

PR #1414, opened 2024-09-26
Release v1.11.0

### LINKED ISSUE #1326: Repair automated release process
The v1.10.0 release to PyPI [failed](https://github.com/stac-utils/pystac/actions/runs/8480760122). I _think_ that the token we're using (via Github Actions secrets) has expired, but I can't be sure b/c I don't have access to the **stac-utils** PyPI account. I'm going to bring up the topic at the next regular STAC community meeting and we can come up with a plan to future-proof our release/maintenance process there. I expect that whatever we come up with will apply to more than just **pystac** (e.g. **pystac-client** at least should use a similar process).

gadomski:
**Related Issue(s):**

- Closes #1326, I've set up the trusted published in PyPI and created the environment here in Github.


jsignell (approved):
Sweeeeettt 🚀

PR #1423, opened 2024-10-02
Fix item field ordering on `save_object`

### LINKED ISSUE #587: Order of Item fields
Hi all! 
When generating Items using `item.save_object( )`, I expected the order of fields in my JSON files to reflect the order as listed [here](https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#item-fields) and the [example file](https://github.com/radiantearth/stac-spec/blob/master/examples/simple-item.json). However, this is not the case (see below) and I was wondering if I'm just missing something or if this is not yet supported by pySTAC? 
In terms of machine-readability it doesn't matter of course, but I'm working on a data product where the metadata output should be consistent and logical (including order of main fields). 

Expected order: 
```
{
    "type": "Feature",    
    "stac_version": "1.0.0",
    "stac_extensions": [...],
    "id": "...",
    "geometry": {...},
    "bbox": [...],
    "properties": {...},
    "links": [...],
    "assets": {...}
}
```
Actual order:
```
{
    "type": "Feature",    
    "stac_version": "1.0.0",
    "id": "...",
    "properties": {...},
    "geometry": {...},
    "links": [...],
    "assets": {...},
    "bbox": [...],
    "stac_extensions": [...]
}
```

moradology:
**Related Issue(s):**

- Closes #587

**Description:**
This PR fixes item field ordering in accordance with the order defined [here](https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#item-fields). Additionally, I've fixed what appears to be a small configuration error in pre-commit.config.yaml such that ruff-format wasn't actually fixing ill-formatted files.

**PR Checklist:**

### DIFF
--- a/pystac/item.py
+++ b/pystac/item.py
@@ -365,9 +365,11 @@ class Item(STACObject, Assets):
         d: dict[str, Any] = {
             "type": "Feature",
             "stac_version": pystac.get_stac_version(),
+            "stac_extensions": self.stac_extensions if self.stac_extensions else [],
             "id": self.id,
-            "properties": self.properties,
             "geometry": self.geometry,
+            "bbox": self.bbox if self.bbox is not None else [],
+            "properties": self.properties,
             "links": [link.to_dict(transform_href=transform_hrefs) for link in links],
             "assets": assets,
         }
@@ -383,6 +385,10 @@ class Item(STACObject, Assets):
 
         for key in self.extra_fields:
             d[key] = self.extra_fields[key]
+
+        # This field is prohibited if there's no geometry
+        if not self.geometry:
+            d.pop("bbox")
 
         return d
 
--- a/pystac/item.py
+++ b/pystac/item.py
@@ -374,12 +374,6 @@ class Item(STACObject, Assets):
             "assets": assets,
         }
 
-        if self.bbox is not None:
-            d["bbox"] = self.bbox
-
-        if self.stac_extensions is not None:
-            d["stac_extensions"] = self.stac_extensions
-
         if self.collection_id:
             d["collection"] = self.collection_id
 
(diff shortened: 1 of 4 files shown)


moradology:
@jsignell I think this ought to be good to go

pystac/item.py
```diff
@@ -365,9 +365,11 @@ def to_dict(
         d: dict[str, Any] = {
             "type": "Feature",
             "stac_version": pystac.get_stac_version(),
+            "stac_extensions": self.stac_extensions if self.stac_extensions else [],
             "id": self.id,
-            "properties": self.properties,
             "geometry": self.geometry,
+            "bbox": self.bbox if self.bbox is not None else [],
+            "properties": self.properties,
```
gadomski, line 372: Need to remove the (now redundant) conditional assignments below: https://github.com/stac-utils/pystac/pull/1423/files#diff-c235f14269d2641c490a6ec16abb7ab96c1f7d4e2ca4514b0ade666b2094469eR377-R381

tests/test_item.py
```diff
@@ -108,6 +108,39 @@ def test_asset_absolute_href_no_item_self(self) -> None:
...
+                item_json = json.load(f)
+
+            # Assert the order matches the expected order
+            actual_order = list(item_json.keys())
+            self.assertEqual(
+                actual_order,
+                expected_order,
+                f"Order was {actual_order}, expected {expected_order}",
+            )
+
```
gadomski, line 143: We're only changing `to_dict`, so we could just serialize to a dict and check that? No need for a temp file.

gadomski (changes requested):

gadomski (approved):

PR #1472, opened 2024-11-05
Add `file://` to self links that don't have a protocol

### LINKED ISSUE #1347: Add `file://` protocol to `self` links that currently don't have a protocol
xref https://github.com/radiantearth/stac-spec/pull/1281, but it seems likely that at some point in the future `self` links will be required to have a protocol (i.e. absolute filesystem paths won't be enough). We should update to **pystac** to write `file://` absolute paths to future-proof ourselves and help ecosystem adoption.

jsignell:
**Related Issue(s):**

- closes #1347

**Description:**

Pystac already coerces self links to absolute, so now it adds `file://` as well. 

Notes:
- TODO:  There are still some tests that are failing. Mostly having to do with `normalize`.
- The json schema spec _does_ define how `"format": "iri"` should be handled, but the Python implementation does seem to validate it very well. For instance this passes:
```python
from jsonschema import validate

validate("any-string", {"type": "string", "format": "iri", "minLength": 1})
```
- Python in general seems rather bad at interpreting the `file://` protocol. Python 3.13 adds some support to `pathlib` in https://docs.python.org/3/library/pathlib.html#pathlib.Path.from_uri

**PR Checklist:**

- [ ] Pre-commit hooks and tests pass (run `scripts/test`)

### DIFF
--- a/pystac/asset.py
+++ b/pystac/asset.py
@@ -8,7 +8,12 @@ from typing import TYPE_CHECKING, Any, Protocol, TypeVar
 
 from pystac import MediaType, STACError, common_metadata, utils
 from pystac.html.jinja_env import get_jinja_env
-from pystac.utils import is_absolute_href, make_absolute_href, make_relative_href
+from pystac.utils import (
+    is_absolute_href,
+    make_absolute_href,
+    make_relative_href,
+    safe_urlparse,
+)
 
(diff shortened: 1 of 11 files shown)


KeynesYouDigIt:
I was gonna take this but I will take a shot at a review!

PR #1495, opened 2025-01-16
fix: use application/geo+json for item links

### LINKED ISSUE #1494: Link `item` type should be `application/geo+json`
Hi!

Pystac is not using [the **preferred** media type for link of rel `item` which should be `application/geo+json` rather than `application/json`](https://github.com/radiantearth/stac-spec/blob/master/commons/links.md#hierarchical-relations): https://github.com/stac-utils/pystac/blob/2261b55ffbac6b52e6f56434bfd5508455dc3ac6/pystac/link.py#L478
Is there a reason for that?

Thank you

gadomski:
**Related Issue(s):**

- Closes #1494

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 ## [Unreleased]
 
-
 ### Added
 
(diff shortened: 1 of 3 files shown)


jsignell (approved):
LGTM

PR #1497, opened 2025-01-20
get_root_link/get_child_links/get_item_links: Ensure correct media types

### LINKED ISSUE #1255: get_child_links: Ensure correct media type?
It looks like e.g `get_child_links` doesn't check the media type.

So if I have child links to STAC Catalogs/Collections and to HTML files (that render the child STAC entities), is it intentional that I also get the HTML files?

Generally, does pystac support hierarchical links (child, item, self, parent) with types that are not STAC media types? I mean support in a sense that it doesn't screw up or doesn't throw an error (e.g. by ignoring them).

### LINKED ISSUE #1325: Root link parsing is not filtering based on media types
I have a STAC Catalog that supports multiple media type implementations of the root link in the Catalog.  It appears that the stac_object.get_root_link call is not passing a media type to stac_object.get_single_link, leading to it trying to parse representations other than expected (eg the HTML version instead of json).

I think this issue is similar to what was discussed in https://github.com/stac-utils/pystac/issues/1255 

Thanks

jsignell:
**Related Issue(s):**

- closes #1255 
- closes #1325 

**Description:**

* Allow media_type on get_links and get_single_link to an iterable
* Use `application/json", "application/geo+json" or None as the media_type for `get_root_link`, `get_child_links` and `get_item_links`

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Top-level `item_assets` dict on `Collection`s ([#1476](https://github.com/stac-utils/pystac/pull/1476))
 - Render Extension ([#1465](https://github.com/stac-utils/pystac/pull/1465))
+- Filter by links by list of media_types
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


gadomski (changes requested):
Needs CHANGELOG as well as one DRY option.

gadomski (approved):

Issue #1503, opened 2025-01-24
Version 1.12.0 breaks ProjectionExtension.crs_string for items using projection extension <2


Kirill888:
```python
import pystac
from pystac.extensions.projection import ProjectionExtension
from pystac_client import Client

stac = Client.open("https://planetarycomputer.microsoft.com/api/stac/v1")

search = stac.search(
    bbox=[-120.0, 35.0, -119.0, 36.0],
    datetime="2023-01-01/2023-01-07",
    collections=["landsat-c2-l2"],
)

print(f"pystac.__version__ = {pystac.__version__}\n")
for it in search.items():
    _it = pystac.item.Item.from_file(it.self_href)
    props = "; ".join(
        f"{k}={v}" for k, v in it.properties.items() if k.startswith("proj:")
    )
    proj_ext, *_ = (ext for ext in it.stac_extensions if "/projection/" in ext)
    print(
        f"""{it.id}
  proj: {props}
  proj.ext: {proj_ext}
  crs_str : {ProjectionExtension.ext(it).crs_string}
  .from_file() item
  crs_str': {ProjectionExtension.ext(_it).crs_string}
"""
    )
```

produces following output:

```
pystac.__version__ = 1.12.0

LC09_L2SP_043035_20230107_02_T1
  proj: proj:epsg=32610; proj:shape=[7731, 7611]; proj:transform=[30.0, 0.0, 564885.0, 0.0, -30.0, 4106115.0]
  proj.ext: https://stac-extensions.github.io/projection/v1.0.0/schema.json
  crs_str : None
  .from_file() item
  crs_str': EPSG:32610

LC08_L2SP_041036_20230101_02_T1
  proj: proj:epsg=32611; proj:shape=[7891, 7761]; proj:transform=[30.0, 0.0, 262785.0, 0.0, -30.0, 3948615.0]
  proj.ext: https://stac-extensions.github.io/projection/v1.0.0/schema.json
  crs_str : None
  .from_file() item
  crs_str': EPSG:32611

LC08_L2SP_041035_20230101_02_T1
  proj: proj:epsg=32611; proj:shape=[7871, 7741]; proj:transform=[30.0, 0.0, 302085.0, 0.0, -30.0, 4106715.0]
  proj.ext: https://stac-extensions.github.io/projection/v1.0.0/schema.json
  crs_str : None
  .from_file() item
  crs_str': EPSG:32611
```

[...]

Issue #1504, opened 2025-01-24
Migrate by default


gadomski:
As described in https://github.com/stac-utils/pystac/issues/1503#issuecomment-2612469196, our `from_file` method migrates to the latest version by default, but `from_dict` does not. This can lead to confusing discrepancies in behavior depending on how STAC data gets into **pystac**. I think we should align on migrate-by-default, as we will continue to see more and more STAC (and extension) versions as time goes on.

This _could_ be as simple as change the `from_dict` default. This could lead to unexpected behavior for some users but IMO that'd be worth it, and not a breaking change. Thoughts?

jsignell:
Yeah I like migrate by default. People can stay on an old version of pystac if they want to be using an old API for interacting with their STAC

PR #1505, opened 2025-01-24
fix: fall back to epsg

gadomski:
**Related Issue(s):**

- Closes https://github.com/stac-utils/pystac/issues/1503
- Follow-on to discuss migrate-by-default here: https://github.com/stac-utils/pystac/issues/1504

**Description:**

If a STAC item is _not_ migrated to the latest version, it would be missing its `code` attribute. This means that `epsg` would be none, even if the incoming dicitonary had an epsg.

I'll want to release this as a bugfix when merged.

cc @Kirill888

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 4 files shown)


jsignell (approved):
Thanks @gadomski. This is a good fix.

PR #1509, opened 2025-01-27
fix: migrate by default

### LINKED ISSUE #1504: Migrate by default

gadomski:
**Related Issue(s):**

- Closes #1504 
- Motivated by #1503

**Description:**

With increased heterogeneity of extensions in the ecosystem, especially breaking changes like [proj v2](https://github.com/stac-extensions/projection/releases/tag/v2.0.0), there's more and more schema allowed in a valid STAC value. Handling these in **pystac** is ... tricky.

On read+load, **pystac** has the concept of "migration", where fields on STAC objects and in STAC extensions are moved around and updated to the latest versions. Migration is currently ON by default when using `from_file`: https://github.com/stac-utils/pystac/blob/21a0bbbf1299e6e3323fea556b39f4ccad54e1f3/pystac/stac_object.py#L633

However, it is OFF by default when using `from_dict`: https://github.com/stac-utils/pystac/blob/21a0bbbf1299e6e3323fea556b39f4ccad54e1f3/pystac/stac_object.py#L650

This inconsistent behavior is confusing to users (and, tbh, developers). We should align on one. As discussed in https://github.com/stac-utils/pystac/issues/1504, defaulting to `migrate=True` seems like the best idea for the ecosystem generally.

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Changed
+
+- `migrate=True` is now the default in `from_dict` ([#1509](https://github.com/stac-utils/pystac/pull/1509))
+
 ### Fixed
 
 - Fall back to `epsg` when `code` is not present in the Projection extension ([#1505](https://github.com/stac-utils/pystac/pull/1505))
--- a/pystac/catalog.py
+++ b/pystac/catalog.py
@@ -1224,7 +1224,7 @@ class Catalog(STACObject):
         d: dict[str, Any],
         href: str | None = None,
         root: Catalog | None = None,
-        migrate: bool = False,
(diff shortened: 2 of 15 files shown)


jsignell (approved):
Nice!

PR #1512, opened 2025-01-27
Make sure that VersionRange has VersionIDs rather than strings

### LINKED ISSUE #1506: STACVersionID assigned string value in migration
https://github.com/stac-utils/pystac/blob/main/pystac/serialization/identify.py#L268 
I believe    `version_range.set_to_single(stac_version)` 
should be `version_range.set_to_single(STACVersionID(stac_version))`

This issue results in `min_version` and `max_version` being assigned strings. Later during migrations it results in bad code paths.

Image

This only occurs when the item stac_version is < STACVersion.DEFAULT_STAC_VERSION, which is why it doesn't happen during `test_older_extension_version` and probably wasn't common scenario until the STAC 1.1.0 update. 

```
    item_as_dict["stac_version"] = "1.0.0"
    migrated_item = pystac.Item.from_dict(item_as_dict, migrate=True)
```
Does cause the issue.

jsignell:
**Related Issue(s):**

- Closes #1506 

**Description:**

The `min_version` and `max_version` should always be coerced to `STACVersionID` objects.

**PR Checklist:**

- [ ] Pre-commit hooks and tests pass (run `scripts/test`)

### DIFF
--- a/tests/serialization/test_identify.py
+++ b/tests/serialization/test_identify.py
@@ -118,3 +118,17 @@ class VersionTest(unittest.TestCase):
 
         version_range = STACVersionRange(min_version="0.6.0-rc1", max_version="0.9.0")
         self.assertTrue(version_range.contains("0.9.0"))
+
+    def test_version_range_set_to_single(self) -> None:
+        version_range = STACVersionRange()
+        version_range.set_min("1.0.0-beta.1")
+        version_range.set_to_single("1.0.0")
+
+        self.assertTrue(version_range.contains("1.0.0"))
+
+    def test_version_range_set_min_and_max_directly(self) -> None:
+        version_range = STACVersionRange()
+        version_range.min_version = "1.0.0-beta.1"  # type:ignore
+        version_range.max_version = "1.1.0"  # type:ignore
+
+        self.assertTrue(version_range.contains("1.0.0"))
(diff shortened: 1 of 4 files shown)


tests/serialization/test_identify.py
```diff
@@ -118,3 +118,17 @@ def test_version_range_ordering(self) -> None:
...
+        version_range = STACVersionRange()
+        version_range.set_min("1.0.0-beta.1")
+        version_range.set_to_single("1.0.0")
+
+        self.assertTrue(version_range.contains("1.0.0"))
+
+    def test_version_range_set_min_and_max_directly(self) -> None:
+        version_range = STACVersionRange()
+        version_range.min_version = "1.0.0-beta.1"  # type:ignore
+        version_range.max_version = "1.1.0"  # type:ignore
```
jsignell, line 132: Mypy hates this type of coersion apparently. https://github.com/python/mypy/issues/3004 although that got closed in the last few days so maybe we can delete these ignore statements someday?

gadomski: Yup, and since they're in tests I'm not too worried about adding them now.

gadomski (approved):
LGTM ... do you think this is serious enough to do an immediate bugfix release? My instinct is to wait and see if any other reports trickle in this week.

jsignell:
> do you think this is serious enough to do an immediate bugfix release? 

I don't have a good sense of the impact. But seems like waiting a few days should be totally fine

PR #1516, opened 2025-02-02
993 progress: convert test_version.py & test_item.py to pytest

teks:
**Related Issue(s):** #993

- Instead of trying to duplicate `unittest.TestCase.maxDiff = None` in pytest, I
decided to just take it out and leave it to the user to use `pytest -vv` if
they want.
- I mostly refrained from redesigning tests. My thinking is that the total #993 changeset is
already going to be enormous so don't make it worse by adding semantic density.
- unittest's `TestCase.assertCountEqual()` isn't conveniently available in
pytest. I didn't try to duplicate the exact semantics since simpler assertions
worked just as well, ie `assert set(some_object) == {"a", "b"}`
- Some classes are retained because they seem like a useful container to
organize tests. The tests inside them were converted to pytest though.
- There's an http metadata entry in the revised cassette, `X-Cache`, that was a MISS in
the original but is a HIT here. I assume that's harmless?
- I suggested I'd make "a few" PRs for #993. I think I underestimated. 😞 

**PR Checklist:**

- [ ] ~Documentation has been updated to reflect changes, if applicable~
- [x] This PR maintains ~or improves~ overall codebase code coverage.
- [ ] ~Changes are added to the [CHANGELOG](https://github.com/stac-utils/pystac/blob/main/CHANGELOG.md).~


tests/test_item.py
```diff
@@ -27,343 +26,335 @@
...
-    def get_example_item_dict(self) -> dict[str, Any]:
-        m = TestCases.get_path("data-files/item/sample-item.json")
-        with open(m) as f:
-            item_dict: dict[str, Any] = json.load(f)
-        return item_dict
+def get_example_item_dict() -> dict[str, Any]:
+    m = TestCases.get_path("data-files/item/sample-item.json")
+    with open(m) as f:
+        item_dict: dict[str, Any] = json.load(f)
+    return item_dict
```
gadomski, line 33: Would this be better as a `pytest.fixture`?

gadomski (commented):
Thanks for the change, in general looks good! just one question.

gadomski (approved):

PR #1521, opened 2025-02-07
feat: add stac-geoparquet reading and writing

### LINKED ISSUE #1455: Use stacrs to read+write stac-geoparquet
**stacrs** is a zero-dependency Python package that can read and write [stac-geoparquet](https://github.com/stac-utils/stac-geoparquet) files. We could add **stacrs** as an optional dependency here to enable **stac-geoparquet** for **pystac**.

The option of adding **stacrs** to **pystac** was discussed in this issue: https://github.com/stac-utils/stac-rs/issues/462

> [!NOTE]
> I am the author of **stacrs** so I may have some bias.

gadomski:
**Related Issue(s):**

- Closes #1455

**Description:**

- Add **stacrs** as an optional dependency and use it to read and write item collections
- I initially thought about embedding it into StacIO but that felt too impactful, since we're just doing ItemCollections for now
- It's a little awkward since **stacrs** is async, so we may want to refactor some stuff out to helper functions if we end up using it other places

**PR Checklist:**


gadomski:
Marking as draft while I dig into some PyPI size restrictions that are keeping me from releasing Windoze wheels that work.

pystac/item_collection.py
```diff
@@ -191,12 +191,19 @@ def from_dict(
         return cls(items=items, extra_fields=extra_fields)
 
     @classmethod
-    def from_file(cls: type[C], href: HREF, stac_io: pystac.StacIO | None = None) -> C:
+    def from_file(
+        cls: type[C],
+        href: HREF,
+        stac_io: pystac.StacIO | None = None,
+        is_geoparquet: bool | None = None,
```
jsignell, line 198: At first read I don't love the idea of adding another "how to read" flag. I feel like that is really what `stac_io` is meant for - this kind of messes up the existing patterns. Like how would you go about defining a different way to read from geoparquet?

On the other hand I guess adding `stacrs` here will make reading from geoparquet work in the pystac-client case as well even thought that one uses a different `stac_io`. So maybe this is a practical solution.

gadomski: Yeah, I don't love it either — it's the complexity of "STAC but in different formats" bleeding into our software implementation. I'll have a think.

PR #1559, opened 2025-05-14
feat: permissive extent deserialization

### LINKED ISSUE #1558: Relax Collection.from_dict
We've got an instance of too-strict deserialization in the wild: https://github.com/EOPF-Sample-Service/eopf-stac/issues/18

gadomski:
**Related Issue(s):**

- Closes #1558 

**Description:**

Relax collection extent serialization, as there's a surprising number of Collections out there in the real world without extents. Includes a sidecar relaxing of the type on the extent constructors.

**PR Checklist:**

### DIFF
--- a/pystac/collection.py
+++ b/pystac/collection.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import warnings
-from collections.abc import Iterable
+from collections.abc import Iterable, Sequence
 from copy import deepcopy
 from datetime import datetime, timezone
 from typing import (
@@ -71,7 +71,7 @@ class SpatialExtent:
 
     def __init__(
         self,
-        bboxes: Bboxes | list[float | int],
+        bboxes: Bboxes | Sequence[float | int],
         extra_fields: dict[str, Any] | None = None,
     ) -> None:
         if not isinstance(bboxes, list):
@@ -199,7 +199,7 @@ class TemporalExtent:
 
     def __init__(
         self,
-        intervals: TemporalIntervals | list[datetime | None],
+        intervals: TemporalIntervals | Sequence[datetime | None],
         extra_fields: dict[str, Any] | None = None,
     ):
         if not isinstance(intervals, list):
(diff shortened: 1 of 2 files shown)


tests/test_collection.py
```diff
@@ -820,3 +820,19 @@ def test_from_items_with_providers(sample_item_collection: ItemCollection) -> No
...
     assert provider and provider.name == "pystac"
+
+
+def test_from_dict_null_extent(collection: Collection) -> None:
+    # https://github.com/stac-utils/pystac/issues/1558
+    # https://github.com/EOPF-Sample-Service/eopf-stac/issues/18
+    d = collection.to_dict()
+    d["extent"] = None
+    with pytest.warns(UserWarning):
+        Collection.from_dict(d)
```
jsignell, line 831: Add an assertion to prove that the extent gets set to global, all time?
```suggestion
        c = Collection.from_dict(d)
        assert c.extent.spatial.to_dict()["bbox"] == [-90, -180, 90, 180]
        assert c.extent.temporal.to_dict()["intervals"] == [None, None]
```

tests/test_collection.py
```diff
@@ -820,3 +820,19 @@ def test_from_items_with_providers(sample_item_collection: ItemCollection) -> No
...
+
+
+def test_from_dict_missing_extent(collection: Collection) -> None:
+    d = collection.to_dict()
+    del d["extent"]
+    with pytest.warns(UserWarning):
+        Collection.from_dict(d)
```
jsignell, line 838: Same comment as above.

jsignell (approved):
This looks good! Just one suggestion.

Also this is a larger topic, but I was mildly persuaded at the STAC birds of a feather at CNG that if we are less permissive then it might make users complain to providers and encourage the production of higher quality STAC catalogs.

gadomski:
> I was mildly persuaded at the STAC birds of a feather at CNG that if we are less permissive then it might make users complain to providers and encourage the production of higher quality STAC catalogs.

I get it, but so often the people creating broken STAC are wayyy separated from the feedback system. To me, warning feels like a nice compromise -- if **pystac** breaks on bad STAC, I think folks are just as likely to just not use **pystac-client**, make HTTP calls directly to dictionaries, and muddle on.

jsignell:
> **pystac** breaks on bad STAC, I think folks are just as likely to just not use **pystac-client**, make HTTP calls directly to dictionaries, and muddle on.

Yeah that makes a ton of sense to me as well. I agree that is the more likely the path that people will see and once we lose them they are probably hard to get back.

pystac/collection.py
```diff
@@ -652,7 +652,17 @@ def from_dict(
...
-        extent = Extent.from_dict(d.pop("extent"))
+        if extent_dict := d.pop("extent", None):
+            extent = Extent.from_dict(extent_dict)
+        else:
+            warnings.warn(
+                "Collection is missing extent, setting default spatial and "
+                "temporal extents"
+            )
+            extent = Extent(
+                spatial=SpatialExtent([-90, -180, 90, 180]),
```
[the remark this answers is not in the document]
jsignell: ok sweet. I've got it in v2, but I can flip it in v1 as well.

PR #1582, opened 2025-09-22
Fix readthedocs build

jsignell:
Same as https://github.com/stac-utils/pystac-client/pull/830/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -92,6 +92,7 @@ instance/
 
 # Sphinx documentation
 docs/_build/
+docs/build/
 
 # PyBuilder
 .pybuilder/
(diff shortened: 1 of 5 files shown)


gadomski (dismissed):

jsignell:
Ok so maybe the issue with the docs build on this repo is different. It looks like the docs are building successfully on release and main. They just aren't getting the right version numbers applied. I am doing an experiment where I try to trigger a manual build of v1.14.1 and v1.13.0 via the readthedocs dashboard.

jsignell:
The last version that had been successfully built with its version info was version 1.0 which is not prefixed with a v and has only 2 positions. Those readthedocs were built off a branch apparently. Not sure if any of that matters or not. 

Anyways versions before 1.11.0 don't build because of changed requirements about what is included in .readthedocs.yml

PR #1584, opened 2025-10-05
Speed up importing pystac

pjonsson:
**Related Issue(s):**

- #

**Description:**

Python import times are noticeable during
interactive use of CLI applications, so
push the (near) single-use imports into
the functions where they are used so
the import cost is paid when calling
the functions instead of when starting
the CLI to print the help or do some
other unrelated action.

Timings on top of #1583 inside
a Python:3.10-bookworm container.

Before this change:

$ uv run hyperfine --warmup 3 "python3 -c 'import pystac'" Benchmark 1: python3 -c 'import pystac'
  Time (mean ± σ):      67.2 ms ±   1.6 ms    [User: 58.6 ms, System: 8.6 ms]
  Range (min … max):    62.5 ms …  69.9 ms    45 runs

After this change:

$ uv run hyperfine --warmup 3 "python3 -c 'import pystac'" Benchmark 1: python3 -c 'import pystac'
  Time (mean ± σ):      59.9 ms ±   1.5 ms    [User: 52.0 ms, System: 7.8 ms]
  Range (min … max):    56.9 ms …  64.2 ms    48 runs

**PR Checklist:**

- [X] Pre-commit hooks pass (run `pre-commit run --all-files`)
- [X] Tests pass (run `pytest`)

### DIFF
--- a/pystac/__init__.py
+++ b/pystac/__init__.py
@@ -44,7 +44,6 @@ __all__ = [
     "set_stac_version",
 ]
 
-import os
 import warnings
 from typing import Any
 
@@ -198,6 +197,8 @@ def write_file(
     """
     if stac_io is None:
         stac_io = StacIO.default()
+    import os
+
     dest_href = None if dest_href is None else str(os.fspath(dest_href))
     obj.save_object(
         include_self_link=include_self_link, dest_href=dest_href, stac_io=stac_io
--- a/pystac/asset.py
+++ b/pystac/asset.py
@@ -1,13 +1,10 @@
 from __future__ import annotations
 
-import os
 import shutil
 from copy import copy, deepcopy
-from html import escape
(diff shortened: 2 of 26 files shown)


gadomski (commented):
To me, it feels strange to move standard library imports down into functions. What happens if you just move `dateutil` and other external library imports?

gadomski (commented):
Thanks for this! I'd be fine with this change, but it looks like it's not correctly implemented (failing CI). It's a pretty minimal improvement (`56.8 ms ±   0.6 ms` down to `52.4 ms ±   0.7 ms` on my local system) but doesn't hurt anything, so 👍🏼 generally.

Thanks for the one-liner to do the benchmarking! It's been a while since I'd used **hyperfine**, it's so good for what it is.

PR #1592, opened 2025-10-21
feat: add cf extension

ahenrij:
**Description:** Add support for the [CF Extension](https://github.com/stac-extensions/cf)

**PR Checklist:**

- [X] Pre-commit hooks pass (run `pre-commit run --all-files`)
- [X] Tests pass (run `pytest`)


gadomski:
> Is there anything blocking that could be done on our end to help in its integration?

Nope, this is in my queue to review, I just haven't gotten to it yet.

pystac/extensions/cf.py
```diff
@@ -0,0 +1,218 @@
...
+
+    collection: pystac.Collection
+    properties: dict[str, Any]
+
+    def __init__(self, collection: pystac.Collection):
+        self.collection = collection
+        self.properties = collection.extra_fields
+
+    def __repr__(self) -> str:
+        return f"<CollectionCFExtension Item id={self.collection.id}>"
```
gadomski:
```suggestion
        return f"<CollectionCFExtension Item id={self.item.id}>"
```

gadomski (changes requested):
Missing a few things:

- Tests
- Needs to implement an instance of `ExtensionHooks`

There might be more, but adding those things will be a good next step.

gadomski (changes requested):
@ahenrij apologies for the rug-pull, but we've redone how extensions work in https://github.com/stac-utils/pystac/pull/1650. There's now a bit of additional work required to add an extension -- see the existing code for examples.

ahenrij:
@gadomski the failing test is from mypy flagging these asserts in [`tests/test_utils.py#L320-L331`](https://github.com/stac-utils/pystac/blob/main/tests/test_utils.py#L320-L331) as redundant:

```python
assert utc_datetime.tzinfo is tz.tzutc()
assert utc_datetime.tzinfo is not tz.tzlocal() # always true if previous assertion is true
```

Could you please confirm so I can remove the duplicate?

FYI @fmigneault

gadomski:
Yup, fixed in https://github.com/stac-utils/pystac/pull/1705

pystac/extensions/cf.py
ahenrij: @gadomski, just to make sure, we are in `CollectionCFExtension`, so shouldn't this be `collection.id`?

gadomski (commented):
Does this consider other parts of the `cf` extnesion, like `standard_name` and `cell_methods`?

fmigneault:
@gadomski 
This is somewhat confusing, but the https://github.com/stac-extensions/cf on main defines `cf:standard_name` and `cf:cell_method`, which is what users will see first. However, this does not have any release version. The code only defines `cf:parameter` as indicated by existing https://github.com/stac-extensions/cf/tree/v0.2.0 and https://github.com/stac-extensions/cf/tree/v0.1.0

gadomski:
@fmigneault I'd like to get clarity from the extension itself before we merge any code to **pystac**, if possible. Looks like it's just currently in folks' backlog: https://github.com/stac-extensions/cf/pull/8#issuecomment-4074569330.

fmigneault:
From what I can see, it is a major refactor that would need a v1.0.0 release. Versions v0.x rely on the CF parameters provided directly in the STAC Asset/Item/Collection. This pending release embeds them within `cube:variables`, which makes sense from a reusable metadata perspective, but requires a completely different JSON schema and an intermediate dependency on datacube extension. Also, its introduces the potential inclusion of CF within `bands`, which would require a STAC 1.1 check. I'm not sure how `pystac` should handle these breaking changes. They can almost be considered different extensions because nothing is portable beside the `cf` prefix.

gadomski:
> They can almost be considered different extensions because nothing is portable beside the cf prefix.

Understood. Since it was approved and merged with **main**, my instinct is to _wait_ to create a **pystac** extension until the new version is released, and then base off that.

FWIW the new extension layout (https://github.com/stac-utils/pystac/pull/1650) _should_ make it easier to host-and-use external implementations of extension, if you wanted to maintain a version of the **cf** extension outside of the **pystac** repo. Right now the new extension layout is only available on **main**, but I'm working on a v1.15.0 release to get it onto PyPI: https://github.com/stac-utils/pystac/issues/1725.

gadomski:
Marking as draft as it has merge conflicts

Issue #1605, opened 2025-12-16
pystac v2 extension class experiments


jsignell:
# Context

## What is useful about pystac extension classes?

For writing:
- easy extension value assignment on a STAC object
- apply the same extension values to many STAC objects
- on the fly validation of extension values

For updating:
- migrating extension values

For reading:
- intuitive exploration of extension values

## What is annoying about pystac extension classes?

- Writing a new class is fiddly and hard to get right, lots of boilerplate
- Updating to a new version is also fiddly

jsignell:
# Experiments with Pydantic

I did an experiment where I autogenerated Pydantic classes for the eo extension from the jsonschema:

```
datamodel-codegen  --url https://stac-extensions.github.io/eo/v1.1.0/schema.json --input-file-type jsonschema --output ./eo.py --output-model-type "pydantic_v2.BaseModel"
```

That generated a few different classes that inherit from each other. I think the most useful will be the top level `Fields` class which contains the nested fields that are described by the extension and can be included in various spots in the STAC hierarchy: top level in collection and asset, within properties in item. 

## Experiment 1: Inherit from Pydantic Fields

My first experiment was to have the `EOExtension` class inherit from the `Fields` class as well as from the base `Extension` class. This means that the `EOExtension` class is _itself_ a Pydantic model - which gives you free validation and json serialization. It also does a nice job with type hints and tab completion.

```python
from src import pystac
from src.pystac.extensions.eo import EOExtension


item = pystac.Item.from_file("tests/data-files/eo/eo-landsat-example.json")
asset = item.assets["B1"]
eo = EOExtension(asset)
```

To get this working, on the `Extension` base class I changed the `__init__` so that it grabs the fields from the STAC object and feeds them into Pydantic:

```python
    def __init__(self, extendable: Extendable, *args) -> None:
        super().__init__(*args, **extendable.get_fields())
        self._extendable: Extendable = extendable
```

Note that private attributes are ignored by pydantic. 

Now the tricky part of this is setting up the bidirectional link between the STAC object fields and the extension attributes. For the top level fields you can get them saving back to the STAC object fields by doing something like the following on the `EOExtension` class:

```python
    model_config = ConfigDict(validate_assignment=True)

    @model_validator(mode='after')
    def _set_on_extendable(self):
        if hasattr(self, "_extendable"):
            fields = self._extendable.get_fields()
            new_fields = self.model_dump(by_alias=True, exclude_unset=False)
            fields.update(new_fields)
[...]

PR #1623, opened 2026-01-06
BREAKING CHANGE: V2 STAC Extensions

jsignell:
**Related Issue(s):**

- #1605 

**Description:**

Lays out a vision for STAC Extension classes in pystac v2. The goal is to improve the user experience of working with pystac objects that include stac extension fields while making it dramatically easier to write and maintain the stac extension classes.

With this PR you would start with a pydantic model which can be autogenerated from the jsonschema and then tweaked a bit or written by hand. Then you just decorate the model with a special method called `register_extension` and add an import to `extensions/__init__.py`. 

Once you have done those steps the user can access the fields (and there is type hinting) like: 

```python
from src import pystac


item = pystac.Item.from_file("tests/data-files/eo/eo-landsat-example.json")
item.ext.eo.cloud_cover
```

There are also some helper functions on `item.ext.eo` for instance to get the related extension url from `stac_extensions`:

```python
item.ext.eo.get_url()
```

or to remove the given extension from `stac_extensions`:

```python
item.ext.eo.remove()
```

To add a bunch of new extension fields the idea is that you would do something like:

```python
from src.pystac.extensions import EOFields

blob = {
    "eo:bands": [
        {
          "name": "B1",
          "common_name": "coastal",
          "center_wavelength": 0.44,
          "full_width_half_max": 0.02,
          "solar_illumination": 2000
        }
    ]
}

f = EOFields(**blob)
item.ext.eo.apply(f)
```

**Ongoing work:**
- [ ] Figure out typing on the `register_extension` mechanism
- [ ] Is there a way to expand this so you can have different versions of extension and aren't forced to migrate?
- [ ] How to exclude certain types of objects from using certain extensions?
- [ ] Make it easier to unset a particular field.
- [ ] Include removing all extension fields within the `remove` method?

**PR Checklist:**


gadomski (commented):
I love (love love love) being able to auto-gen extensions from the JSON schema. This is a big enough win to justify lots of other pain, IMO.

I have a thought, not fully formed, that we could break extensions out of the **pystac** Python package entirely for v2. I'm thinking about namespace package plugins (probably in `pystac.extensions`): https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/#using-namespace-packages. We could maintain "core" extensions in the **pystac** repo but in their own Python packages (e.g. `pystac-projection` or `pystac-extension-projection`). This would allow us to version the plugins separately (maybe even keep extension code versions matching their latest-supported jsonschema verision?). This would make it easier for folks to contribute extensions ... they could just publish their own Python package and it would be auto-detected by our plugin mechanism, if present.

Issue #1633, opened 2026-01-09
User-Agent request header makes release PRs unnecessary verbose


thomas-maschler:
Request headers are recorded in VCR cassettes for all tests and request headers including the User-Agent key change with every new release. 
This makes release PRs unnecessary verbose, changing nearly 300 cassettes files.

PR #1634, opened 2026-01-09
test: Scrub request headers

thomas-maschler:
**Related Issue(s):**

- #1633

**Description:**

Request headers are recorded in VCR cassettes for all tests and request headers including the User-Agent key change with every new release.
This makes release PRs unnecessary verbose, changing nearly 300 cassettes files.

This PR scrubs the User-Agent request header before recording. This will keep future release PRs more light weight and the git history slim. 

**PR Checklist:**

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,11 +2,11 @@
 
 import json
 import shutil
-from urllib.request import Request
 import uuid
 from datetime import datetime
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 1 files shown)


jsignell (approved):
Nice! The response headers were so huge until #1627 that we never noticed the size of the request headers :)

I did notice that #1627 uses `test:` rather than `chore:` probably would be good to use `test:` here as well just to match.

PR #1636, opened 2026-01-13
feat: V2 common metadata

jsignell:
**Related Issue(s):**

- test/v1/test_common_metadata.py passing

**Description:**

Start on common metadata fields. I am not sure if these should be on all objects or just the ones they are most useful on.

The main benefit is the code completion powered by the static typing on the properties. Right now there is no validation on these setters, but there could be custom validation if needed.

An alternative to this would be to use a pydantic model but that has a few downsides:
- harder to use with this mixin class inheritance approach
- harder to use the `extra_fields` dict as the backing for these common fields
- adds a dependency on pydantic

**PR Checklist:**

### DIFF
--- a/src/pystac/item.py
+++ b/src/pystac/item.py
@@ -11,7 +11,14 @@ from pystac.rel_type import RelType
 
 from .asset import Asset, Assets
 from .band import Band
-from .common_metadata import CommonMetadata, DateTime
+from .common_metadata import (
+    Basics,
+    CommonMetadata,
+    DateTime,
+    Instrument,
+    Licensing,
+    Providers,
+)
 from .constants import DEFAULT_STAC_VERSION, STAC_OBJECT_TYPE
 from .container import Container
 from .geo_interface import GeoInterface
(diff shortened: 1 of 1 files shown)


src/pystac/asset.py
```diff
@@ -103,8 +101,7 @@ def get_absolute_href(self) -> str | None:
     @override
     def to_dict(self) -> dict[str, Any]:
         data = super().to_dict()
-        data["href"] = self.href
-        return data
+        return {"href": self.href, **data}
```
jsignell: Put the href first in the dict

src/pystac/asset.py
```diff
@@ -46,15 +47,12 @@ def clone[T: ItemAsset](self: T) -> T:
...
-            data["type"] = self.type
-        if self.roles is not None:
-            data["roles"] = self.roles
-        return data
+        data["title"] = self.title
+        data["description"] = self.description
+        data["type"] = self.type
+        data["roles"] = self.roles
+        data["statistics"] = self.statistics.to_dict() or None
+        return {k: v for k, v in data.items() if v is not None}
```
jsignell: I think the default behavior should be no null values in the output dict.

PR #1646, opened 2026-02-04
feat: V2 asset

jsignell:
Getting `test_asset.py` passing and `test_item_assets` passing with the old `ItemAssetDefinition`. Adds a new `test_item_assets.py` for the new `ItemAsset`. I think we should deprecate `ItemAssetDefiniton` it is unlikely people are really using it since it only came in a year or so ago to replace the old `item_assets` extension.

### DIFF
--- a/src/pystac/__init__.py
+++ b/src/pystac/__init__.py
@@ -2,7 +2,7 @@ from typing import Any
 
 from typing_extensions import deprecated
 
-from .asset import Asset
+from .asset import Asset, ItemAsset
 from .band import Band
 from .catalog import Catalog
 from .collection import (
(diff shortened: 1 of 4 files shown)


src/pystac/asset.py
```diff
@@ -28,6 +35,14 @@ def __init__(
         self.roles: list[str] | None = roles
         self.extra_fields: dict[str, Any] = kwargs
 
+    @override
+    def __eq__(self, other: Any) -> bool:
+        return self.to_dict() == other.to_dict()
+
+    def create(self):
+        # TODO deprecated version of __init__
```
gadomski: can we clean this TODO?

gadomski (changes requested):

jsignell:
Sorry I should have marked this draft again. I am going to update this to remove weird methods like `copy`, `move` and `delete`

PR #1650, opened 2026-02-12
refactor: move extensions to their own packages

gadomski:
This is a proposed refactor of each extension into its own Python package. Basic idea goes like this:

- Change the main Python package to be named `pystac-core`, but keep the code and import paths the same
- Move each extension to its own Python package
- Create a "meta" package named `pystac` that depends on everything
- Create a custom **pytest** plugin so we can share config among all the packages
- Relax our coverage expectation to 70%, and only run coverage on the core code (not extensions)

This _should_ keep the install story and API identical, while simultaneously de-coupling our extension versions from our main PySTAC version.

Opening as a draft to get a smell check on the idea.

> [!NOTE]
> I've disabled the benchmark CI for the time being to get things green, as `asv` is cranky with the new setup.

BEGIN_COMMIT_OVERRIDE
feat: move extensions to their own packages

END_COMMIT_OVERRIDE


jsignell (approved):
Whoa. I love it 😍 

The only part that is  hitting me a little weird is that the extension logic is nested under pystac/extensions. Aside from that looking a little odd I am also wondering if there is a risk of a conflicting file existing in pystac core.

extensions/projection/pyproject.toml
```diff
@@ -0,0 +1,36 @@
+[project]
+name = "pystac-ext-projection"
+description = "Projection extension for PySTAC"
+readme = "README.md"
+version = "2.0.0"
```
jsignell: Wait @gadomski what happens if we need to tweak the implementation of this extension? Is it wise to have the version be _exactly_ the extension schema version?

jsignell (approved):
This looks great! Just a few little comments

Issue #1657, opened 2026-03-16
Pystac v2: Get tests/v1 passing


jsignell:
I was just chatting with @ircwaves and he is down to help out with getting v1 tests passing on v2, so I figured we might need somewhere to coordinate work:

- [x] ├── html
      │   └── test_html.py -> #1659 
- [x] ├── posix_paths
      │   └── test_posix_paths.py -> #1658 
      ├── serialization
- [ ] │   ├── test_identify.py
- [ ] │   └── test_migrate.py
- [x] ├── test_asset.py  -> #1646 
- [ ] ├── test_cache.py
- [ ] ├── test_catalog.py -> @ircwaves working on this
- [x] ├── test_collection.py  -> #1668
- [x] ├── test_common_metadata.py
- [x] ├── test_item_assets.py  -> #1646 
- [ ] ├── test_item_collection.py -> @jsignell working on this
- [x] ├── test_item.py
- [ ] ├── test_layout.py
- [ ] ├── test_link.py -> @ircwaves working https://github.com/stac-utils/pystac/pull/1693
- [ ] ├── test_pystac_client.py
- [ ] ├── test_stac_io.py
- [ ] ├── test_summaries.py -> **this one might be worth implementing**
- [ ] ├── test_utils.py
- [ ] ├── test_version.py
- [ ] ├── test_writing.py
- [ ] ├── utils
      │   └── test_cases.py
      └── validation
- [ ]     ├── test_schema_uri_map.py
- [ ]     └── test_validate.py

PR #1658, opened 2026-03-16
test: enable posix path tests

ircwaves:
a trivial one to get started on the #1657 effort


jsignell (approved):
nice one! I guess in this case we can delay the decision about: "Is `link.get_target_str` a method that we need to deprecate? Or is it weird and unused and we can just delete it in which case we can xfail tests for it."

PR #1660, opened 2026-03-20
fix: asset `__eq__`

jsignell:
Fix asset `__eq__` as recommended in https://github.com/stac-utils/pystac/pull/1646#discussion_r2965623293

### DIFF
--- a/src/pystac/asset.py
+++ b/src/pystac/asset.py
@@ -43,6 +43,8 @@ class ItemAsset(DataValue, Instrument):
 
     @override
     def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, ItemAsset):
+            return NotImplemented
         return self.to_dict() == other.to_dict()
 
     @classmethod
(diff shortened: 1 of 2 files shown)


src/pystac/asset.py
```diff
@@ -43,6 +43,8 @@ def __init__(
 
     @override
     def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, ItemAsset):
+            return NotImplemented
```
pjonsson: Why not `return isinstance(other, ItemAsset) and self.to_dict() == other.to_dict()` so the user gets false when comparing `asset == y` for y's that are different from the asset?

gadomski: It's a good point to think about what Python developers might expect. I work in strongly-typed environments more often, which give a build-time error (instead of a runtime exception) when you attempt equality between two things that don't have it explicitly or implicitly implemented.

I don't really have a strong opinion, so maybe I'll throw to @jsignell for thoughts?

PR #1662, opened 2026-03-25
refactor: update v2 links for v1 test compatiliby

ircwaves:
continued #1657 efforts

Need to look into back-porting transform_href code temporarily.

Also still considering what to do with Link.get_absolute_href.


ircwaves:
wrong source branch

PR #1668, opened 2026-04-01
feat: V2 collection tests

jsignell:
**Related Issue(s):**

- #1657 

**Description:**

The big question that came up in this work was: do we want datetimes to be `dt.datetime` or iso strings? So far v2 has the  `DateTime` protocol defined in common_metadata.py has datetimes stored as strings in `extra_fields` but returns them as `dt.datetime` when accessed as a property. This is what is used by `item.properties`. But `SpatialExtent` just uses strings

```python

collection.extent.temporal.interval: list[list[str]] 
```

### DIFF
--- a/src/pystac/container.py
+++ b/src/pystac/container.py
@@ -169,11 +169,12 @@ class Container(STACObject, ABC):
     def get_child(
         self, id: str, recursive: bool = False, sort_links_by_id: bool = True
     ) -> Container | None:
-        for child in self.get_children(
-            id, recursive=recursive, sort_links_by_id=sort_links_by_id
-        ):
-            if child.id == id:
-                return child
+        return next(
+            self.get_children(
+                id, recursive=recursive, sort_links_by_id=sort_links_by_id
+            ),
+            None,
+        )
 
     def get_children(
         self, *ids: str, recursive: bool = False, sort_links_by_id: bool = True


gadomski:
> do we want datetimes to be dt.datetime or iso strings? 

My gut tells me to store and retrieve as datetimes, and only go to/from strings on (de)serialization. That implies that it's awkward to intentionally store as a string in `extra_fields`.

Very open to different models, though, so curious what ya'lls instincts are.

jsignell:
> > do we want datetimes to be dt.datetime or iso strings?
> 
> My gut tells me to store and retrieve as datetimes, and only go to/from strings on (de)serialization. That implies that it's awkward to intentionally store as a string in `extra_fields`.
> 
> Very open to different models, though, so curious what ya'lls instincts are.

I think your instincts are good. I updated this PR make all the regular datetime fields convert to `dt.datetime` on initialization and when setting. I didn't do this for `collection.extent.temporal.interval` yet because it looked like there was some intentional logic there. Also because of rounding you can't guarantee that roundtripping from str -> dt.datetime -> str will give exactly the same string. Probably not super important, but if we are trying to not mutate on read...

src/pystac/container.py
```diff
@@ -145,19 +149,53 @@ def remove_item(self, id: str) -> Item | None:
...
+        yield from self.get_collections(recursive=True)
+
+    def get_child(
+        self, id: str, recursive: bool = False, sort_links_by_id: bool = True
+    ) -> Container | None:
+        for child in self.get_children(
+            id, recursive=recursive, sort_links_by_id=sort_links_by_id
+        ):
             if child.id == id:
                 return child
```
ircwaves: How about this instead:

```suggestion
        return next(
            self.get_children(
                id, recursive=recursive, sort_links_by_id=sort_links_by_id
            ),
            None,
        )
```

tests/v1/test_common_metadata.py
```diff
@@ -43,6 +43,7 @@ def test_datetimes(date_time_range_item: Item) -> None:
     assert cm.providers is None
 
 
+@pytest.mark.xfail(reason="pystac v2 purposefully converts datetimes to datetime objects")
```
ircwaves, line 46: I assume that these are to be updated to work with the datetime object instead of string.   Correct?

src/pystac/container.py
jsignell: Yeah I like that!

tests/v1/test_common_metadata.py
jsignell: I was trying to us xfail to highlight places where the behavior is changing compared to v1 so my plan is actually to leave these as is.

jsignell:
@gadomski do you want to take a look at this one? I think it would be best to merge it before working on item_collection and catalog tests.

PR #1693, opened 2026-04-06
feat: V2 links updates

ircwaves:
**Related Issue(s):**

- #1657 
- adds `Container.get_item`
- adds `Link.clone` via `deepcopy`
- adds `Link.target` setter with deprecation notice
- adds `Link.get_absolute_href`, as it was needed for some v1 tests
- restores `Link.resolve_stac_object`
- restores convenience methods for common link types (`child, collection, item, derived_from`)

**Description:**

**PR Checklist:**

### DIFF
--- a/src/pystac/link.py
+++ b/src/pystac/link.py
@@ -159,6 +159,20 @@ class Link:
             if href := value.get_self_href():
                 self.set_href(href)
 
+    @deprecated("use .get_href()")
+    def get_target_str(self) -> str | None:
+        """Returns this link's target as a string.
+
+        If a string href was provided, returns that. If not, tries to resolve
+        the self link of the target object.
+        """
+        return self.get_href()
+
+    @deprecated("use bool(link.get_href())")
+    def has_target_href(self) -> bool:
+        """Returns true if this link has a string href in its target information."""
+        return bool(self.get_href())
+
     def get_target(self, start_href: str | None, reader: Reader) -> STACObject:
         from .stac_object import STACObject
 
--- a/src/pystac/link.py
+++ b/src/pystac/link.py
@@ -257,6 +257,16 @@ class Link:
 
     ##### Convenience methods for Link creation #####
     @classmethod
+    def root(cls: type[Self], c: Catalog) -> Self:
+        """Creates a link to a root Catalog or Collection."""
+        return cls(RelType.ROOT, c, media_type=MediaType.JSON)
+
+    @classmethod
+    def parent(cls: type[Self], c: Catalog, title: str | None = None) -> Self:
+        """Creates a link to a parent Catalog or Collection."""
+        return cls(RelType.PARENT, c, title=title, media_type=MediaType.JSON)
+
+    @classmethod
     def collection(cls: type[Self], c: Collection) -> Self:
         """Creates a link to a Collection."""
         return cls(RelType.COLLECTION, c, media_type=MediaType.JSON)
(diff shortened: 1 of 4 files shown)


src/pystac/container.py
```diff
@@ -44,6 +44,9 @@ def get_items(self, *ids: str, recursive: bool = False) -> Iterator[Item]:
                 if isinstance(stac_object, Container):
                     yield from stac_object.get_items(*ids, recursive=True)
 
+    def get_item(self, id: str, recursive: bool = False) -> Item | None:
```
jsignell: I think we've already got this one at line 31 :)

src/pystac/link.py
```diff
@@ -133,6 +145,20 @@ def href(self) -> str | None:
...
 
+    @target.setter
+    @deprecated("target is deprecated, either use .set_href() or .set_target()")
+    def target(self, value: str | STACObject) -> None:
+        if isinstance(value, str):
+            warnings.warn(
+                "Setting Link.target to href is no longer supported pystac v2.  "
+                "Assigning value to href instead."
+            )
+            self.set_href(value)
```
jsignell, line 163: Ooh I'd just been using `@deprecated`, TIL that that doesn't raise a warning on its own, it's just a typing thing.

tests/v1/test_link.py
```diff
@@ -51,26 +54,22 @@ def test_minimal(item: pystac.Item) -> None:
...
-    assert item == link.owner
-
 
 def test_relative() -> None:
     rel = "my rel"
     target = "../elsewhere"
     mime_type = "example/stac_thing"
-    link = pystac.Link(rel, target, mime_type, "a title", extra_fields={"a": "b"})
+    link = pystac.Link(
+        rel, target, media_type=mime_type, title="a title", extra_fields={"a": "b"}
```
jsignell, line 63: I think we should look for `extra_fields` in `init` `kwargs` and if they are in there, raise a warning, but use them to update `kwargs`. It would be a surprising change to have `extra_fields` show up in the `dict`.

src/pystac/container.py
ircwaves: Thanks for catching this.

tests/v1/test_link.py
```diff
@@ -232,38 +256,55 @@ def test_static_from_dict_round_trip() -> None:
...
-        with pytest.raises(KeyError):
-            pystac.Link.from_dict(d)
+        with pytest.raises((TypeError, ValueError)):
+            Link.from_dict(d)
 
 
-def test_static_collection(collection: pystac.Collection) -> None:
-    link = pystac.Link.collection(collection)
-    expected = {"rel": "collection", "href": None, "type": "application/json"}
+def test_static_collection(self_link_collection: Collection) -> None:
```
jsignell, line 268: These tests look like they were originally with a collection that didn't have a self_href. Is there a reason why they can't just be left as-is?

ircwaves: The current implementation of `Link.to_dict` requires a `Link` to have a `_href` or a `_target` with a `_href` (self href):

```python
        if self._href:
            data["href"] = self._href
        elif self._target and (href := self._target.get_self_href()):
            data["href"] = href
        else:
           raise ValueError("No href or target self href on the link")
```

Open to (and I'll try to think of) ideas of what might be a better way of handling this.

tests/v1/test_link.py
ircwaves: Good point.  incoming

tests/v1/test_link.py
jsignell: No that's fine I just wanted to make sure that this change wasn't inadvertent.

jsignell (approved):
This looks good to me!

PR #1694, opened 2026-04-07
chore: set all releases to rc

gadomski:
I'm still feeling my way around **release-please**, and I'm pretty sure this is the correct way to set all of our packages release to be release candidates. We'll see what the **release-please** action does to https://github.com/stac-utils/pystac/pull/1632 after merging.

We want to be on `rc` releases because of all the changes in https://github.com/stac-utils/pystac/pull/1650.


jsignell (approved):
Worth a shot!

PR #1697, opened 2026-04-08
fix: vcr header filtering

### LINKED ISSUE #1696: Get the header filtering back for cassettes
It looks like we lost vcr header filtering when we merged https://github.com/stac-utils/pystac/pull/1650: https://github.com/stac-utils/pystac/pull/1632/changes#diff-08c15b9339f64627fb39c07640b2c14864e9edacff18be071f7bc4575fb7ca5eR99-R142

Let's fix the tests so we don't get a bunch of VCR noise on every re-write.

gadomski:
**Related Issue(s):**

- Closes #1696

**Description:**

It's hard to confirm, but I _think_ I was running into something around plugin resolution order (`vcr_config` being defined in a plugin was making it be overwritten by another plugin, maybe?). I've moved `vcr_config` to a top-level `conftest.py` and then done a local `uv run pytest --record-mode rewrite` to confirm that it's working.

I sidecar tried to fix the RC versioning by setting the versioning strategy, per https://github.com/googleapis/release-please/blob/main/docs/customizing.md#versioning-strategies.


jsignell (approved):
This looks good, but I am thinking it probably won't work for externally managed extension classes. That is fine for now, but something to keep in mind.

gadomski:
Agreed ... I think that's on them, since they get to choose whether they do vcr-managed tests at all?

PR #1705, opened 2026-04-22
fix: type error

gadomski:
Flagged in https://github.com/stac-utils/pystac/pull/1592, popped up due to a dependency update.


jsignell (approved):
What a weird little typing-driven change!

gadomski:
> What a weird little typing-driven change!

Right?!

PR #1709, opened 2026-04-23
fix: V2 item collection

jsignell:
Get V1 tests passing for item collections

### DIFF
--- a/src/pystac/item_collection.py
+++ b/src/pystac/item_collection.py
@@ -69,7 +69,7 @@ class ItemCollection:
     def clone(self) -> ItemCollection:
         return copy.deepcopy(self)
 
-    @deprecated("Use `ItemCollection.from_dict` instead")
+    @deprecated("Try `ItemCollection.from_dict` and handle any exceptions instead")
     @staticmethod
     def is_item_collection(data: dict[str, Any]) -> bool:
         return data.get("type", "") == "FeatureCollection"
--- a/src/pystac/item_collection.py
+++ b/src/pystac/item_collection.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import copy
+import warnings
 from collections.abc import Iterator
 from pathlib import Path
 from typing import Any, Literal, TypedDict, override
@@ -20,14 +21,18 @@ class T_ItemCollection(TypedDict):
 
 
 class ItemCollection:
-    def __init__(self, items: list[Item], **kwargs: Any):
+    def __init__(self, items: list[Item], root: Container | None = None, **kwargs: Any):
         self.items: list[Item] = [Item.try_from(item) for item in items]
 
         if "root" in kwargs:
-            raise
+            raise KeyError(
+                "root is not a valid key word argument to ``ItemCollection()``. "
+                "If root is required try instantiating the items directly and "
+                "providing root."
+            )
 
         extra_fields = kwargs.pop("extra_fields", {})
-        self.extra_fields: dict[str, Any] =  kwargs
+        self.extra_fields: dict[str, Any] = kwargs
         if extra_fields:
             warnings.warn(
                 "Pass extra_fields entries as kwargs, "
--- a/src/pystac/item_collection.py
+++ b/src/pystac/item_collection.py
@@ -5,6 +5,10 @@ from collections.abc import Iterator
 from pathlib import Path
 from typing import Any, Literal, TypedDict, override
 
+from typing_extensions import deprecated
+
+from .container import Container
+from .errors import STACTypeError
 from .item import Item
 from .reader import DEFAULT_READER, Reader
 from .utils import make_absolute_href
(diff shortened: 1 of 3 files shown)


src/pystac/item_collection.py
```diff
@@ -17,8 +21,17 @@ class T_ItemCollection(TypedDict):
...
+
+        if "root" in kwargs:
+            raise
+
+        self.extra_fields: dict[str, Any]
+        if "extra_fields" in kwargs:
+            self.extra_fields = kwargs.pop("extra_fields")
+            assert len(kwargs) == 0
+        else:
+            self.extra_fields = kwargs
```
ircwaves: So, this translates into my head as constructing with `extra_fields` or arbitrary `kwargs`, but not both.  

Following your comment [here](https://github.com/stac-utils/pystac/pull/1693#discussion_r3124875530), I took it as a **_use kwargs, and then update from extra_fields_**:


```suggestion
        extra_fields = kwargs.pop("extra_fields", {})
        self.extra_fields: dict[str, Any] =  kwargs
        if extra_fields:
            warnings.warn(
                "Pass extra_fields entries as kwargs, "
                "instead of in extra_fields dictionary."
            )
            self.extra_fields.update(extra_fields)
```

Do other objects need to handle this too?  Is this all wrong, and `extra_fields` is a logical leak of a pystac internal field?


```suggestion
        self.extra_fields: dict[str, Any] = kwargs
        if "extra_fields" in kwargs:
            warnings.warn(
                "Extra field named "extra_fields" detected.  "
                "Potential naming collision with internal pystac fields."
            )
  
```

src/pystac/item_collection.py
```diff
@@ -29,6 +42,16 @@ def __getitem__(self, index: int) -> Item:
     def __iter__(self) -> Iterator[Item]:
         return iter(self.items)
 
+    def __contains__(self, __x: Item) -> bool:
```
ircwaves, line 52: Neat. I never thought of mangling an argument.

ircwaves (approved):
This looks good.  One question about `extra_fields`.

src/pystac/item_collection.py
jsignell: No that's perfect. I'll copy it here

src/pystac/item_collection.py
```diff
@@ -17,8 +21,19 @@ class T_ItemCollection(TypedDict):
 
 class ItemCollection:
     def __init__(self, items: list[Item], **kwargs: Any):
-        self.items: list[Item] = items
-        self.extra_fields: dict[str, Any] = kwargs
+        self.items: list[Item] = [Item.try_from(item) for item in items]
+
+        if "root" in kwargs:
+            raise
```
gadomski: Blank `raise`?

src/pystac/item_collection.py
```diff
@@ -29,6 +44,16 @@ def __getitem__(self, index: int) -> Item:
     def __iter__(self) -> Iterator[Item]:
         return iter(self.items)
 
+    def __contains__(self, __x: Item) -> bool:
+        return __x in self.items
+
+    def __add__(self, other: Any) -> ItemCollection:
+        if not isinstance(other, ItemCollection):
+            return NotImplemented
```
gadomski, line 57: Is this preferred over a `ValueError`? I never was sure when the "right" place was for `NotImplemented`....

src/pystac/item_collection.py
```diff
@@ -41,6 +66,14 @@ def to_dict(self) -> dict[str, Any]:
             **data,
         }
 
+    def clone(self) -> ItemCollection:
+        return copy.deepcopy(self)
+
+    @deprecated("Use `ItemCollection.from_dict` instead")
```
gadomski: Is this the correct deprecation message? `from_dict` is a conversion method, not a validation.

src/pystac/item_collection.py
jsignell: I must have gotten distracted halfway through that one 😬

src/pystac/item_collection.py
jsignell: Yeah I think for commutative mehods like `__add__` and `__eq__` specifically returning a `NotImplemented` is what allows the reverse operation to be used instead. So when you call `+` it tries to use the left hand side as self and right hand side as other and then if that returns `NotImplemented` it tries rhs as self and lhs as other.

src/pystac/item_collection.py
jsignell: Hmm I guess I maybe skipped a few steps in my thinking. I was imagining that you would replace:

```python
if ItemCollection.is_item_collection(data):
    ic = ItemCollection.from_dict(data)
```

with this:
```python
try:
    ic = ItemCollection.from_dict(data)
except:
    pass
```
I couldn't imagine how else you'd use it. 

Do you agree that it should be deprecated? One the one hand it is just a little convenience method. On the other hand it's weird.

src/pystac/item_collection.py
gadomski: TIL! Thanks!

src/pystac/item_collection.py
gadomski: Totally, the second is more "pythonic". Recommend maybe this language then:

```suggestion
    @deprecated("Try `ItemCollection.from_dict` and handle any exceptions instead")
```

gadomski (approved):

PR #1710, opened 2026-04-23
fix: V2 summaries

jsignell:
~I don't quite have typing figured out on `Summaries` yet, but wanted to get this up.~

This one is ready to go


scripts/pull-static
```diff
@@ -5,7 +5,7 @@ set -e
 VERSION="1.5.0"
 SRC="https://cdn.jsdelivr.net/npm/@radiantearth/stac-fields@$VERSION/fields-normalized.json"
 HERE=$(dirname "$0")
-DEST=$(dirname "$HERE")/pystac/static/fields-normalized.json
```
jsignell, line 8: @tylanderson if you do `mkdir` on this `DEST` and then run `./scripts/pull-static` with this diff you'll get your static file populated.

src/pystac/item.py
```diff
@@ -235,7 +236,9 @@ def __geo_interface__(self) -> dict[str, Any]:
 
 
 @final
-class Properties(Basics, DateTime, Licensing, Providers, Instrument):
+class Properties(
+    MutableMapping[str, Any], Basics, DateTime, Licensing, Providers, Instrument
```
jsignell, line 240: This makes it so properties can pretty much be treated like a dict. We might want to do this in some other places in the codebase.

jsignell:
Ok so I was worried about changing `Summaries` because it is used a lot in the extension classes which we are trying not to change too much, but it looks like they only use `get_list` and `get_range`. So I will see if there is a way to simplify the class a bit and mostly just offer those two.

jsignell:
Ok I didn't quite get the typing working, but I did rewrite `Summaries`

jsignell:
Ok I have the typing working for summaries.

gadomski (approved):

PR #1723, opened 2026-05-08
chore: actually set release-please manifest correctly

gadomski:
🤦🏼 I did it wrong in https://github.com/stac-utils/pystac/pull/1721

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,6 +1,6 @@
 {
-  ".": "1.14.3-rc.0",
-  "core": "1.14.3-rc.0",
+  ".": "1.15.0-rc.0",
+  "core": "1.15.0-rc.0",
   "extensions/classification": "2.0.0-rc.0",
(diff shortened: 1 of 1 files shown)


ciaransweet (commented):
Should every version still be an rc? (and should they be getting bumps?)

gadomski:
> Should every version still be an rc? (and should they be getting bumps?)

None of the extension packages have ever been released, so they don't need bump.

For the extensions, I _think_ an rc would be a good idea, but I'm not 100% sure ... thoughts?

ciaransweet:
> > Should every version still be an rc? (and should they be getting bumps?)
> 
> None of the extension packages have ever been released, so they don't need bump.
> 
> For the extensions, I _think_ an rc would be a good idea, but I'm not 100% sure ... thoughts?

I guess it's a question of 'how likely are they to be installed separately vs just bundle in with this package'

I think having them isn't an issue anyway.

ciaransweet (approved):

Issue #1725, opened 2026-05-11
pystac v1.15.0 release checklist


gadomski:
Since we refactored all the extensions to their own packages in https://github.com/stac-utils/pystac/pull/1650, we will move a little more slowly for the v1.15.0 release of PySTAC. The plan is:

- [x] Update our release workflow to release each extension's package, in addition to the core packages
- [x] Update the main packages and each extensions' version to have a `-rc.0` suffix (we've tried to do this with **release-please** but have failed, so falling back to doing it manually)
- [x] Manually (at first, maybe eventually automatically) create `-rc.0` releases for the main packages and each extensions
  - [x] pystac-core
  - [x] pystac-ext-classification
  - [x] pystac-ext-datacube
  - [x] pystac-ext-eo
  - [x] pystac-ext-file
  - [x] pystac-ext-grid
  - [x] pystac-ext-item-assets
  - [x] pystac-ext-label
  - [x] pystac-ext-mgrs
  - [x] pystac-ext-mlm
  - [x] pystac-ext-pointcloud
  - [x] pystac-ext-projection
  - [x] pystac-ext-raster
  - [x] pystac-ext-render
  - [x] pystac-ext-sar
  - [x] pystac-ext-sat
  - [x] pystac-ext-scientific
  - [x] pystac-ext-storage
  - [x] pystac-ext-table
  - [x] pystac-ext-timestamps
  - [x] pystac-ext-version
  - [x] pystac-ext-view
  - [x] pystac-ext-xarray-assets
  - [x] pystac
- [x] Check for errors/problems and correct
- [x] Once everthing's ok, release the non-rc versions

> [!WARNING]
> I'm going to be pausing any extension PR reviews until we get v1.15.0 released.

gadomski:
Our first new package's rc successfully published: https://pypi.org/project/pystac-core/

gadomski:
I've completed the initial task of creating each new package on PyPI, and am currently working through debugging the rc release: https://github.com/stac-utils/pystac/pull/1735

gadomski:
I've made announcements in CNG Slack and on [Bluesky](https://bsky.app/profile/gadom.ski/post/3mmwolynyqc2o), and created a discussion for the rc: https://github.com/radiantearth/stac-spec/discussions/1385. I'm going to let things cook about a week or so before doing the actual release.

gadomski:
https://pypi.org/project/pystac/1.15.1/

PR #1726, opened 2026-05-11
fix: update all of our versions to be rc.0

gadomski:
**Related Issue(s):**

- Helping https://github.com/stac-utils/pystac/issues/1725

**Description:**

- Update all of our package versions to have a `-rc.0` suffix
- Update our release workflow to work for all tags

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,20 +9,38 @@ permissions: {}
 
 jobs:
   release:
-    name: release
+    name: Release
     runs-on: ubuntu-latest
+    if: github.repository == 'stac-utils/pystac'
     environment:
       name: pypi
-      url: https://pypi.org/p/pystac
     permissions:
       id-token: write
-    if: ${{ github.repository }} == 'stac-utils/pystac'
     steps:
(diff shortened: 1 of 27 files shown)


.github/workflows/release.yml
```diff
@@ -9,20 +9,38 @@ permissions: {}
 
 jobs:
   release:
-    name: release
+    name: Release
```
gadomski, line 12: Nitpicky change

.github/workflows/release.yml
```diff
@@ -9,20 +9,38 @@ permissions: {}
...
     runs-on: ubuntu-latest
+    if: github.repository == 'stac-utils/pystac'
     environment:
       name: pypi
-      url: https://pypi.org/p/pystac
```
gadomski, line 16: This won't be true, so we just drop it

release-please-config.json
```diff
@@ -6,6 +6,7 @@
     ".": {
       "changelog-path": "CHANGELOG.md",
       "component": "pystac",
+      "include-component-in-tag": false,
```
gadomski, line 9: We want the main, top-level package to still just have a `vX.Y.Z` tag, not `pystac-vX.Y.Z`

ciaransweet (approved):

Issue #1716, opened 2026-04-30
V2: setting custom global reader/writer fails


tylanderson:
In pystac v1, StacIO was used for reading and writing, and users could define a custom StacIO class and set it as the default to be used. This was used in stactools, e.g. `use_fsspec()`, which inside the function would set a new default reader.

In v2, there are functions, `set_default_reader()`/`set_default_writer()` which are to be used to achieve a similar objective, but it looks like what happens is usage of this fails as functions bind to the standard `DEFAULT_READER` even after updating the global via `set_default_reader()`

Small Example:

```python
import pystac

from pystac.reader import set_default_reader, StandardLibraryReader

class CustomReader(StandardLibraryReader):
    def get_json(self, href, **kwargs):
        print("Using CustomReader to read:", href)
        return super().get_json(href, **kwargs)

set_default_reader(reader=CustomReader())
stac_item = pystac.read_file('tests/data-files/examples/v1.1.0/collection.json')
print(stac_item.reader)
items = list(stac_item.get_items())
print(items[0].reader)
```

Issue #1715, opened 2026-04-28
V2: Reading io.read_file, custom reader is not set on object


tylanderson:
Looks like when using the top level `read_file` with a custom reader, the custom reader is not propagated down to the object. This means that then calling `get_items()`, for example, the target will be read with the standard default reader.

It looks like STACObject has a [from_file](https://github.com/stac-utils/pystac/blob/1a37397982f346277373b565db607cc229fb57cd/src/pystac/stac_object.py#L92) method that handles setting a reader on the object that could be used

PR #1731, opened 2026-05-15
V2 reading and writing fixes

tylanderson:
**Related Issue(s):**

- #1715 
- #1716 

**Description:**

For #1715 `read_file` now sets `reader` on the returned object (using the reader used to load it), so that subsequent link resolutions (e.g. `get_items`) uses the same reader. `from_dict` is still used for deserialization to preserve the types.

For #1716, replaced direct references to the `DEFAULT_READER`/`DEFAULT_WRITER` module-level globals with `get_default_reader()` / `get_default_writer()` accessor functions. This fixes a subtle bug where functions would capture the original instance at import time and never see updates made via `set_default_reader()`.

**PR Checklist:**

- [ ]


tylanderson:
Are v1 tests just pulled over and not planned to be used eventually after v2 is sorted? I added a new io test file (not in V1) for these

gadomski:
> Are v1 tests just pulled over and not planned to be used eventually after v2 is sorted? I added a new io test file (not in V1) for these

We're hoping to keep as many v1 tests as we can, to maximize our backwards compatibility. Any new tests should go in `tests/` (not `tests/v1`).

tylanderson:
Noting that we could also probably approach the custom reader issue differently, e.g. having set_default_reader mutate an reader attribute of some root reader class

tests/utils/mock_io.py
gadomski, line 1: Can you just throw this in `conftest.py` and make it a fixture? That way you can use fixture clean-up instead of having to wrap in a `try/except`.

gadomski (changes requested):

gadomski (approved):

PR #1745, opened 2026-06-10
fix: persist stac_io when reading items

### LINKED ISSUE #1743: Item.links[].get_href() does not pass through stac_io
`Link.get_href` has this "transform_href" feature, which involves getting and resolving the root of the link owner, but in this code path, the custom stac_io of that owner is not passed through.

We use a custom StacIO implementation to provide some client side caching, but that cache is ignored here.

Reproduction example:

```python
import pystac
import pystac.stac_io
import logging

logging.basicConfig(level=logging.INFO)
logging.getLogger("pystac.stac_io").setLevel(logging.DEBUG)

class CustomStacIO(pystac.stac_io.DefaultStacIO):
    def read_text_from_href(self, href: str) -> str:
        print(f"!!! CUSTOM GET OF {href}")
        return super().read_text_from_href(href=href)

custom_stac_io = CustomStacIO()

item_url = "https://stac.dataspace.copernicus.eu/v1/collections/sentinel-2-l2a/items/S2B_MSIL2A_20260608T081559_N0512_R121_T48XVQ_20260608T104640"

item = pystac.read_file(item_url, stac_io=custom_stac_io)

item.get_single_link(rel="self").get_href()
```

On the `read_file`, there is both logging and printing of getting the item url:
```
DEBUG:pystac.stac_io:GET https://stac.dataspace.copernicus.eu/v1/collections/sentinel-2-l2a/items/S2B_MSIL2A_20260608T081559_N0512_R121_T48XVQ_20260608T104640 Headers: {}
!!! CUSTOM GET OF https://stac.dataspace.copernicus.eu/v1/collections/sentinel-2-l2a/items/S2B_MSIL2A_20260608T081559_N0512_R121_T48XVQ_20260608T104640
```

On the `get_href` there is only logging of the catalog URL, because `CustomStacIO` is not passed through to the related logic
```
DEBUG:pystac.stac_io:GET https://stac.dataspace.copernicus.eu/v1/ Headers: {}
```

gadomski:
**Related Issue(s):**

- Closes #1743 

**Description:**

Previously, we only persisted StacIO when reading catalogs. _To me_, there's no reason to limit StacIO persistence to catalogs, so we add it to collections and items here. We also update links to check to see if their owner has a StacIO to use.

**PR Checklist:**

### DIFF
--- a/core/pystac/item.py
+++ b/core/pystac/item.py
@@ -108,6 +108,11 @@ class Item(STACObject, Assets):
     stac_extensions: list[str]
     """List of extensions the Item implements."""
 
+    _stac_io: pystac.StacIO | None = None
+    """Optional instance of StacIO that will be used by default for any IO
+    operations. This is set when an item is read by a StacIO instance.
+    """
+
     STAC_OBJECT_TYPE = STACObjectType.ITEM
 
     def __init__(
--- a/core/pystac/link.py
+++ b/core/pystac/link.py
@@ -321,6 +321,8 @@ class Link(PathLike):
             if root is not None:
                 obj = root._resolved_objects.get_by_href(target_href)
                 stac_io = root._stac_io
+            if stac_io is None and self.owner and hasattr(self.owner, "_stac_io"):
+                stac_io = self.owner._stac_io
 
             if obj is None:
                 if stac_io is None:
--- a/core/pystac/stac_io.py
+++ b/core/pystac/stac_io.py
@@ -162,21 +162,25 @@ class StacIO(ABC):
         d = migrate_to_latest(d, info)
 
         if info.object_type == pystac.STACObjectType.CATALOG:
-            result = pystac.Catalog.from_dict(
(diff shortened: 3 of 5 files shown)


hrodmn (approved):
This seems like a good change to me because it makes it easier to use custom StacIO objects in a more targeted fashion instead of setting `DefaultStacIO` like I always have 👍. 

I just want to make sure I understand the implications since `Collection` -> `Item` types of traversals are the thing I have least experience with in pystac.

Before this change, if you provided a custom `stac_io` while loading a `Catalog` that `stac_io` would persist for all STAC operations on related links. However if you provided a custom `stac_io` when loading a `Collection` or `Item`, that `stac_io` would just disappear and wouldn't be used for subsequent i/o tasks (like fetching from related links). Now the `Item`'s `stac_io` persists and gets used for i/o operations on related links.

gadomski:
Thanks @hrodmn , and your narrative of the changes is correct.

PR #1756, opened 2026-06-25
fix: remove invalid version specifier

gadomski:
Our v1.15.0 release was bad b/c we had a leftover invalid dependency specifier, very much my fault 🤦🏼 . I've yanked the bad release, and will do a post-release with this fix. I don't think we need to go to v1.15.1 but happy to be convinced otherwise.

**release-please** is being a bit awkward with this new layout, so I might have to rethink our releasing after the extension breakout.

Sorry for the churn @tylanderson

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ classifiers = [
 ]
 requires-python = ">=3.10"
 dependencies = [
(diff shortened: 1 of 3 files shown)


tylanderson (approved):
No worries, yeah seems like something to revisit to see how we can make it easier

PR #1764, opened 2026-07-01
fix: circular imports for extensions

### LINKED ISSUE #1763: Fix circular dependency in pystac-core
We didn't complete the extension migration correctly, https://github.com/stac-utils/pystac/blob/main/core/pystac/extensions/ext.py still contains references to the extension modules. That file should be distributed with **pystac**, not **pystac-core**.

Initially reported in https://github.com/conda-forge/staged-recipes/pull/34020#discussion_r3505896782.

gadomski:
We didn't get the extension refactor correct in https://github.com/stac-utils/pystac/pull/1650 (my fault). Rather than static imports, this PR converts things to an extension registry via an entry point group.

I'm not sure how bad the https://pypi.org/project/pystac/1.15.1/ release is — while the new sub-packages aren't correct, it's still fine to use the top level **pystac** package (which is all that existed before v1.15). I'm inclined to _not_ yank v1.15.1, but happy to hear other opinions.


core/pystac/extensions/hooks.py
```diff
@@ -75,8 +83,27 @@ def migrate(
...
+        """
+        if self._discovered:
+            return
+        self._discovered = True
+        import warnings
+        from importlib.metadata import entry_points
+
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", DeprecationWarning)
+            for entry_point in entry_points(group="pystac.extensions"):
```
hrodmn, line 104: the `entry_points` approach is elegant!

core/pystac/__init__.py
```diff
@@ -91,56 +90,11 @@
 from pystac.utils import HREF
 
 import pystac.extensions.hooks
-import pystac.extensions.classification
```
hrodmn, line 94: ah yeah, pystac.extensions.classification imports pystac 🔃

extensions/classification/pyproject.toml
```diff
@@ -19,7 +19,13 @@ classifiers = [
     "Programming Language :: Python :: 3.13",
 ]
 requires-python = ">=3.10"
-dependencies = ["pystac-core"]
+dependencies = [
+    "pystac-core",
+    "pystac-ext-raster",
```
hrodmn, line 24: I like how the related extensions are explicitly listed as dependencies here

hrodmn (approved):
The entry_points system seems like a great way to handle this when the extension packages are truly optional. Thanks for the fix @gadomski!

core/pystac/__init__.py
gadomski: Yeah I missed this one _hard_ on the initial refactor

PR #1796, opened 2026-08-04
docs: tweak how we version extension packages

gadomski:
It was just going to be too much of a PITA to use post-release versions for extension packages with **release-please**. I'd like to relax our extension versioning to use PATCH releases for all PyPI updates that don't mirror a change in the extension specification itself.

This gives us the slightly awkward situation where a v1.2.3 **pystac** extension version could correspond to a v1.2.1 extension version, but in my experience most extensions don't do PATCH versions, so it doesn't feel like _that_ big a problem? Happy to get push-back, or feedback that I need to make more documentation — this new "extensions as their own packages" thing is pretty fresh.


hrodmn (approved):
It makes sense to me to leave room for code patches outside of the release cycle of the actual extension specs.

PR #1804, opened 2026-09-11
fix: recursively convert values in PropertiesExtension._set_property

### LINKED ISSUE #1802: Method `pystac.extensions.base.PropertiesExtension._set_property` does not recursively convert the value being set to `dict`
## Bug
Method [`pystac.extensions.base.PropertiesExtension._set_property`](https://github.com/stac-utils/pystac/blob/1b6faa38815ba3cd02d75a2ac367f0e18bd0021d/core/pystac/extensions/base.py#L84) is used by many `pystac` extensions to set a property of name `prop_name` to value `v` transforming it into a dictionary so that the `PropertiesExtension` instance reamains serializable. However, it does not perform recursive conversion and, therefore, the `PropertiesExtension` object becomes unserializable when deeply nested elements of `v` correspond to class instances.

## Proposed Solution

Define function `to_dict_rec` that recursively converts an object to a dictionary using its very `to_dict` method and all the other ones of its nested elements:

```python
def to_dict_rec(obj: Any) -> Any:
    """
    Recursively convert an object to a dictionary, if it has a `to_dict()` method.

    Parameters
    ----------
    obj : Any
        The object to convert to a dictionary, if it has a `to_dict()` method.

    Returns
    -------
    Any
        A dictionary representation of the object if it has a `to_dict()` method,
        otherwise the object itself.
    """

    # Get object's to_dict method, if it exists
    to_dict = getattr(obj, "to_dict", None)

    # If the method is indeed a method (callable), proceed with the recursive conversion
    if callable(to_dict):
        obj = to_dict_rec(to_dict())

    # If the object is a dictionary, list or tuple, recursively convert its elements
    if isinstance(obj, dict):
        obj = {key: to_dict_rec(value) for key, value in obj.items()}
    if isinstance(obj, (list, tuple)):
        obj = type(obj)(to_dict_rec(value) for value in obj)

    return obj
````

Replace method [`pystac.extensions.base.PropertiesExtension._set_property`](https://github.com/stac-utils/pystac/blob/1b6faa38815ba3cd02d75a2ac367f0e18bd0021d/core/pystac/extensions/base.py#L84) by another that uses such `to_dict_rec` method:

```python
def _set_property_fixed(
    self, prop_name: str, v: Any | None, pop_if_none: bool = True
) -> None:
    """
    Set a property of name (key) `prop_name` and value `v` to the instance, recursively
    converting it to a dictionary if it is dictionary-convertible.

    Parameters
    ----------
    prop_name : str
        The name of the property to set.

    v : Any or None
        The value to set to the property. If `None` and `pop_if_none` is True, the
        property is removed from the instance.

    pop_if_none : bool, default=True
        Whether to remove the property from the instance if the value `v` is `None`.
    """
    if v is None and pop_if_none:
        self.properties.pop(prop_name, None)
    else:
        self.properties[prop_name] = to_dict_rec(v)
```

devtejasx:
**Related Issue(s):**

- Closes #1802

**Description:**

`_set_property` only called `to_dict()` on the items of a top-level list, so an object nested any deeper (in a dict, in a tuple, or inside what another `to_dict()` returned) was stored as-is and the item could no longer be serialized to JSON.

This adds a small `_to_dict_recursive` helper in `pystac/extensions/base.py` and uses it for every value `_set_property` stores. It follows the approach suggested in the issue.

One behaviour change to be aware of: a dict passed to `_set_property` is now stored as a converted copy rather than the same object, the same way lists already were.

Added `test_set_property_converts_nested_objects` to `tests/extensions/test_custom.py`. It fails on `main` and passes here; the rest of the suite is unchanged (1542 passed, 5 skipped). The 20 errors I get locally are the Windows path-length `shutil` errors in the asset fixtures and happen on `main` too.

**PR Checklist:**

**AI tool usage:**

- [x] AI tools supported the development of this PR, per the [stac-utils AI tool policy](https://stac-utils.github.io/ai-contribution-policy/). I have reviewed the change and can answer questions about it.

### DIFF
--- a/core/pystac/extensions/base.py
+++ b/core/pystac/extensions/base.py
@@ -45,6 +45,22 @@ class SummariesExtension:
 P = TypeVar("P")
 
 
+def _to_dict_recursive(v: Any) -> Any:
+    """Replaces any object that has a ``to_dict`` method with the result of calling
+    it, including objects nested at any depth inside lists, tuples and dicts."""
+    to_dict = getattr(v, "to_dict", None)
+    if callable(to_dict):
+        v = to_dict()
+    if isinstance(v, dict):
+        return {key: _to_dict_recursive(value) for key, value in v.items()}
+    elif isinstance(v, list):
+        return [_to_dict_recursive(x) for x in v]
+    elif isinstance(v, tuple):
+        return tuple(_to_dict_recursive(x) for x in v)
+    else:
+        return v
+
+
 class PropertiesExtension(ABC):
     """Abstract base class for extending the properties of an :class:`~pystac.Item`
     to include properties defined by a STAC Extension.
@@ -86,12 +102,8 @@ class PropertiesExtension(ABC):
     ) -> None:
         if v is None and pop_if_none:
             self.properties.pop(prop_name, None)
-        elif isinstance(v, list):
-            self.properties[prop_name] = [
-                x.to_dict() if hasattr(x, "to_dict") else x for x in v
-            ]
         else:
-            self.properties[prop_name] = v
+            self.properties[prop_name] = _to_dict_recursive(v)
 
 
 class ExtensionManagementMixin(Generic[S], ABC):
--- a/tests/extensions/test_custom.py
+++ b/tests/extensions/test_custom.py
@@ -1,5 +1,6 @@
 """Tests creating a custom extension"""
 
+import json
 from collections.abc import Generator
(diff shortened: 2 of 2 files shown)


gadomski (approved):
LGTM, though I do want to call out that deeply nested objects in STAC is a bit of an anti-pattern, which might be why we haven't run into this before.

